Issue #4, opened 2017-07-25
Support preact out of the box


kentcdodds:
@zouhir says he'll be working on this a bit :)

PR #15, opened 2017-07-26
WIP: preact alias, rollup configs + compat-lite

zouhir:
🚫 Not ready for merge.

I've only sent a PR from my private fork for reviews and opinions. 

Having a little issue with #4 when I rollup using the configs below, only reason I've posted here is I have not fiddled much recently with React tools and specs in depth. I'll be investigating that..

- [x] remove `dom-scroll-into-view`
- [x] update proptypes to compat `number, obj, any`
- [ ] add build default to build both preact & react one
- [ ] pass environment variable from Rollup to save a bunch of no-prod React stuff.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,8 @@
     "npm": "> 3"
   },
   "scripts": {
+    "build": "rollup -c --environment LIBRARY:react -o dist/index.js",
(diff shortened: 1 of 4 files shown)


kentcdodds:
Alrighty! I've created a build and added preact support!! :tada:

Thank you @zouhir! I used this as an example for a bunch of stuff. Much appreciated :smile:

Issue #1, opened 2017-07-25
Help make examples


kentcdodds:
**Latest Update**:

Alrighty! We're getting really close now! Just need to tie up some loose ends and make sure we cover all the use cases we want to (ALL OF THEM).

We've released an official "release candidate"

You can install it with:

```
npm install --save downshift@rc
```

If you want to help, try implementing an autocomplete/combobox/multi-select/etc. with downshift. You can:

1. fork [the example codesandbox here](http://kcd.im/ds-example)
2. give it a good title and description
3. Update the version (go to dependencies, type `downshift` for the "package name" and `rc` for the "version"). Do this even though it already appears in the list of dependencies.
4. Add the `downshift:example` tag. That way it shows up in [this search](https://codesandbox.io/search?refinementList%5Btags%5D%5B0%5D=downshift%3Aexample&page=1)
5. Ping us here when you're done. Thanks!

---

**Old Update**:

Hey folks! :wave: now that we've released a beta version, it's _much_ easier to make examples. Now you can make them _right in the browser_ :tada: (thank you @CompuIves!!!)

So, if you've made an example already, would you mind forking this example: http://kcd.im/rac-example and updating it to your example? Then give it a good title and description and **add the `react-autocompletely:example` tag** That way it shows up in [this search](https://codesandbox.io/search?refinementList%5Btags%5D%5B0%5D=react-autocompletely%3Aexample&page=1)

I'll add a link to that in the README so folks can go see all the awesome examples you all make. Thanks!!!

---

**Old stuff**

Hey :wave:

We need help making examples. I've scaffolded out things for the examples already, but we need more and need to improve the ones that we already do have. Here's a quick rundown of how to get up and running with the examples:

```bash
git clone https://github.com/kentcdodds/react-autocompletely.git
cd react-autocompletely
npm install
cd examples
npm install
npm start
```

Then open http://localhost:3000/example-name

In the `examples` directory, you'll find a `pages` directory. Each of those has an `index.js` file which is rendered at that route. For example, [`semantic-ui/index.js`](https://github.com/kentcdodds/react-autocompletely/blob/master/examples/pages/semantic-ui/index.js) will be rendered at: http://localhost:3000/semantic-ui

So, the idea for these examples is they can be examples of just about anything. I have a few in there to show examples of how to implement the API or look and feel that other libraries have. That's mostly to demonstrate the power of the primitives available in react-autocomplete.

[...]

PR #33, opened 2017-07-27
Examples: update forwarded prop item -> value

rezof:
the basic and sementic-ui examples had the old item prop in the forwardProps


kentcdodds:
Thanks! We _are_ going to be deleting this folder in favor of [this](https://codesandbox.io/search?refinementList%5Btags%5D%5B0%5D=react-autocompletely%3Aexample&page=1) (see #1) 
But I'll merge this for now :smile:

PR #45, opened 2017-07-28
feat(api): only return render prop from Menu component

souporserious:
**What**:
This allows users to always render the menu similar to the option here in [react-autosuggest](https://github.com/moroshko/react-autosuggest#alwaysRenderSuggestionsProp). It also keeps a clean DOM by rendering the status to the end of the document as well as switches `children` function props to be called `render` for better API clarity.

**Why**:
This will allow more complex menus and animations to be made by giving the user full control over rendering the `Menu` component. It's a little more work to hook `isOpen` up, but worth the benefit in my opinion. Plus you're already pulling things like `selectedItem` and `highlightedIndex` off.

I switched the `children` prop to `render` to follow conventions similar to React Router. After implementing just a returned `children` function, it felt weird that it was called "children" when it wasn't really returning any children. I can change this back, just let me know. I figured I'd see how you felt and then could revert if necessary 😇 

**How**:
I moved the `MenuStatus` into pretty much a small portal that renders the status to the end of the document so we can keep a clean DOM and simple API for the `Menu` component. 

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

I'd love to hear your thoughts on this 🙂 I'll need to update the other examples if you are ok with these changes. It's a lot of moving parts for the `Menu` component, but I think it will allow for a little more flexibility.

This was the inspiration for this:
image
We need more control over the menu's rendering in a situation like this since the input is grouped alongside the items.

### DIFF
--- a/examples/pages/basic/index.js
+++ b/examples/pages/basic/index.js
@@ -109,21 +109,27 @@ function BasicAutocomplete({items, onChange}) {
     <Autocomplete onChange={onChange}>
       <Input placeholder="Favorite color ?" />
       <Autocomplete.Menu style={{border: '1px solid rgba(34,36,38,.15)'}}>
-        {({inputValue, selectedItem, highlightedIndex}) =>
-          // prettier is doing weeeeird things to this
-          // prettier-ignore
-          items.filter(i => !inputValue || i.toLowerCase().includes(inputValue.toLowerCase()))
-            .map((item, index) => (
-              <Item
-                value={item}
-                index={index}
-                key={item}
-                highlightedIndex={highlightedIndex}
-                selectedItem={selectedItem}
-              >
-                {item}
-              </Item>
-            ))}
+        {({isOpen, inputValue, selectedItem, highlightedIndex}) =>
+          isOpen &&
+          <div>
+            {items
+              .filter(
+                i =>
+                  !inputValue ||
+                  i.toLowerCase().includes(inputValue.toLowerCase()),
+              )
+              .map((item, index) =>
+                (<Item
+                  value={item}
+                  index={index}
+                  key={item}
+                  highlightedIndex={highlightedIndex}
+                  selectedItem={selectedItem}
+                >
+                  {item}
+                </Item>),
+              )}
+          </div>}
       </Autocomplete.Menu>
     </Autocomplete>
   )
--- a/src/controller.js
+++ b/src/controller.js
@@ -17,9 +17,7 @@ class Controller extends Component {
   }
 
   render() {
-    return this.props.children(
-      this.autocomplete.getControllerStateAndHelpers(),
-    )
+    return this.props.children(this.autocomplete.getControllerStateAndHelpers())
   }
 }
 
--- a/src/menu.js
+++ b/src/menu.js
@@ -1,10 +1,10 @@
 import React, {Component} from 'react'
+import ReactDOM from 'react-dom'
 import PropTypes from 'prop-types'
-import mitt from 'mitt'
 
 import MenuStatus from './menu-status'
 import {AUTOCOMPLETE_CONTEXT, MENU_CONTEXT} from './constants'
-import {cbToCb, compose, scrollIntoView} from './utils'
+import {cbToCb, scrollIntoView} from './utils'
 
 class Menu extends Component {
   static contextTypes = {
@@ -16,7 +16,6 @@ class Menu extends Component {
   }
 
   static propTypes = {
-    ref: PropTypes.func,
     defaultHighlightedIndex: PropTypes.number,
     getA11yStatusMessage: PropTypes.func,
     children: PropTypes.func.isRequired,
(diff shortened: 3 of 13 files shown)


souporserious:
@zouhir I'd love to hear your thoughts on this and if it helps or worsens this issue https://github.com/paypal/react-autocompletely/issues/44. How does ReactDOM play with Preact?

kentcdodds:
I haven't looked at the code and only looked a little bit at your description and can say that I already love it. I didn't like rendering divs for people and wanted to give total rendering power to users so this is awesome. Let me give it a look later. Thanks!

souporserious:
Awesome! Please let me know anything you want changed. Also, I was thinking and I'm wondering if we might be able to do better with the `MenuStatus` and build it into a small helper that just renders a div to the end of the page using just `render` instead of `unstable_renderSubtreeIntoContainer`. I'm going to play around with it right now, I'm worried what I have right now will cause problems with Preact.

souporserious:
I tried using just `render` but we can't since we use context 😕 @zouhir is `unstable_renderSubtreeIntoContainer` useable with Preact or will what I have be ok?

zouhir:
thanks @souporserious for this idea, I quite like what you've done!!

Eventually everything ReactDOM can Render is compatible with preact, including `unstable_renderSubtreeIntoContainer` but then all goes down to how thin and light our compat layer going to be, for me, I'd like to keep as light as possible for minimal to non potential failures.

in 5-10 lines of code we can do everything except `children` which has to be `children[0]` or children.only.

I think I can compat `unstable_renderSubtreeIntoContainer` have to try!

kentcdodds:
> I tried using just `render` but we can't since we use context

Is there a reason that we couldn't use `render` in `componentDidUpdate` to render pure JSX to the bottom of the body? So all the logic could happen in the `MenuStatus` where it is, and the `MenuStatus` just renders nothing itself and then imperatively renders maybe?

examples/pages/basic/index.js
```diff
@@ -108,23 +108,30 @@ function BasicAutocomplete({items, onChange}) {
...
-                highlightedIndex={highlightedIndex}
-                selectedItem={selectedItem}
-              >
-                {item}
-              </Item>
-            ))}
-      </Autocomplete.Menu>
+      <Autocomplete.Menu
+        style={{border: '1px solid rgba(34,36,38,.15)'}}
+        render={({isOpen, inputValue, selectedItem, highlightedIndex}) =>
```
kentcdodds: I really prefer to use the `children` prop rather than a `render` prop :wink:

examples/pages/basic/index.js
```diff
@@ -108,23 +108,30 @@ function BasicAutocomplete({items, onChange}) {
...
+          isOpen &&
```
kentcdodds: I kinda like that the autocomplete component renders `null` for me if `isOpen` is true, that way I can avoid this check. Otherwise _everyone_ will need to do this... Unless you can think of another use case where I'd want to render if if `isOpen` is false... 🤔

src/controller.js
```diff
@@ -17,9 +17,7 @@ class Controller extends Component {
   }
 
   render() {
-    return this.props.children(
-      this.autocomplete.getControllerStateAndHelpers(),
-    )
+    return this.props.render(this.autocomplete.getControllerStateAndHelpers())
```
kentcdodds: Here too, I prefer the `children` prop...

src/menu.js
```diff
@@ -116,28 +125,28 @@ class Menu extends Component {
   componentWillUnmount() {
     this.autocomplete.removeMenu(this)
     this.autocomplete.emitter.off('menu:open', this.setDefaultHighlightedIndex)
+
+    // cleanup MenuStatus component
+    ReactDOM.unmountComponentAtNode(this._statusNode)
+    document.body.removeChild(this._statusNode)
```
kentcdodds: Good call :+1:

I wonder if we could move all this `componentWillUnmount` stuff to a method called `cleanup` that's set on the instance in `componentDidMount` that way `componentWillUnmount` just calls `cleanup` and we keep this logic closer together?

kentcdodds:
FYI, I just merged something that'll probably conflict with you. Sorry!

examples/pages/basic/index.js
souporserious: Cool, no worries, I'll add it back 👍

examples/pages/basic/index.js
souporserious: I mentioned this a little above, but basically it allows for things [like this](https://github.com/moroshko/react-autosuggest#alwaysRenderSuggestionsProp) as well as more control when using animations. It will also help when creating a dropdown that contains an input inside the menu.

kentcdodds: Ok, sounds awesome :+1:

souporserious:
Alright, I think I did everything! 😅 I was able to use `render` by just passing props to `MenuStatus` and ditching context. 🙌  Updated all of the examples and everything seems to be working 👌

kentcdodds:
The test is busted 😘

souporserious:
On it 🏃

souporserious:
Done ✅

kentcdodds:
Thanks!

PR #46, opened 2017-07-28
Correction on blur event handler on iOS

psicotropicos:
Removing onBlur handler and reset function


src/input.js
```diff
@@ -67,12 +65,6 @@ class Input extends Component {
...
   }
 
-  handleBlur = () => {
-    if (!this.autocomplete.isMouseDown) {
-      this.autocomplete.reset()
-    }
-  }
-
   reset = () => {
     moveCursorToTheEnd(this._inputNode)
```
dashed: @psicotropicos I think this was also supposed to be removed. 

I tried this PR in a project and wasn't unable to unfocus in Safari.

kentcdodds: Hah, yeah, maybe we should remove this call (and the method itself from utils because it's not used elsewhere).

Would you like to make that change @dashed?

dashed: PR @ https://github.com/paypal/react-autocompletely/pull/50

PR #47, opened 2017-07-28
feat(component): added Button component to control Menu

souporserious:
**What**:
A `Button` component that applies proper ARIA attributes and allows controlling the opening and closing of the menu as well as keyboard navigation of the list of items.

**Why**:
This helps us build accessible dropdown + autocomplete solutions.

**How**:
I added a `Button` component that mimics some of the behavior of the `Input`.

**Checklist**:

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged 

Another addition to the API 😇 I think this will help with having an accessible button handy to control the `Menu`. I used it real quick in your semantic-ui demo and it worked nice 👌 applies all the proper ARIA attributes and allows for keyboard navigation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,10 @@ This is called with an object with the properties listed below:
 | `selectItemAtIndex`     | `function(index: number)`  | selects the item at the given index                                                                              |
 | `selectHighlightedItem` | `function()`               | selects the item that is currently highlighted                                                                   |
 
+### Autocomplete.Button
+
+This component renders a `button` tag and allows you to toggle the `Menu` component. It will also apply all of the proper ARIA attributes.
+
 ### Autocomplete.Menu
 
 This component allows you to render the items based on the user input. It must
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ This is called with an object with the properties listed below:
 
 ### Autocomplete.Button
 
-This component renders a `button` tag and allows you to toggle the `Menu` component. It will also apply all of the proper ARIA attributes.
+This component renders a `button` tag and allows you to toggle the `Menu` component. You can definitely build something like this yourself (all of the available APIs are exposed to you via the `Controller`), but this is nice because it will also apply all of the proper ARIA attributes.
 
 ### Autocomplete.Menu
 
--- a/examples/pages/semantic-ui/index.js
+++ b/examples/pages/semantic-ui/index.js
@@ -120,7 +120,7 @@ const Menu = glamorous.div({
   borderStyle: 'solid',
 })
 
-const ControllerButton = glamorous.button({
+const ControllerButton = glamorous(Autocomplete.Button)({
   backgroundColor: 'transparent',
   border: 'none',
   position: 'absolute',
@@ -145,18 +145,23 @@ function SemanticUIAutocomplete() {
     >
       <Autocomplete.Controller>
         {({isOpen, toggleMenu, clearSelection, selectedItem}) =>
-          (<Div
-            position="relative"
-            css={{display: 'flex', paddingRight: '1.75em'}}
-          >
+          (<Div position="relative" css={{paddingRight: '1.75em'}}>
             <Input
               getValue={i => i.name}
               isOpen={isOpen}
               placeholder="Enter some info"
             />
-            <Autocomplete.Button>
-              <ArrowIcon isOpen={isOpen} />
-            </Autocomplete.Button>
+            {selectedItem
+              ? <ControllerButton
+                css={{paddingTop: 4}}
+                onClick={clearSelection}
+                aria-label="clear selection"
+                >
+                <XIcon />
+              </ControllerButton>
+              : <ControllerButton>
+                <ArrowIcon isOpen={isOpen} />
+              </ControllerButton>}
           </Div>)}
       </Autocomplete.Controller>
       <Autocomplete.Menu defaultHighlightedIndex={0}>
--- a/examples/pages/semantic-ui/index.js
+++ b/examples/pages/semantic-ui/index.js
@@ -145,26 +145,18 @@ function SemanticUIAutocomplete() {
     >
       <Autocomplete.Controller>
         {({isOpen, toggleMenu, clearSelection, selectedItem}) =>
-          (<Div position="relative" css={{paddingRight: '1.75em'}}>
+          (<Div
+            position="relative"
+            css={{display: 'flex', paddingRight: '1.75em'}}
+          >
             <Input
               getValue={i => i.name}
               isOpen={isOpen}
               placeholder="Enter some info"
             />
-            {selectedItem
-              ? <ControllerButton
-                css={{paddingTop: 4}}
-                onClick={clearSelection}
-                aria-label="clear selection"
-                >
-                <XIcon />
-              </ControllerButton>
-              : <ControllerButton
-                onClick={toggleMenu}
-                aria-label={isOpen ? 'close menu' : 'open menu'}
-                >
-                <ArrowIcon isOpen={isOpen} />
-              </ControllerButton>}
+            <Autocomplete.Button>
+              <ArrowIcon isOpen={isOpen} />
+            </Autocomplete.Button>
           </Div>)}
       </Autocomplete.Controller>
       <Autocomplete.Menu defaultHighlightedIndex={0}>
--- a/src/button.js
+++ b/src/button.js
@@ -1,0 +1,92 @@
+import React, {Component} from 'react'
+import PropTypes from 'prop-types'
+
+import {AUTOCOMPLETE_CONTEXT} from './constants'
+import {compose} from './utils'
+
+class Button extends Component {
+  static contextTypes = {
+    [AUTOCOMPLETE_CONTEXT]: PropTypes.object.isRequired,
+  }
+
+  static propTypes = {
+    disabled: PropTypes.bool,
+    onKeyDown: PropTypes.func,
+    onClick: PropTypes.func,
+  }
+
+  constructor(props, context) {
+    super(props, context)
+    this.autocomplete = this.context[AUTOCOMPLETE_CONTEXT]
+    this.handleKeyDown = compose(this.handleKeyDown, props.onKeyDown)
+    this.handleClick = compose(this.handleClick, props.onClick)
+  }
+
+  keyDownHandlers = {
+    ArrowDown(event) {
+      event.preventDefault()
+      const amount = event.shiftKey ? 5 : 1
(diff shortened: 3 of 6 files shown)


examples/pages/semantic-ui/index.js
```diff
@@ -148,26 +148,18 @@ function SemanticUIAutocomplete() {
...
               isOpen={isOpen}
               placeholder="Enter some info"
             />
-            {selectedItem
-              ? <ControllerButton
-                css={{paddingTop: 4}}
-                onClick={clearSelection}
-                aria-label="clear selection"
-                >
-                <XIcon />
```
kentcdodds: Looks like we lost this functionality here. Can we keep the `XIcon` somehow?

src/button.js
```diff
@@ -0,0 +1,92 @@
...
+
+  constructor(props, context) {
+    super(props, context)
+    this.autocomplete = this.context[AUTOCOMPLETE_CONTEXT]
+    this.handleKeyDown = compose(this.handleKeyDown, props.onKeyDown)
+    this.handleClick = compose(this.handleClick, props.onClick)
+  }
+
+  keyDownHandlers = {
+    ArrowDown(event) {
```
kentcdodds: Oh nice! I was thinking about this! I was actually thinking that once the dropdown is expanded we should focus on the input instead. What do you think about that?

src/button.js
```diff
@@ -0,0 +1,92 @@
...
+      this.keyDownHandlers[event.key].call(this, event)
+    }
+  }
+
+  handleClick = event => {
+    event.preventDefault()
+    this.autocomplete.toggleMenu()
+  }
+
+  focusButton = () => {
```
kentcdodds: Should we expose this method in the `Controller`?

examples/pages/semantic-ui/index.js
souporserious: Oops! Yep, I was just testing it out real quick. I'll add it back.

src/button.js
souporserious: Yes :) I had that working for the Label component and then I started making this one 😆 I'll add it here.

src/button.js
souporserious: Yup 👍

souporserious:
Alright, I think this one is good to go now :) I removed the `focusButton` method because I don't think it will be necessary and would like to keep the API surface small if we can.

I also added a returnFocus method to restore the last focused item when the menu closes, and made sure the status menu is created and destroyed when the menu opens/closes.

README.md
```diff
@@ -120,6 +120,10 @@ This is called with an object with the properties listed below:
 | `selectItemAtIndex`     | `function(index: number)`  | selects the item at the given index                                                                              |
 | `selectHighlightedItem` | `function()`               | selects the item that is currently highlighted                                                                   |
 
+### Autocomplete.Button
```
kentcdodds: We should document the props as we do for the other components

examples/pages/semantic-ui/index.js
```diff
@@ -160,7 +160,6 @@ function SemanticUIAutocomplete() {
                 <XIcon />
               </ControllerButton>
               : <ControllerButton
-                onClick={toggleMenu}
                 aria-label={isOpen ? 'close menu' : 'open menu'}
```
kentcdodds: Can we remove this? The built in bottom should now handle all a11y for us right?

src/button.js
kentcdodds: If we make clicking this button automatically focus the input, then we won't need these event handlers right? Should we do that? Or is that too opinionated?

src/input.js
```diff
@@ -105,6 +107,7 @@ class Input extends Component {
         aria-expanded={isOpen}
         autoComplete="off"
         value={(inputValue === null ? selectedItemValue : inputValue) || ''}
+        autoFocus={this.menu}
```
kentcdodds: Interesting! What does this do? Should probably pass this to the boolean constructor to make this more explicit, but what's it for?

src/menu.js
```diff
@@ -103,31 +103,50 @@ class Menu extends Component {
...
-        getA11yStatusMessage={this.props.getA11yStatusMessage}
-        getInputValue={this.autocomplete.getInputValue}
-        getItemFromIndex={this.getItemFromIndex}
-        highlightedIndex={this.state.highlightedIndex}
-        inputValue={this.autocomplete.state.inputValue}
-        resultCount={this.items.length}
-      />,
-      this._statusNode,
-    )
+    this._statusNode &&
```
kentcdodds: I'd prefer an if statement here 👍

src/menu.js
```diff
@@ -103,31 +103,50 @@ class Menu extends Component {
...
   }
 
-  componentDidMount() {
-    this.autocomplete.setMenu(this)
-    this.autocomplete.emitter.on('menu:open', this.setDefaultHighlightedIndex)
+  handleMenuOpen = () => {
+    const lastActiveElement = document.activeElement
+    this.returnFocus = () => {
+      // only return focus if a non-focusable item was selected when closing
+      if (document.activeElement === document.body) {
```
kentcdodds: Hmmm, how does this work? Will this ever be true?

README.md
souporserious: There aren't any props 😁 we pull `disabled` off, but I think it's fine since that's a normal prop for button.

src/menu.js
souporserious: Yes as long as you don't click another focusable item this will return true. So imagine you click another menu button, you would want that to be focused and not return focus to the button that opened it. I hope that makes sense.

src/input.js
souporserious: It's just checking for the presence of `Menu` and then evaluates `autoFocus` to true. So no matter if `Button` opens it or not, it will autoFocus based on being a child of `Menu`.

README.md
kentcdodds: Ah, gotcha :+1: good call :)

src/input.js
kentcdodds: Gotcha. Do you mind wrapping it in `Boolean(this.menu)`? I just prefer to avoid type coercion in certain scenarios (this being one of them) :smile:

examples/pages/semantic-ui/index.js
souporserious: Done 👍

src/menu.js
kentcdodds: I see, so you're expecting that if they don't click on a focusable element, the body will receive focus (which is correct IIUC) :+1: Thanks!

souporserious:
Alright, now I _think_ it's all good to go :) pertaining to your question above (I couldn't reply for some reason), having the key handlers is beneficial to allow selecting elements from the list even when the input is not present. This is where I was saying it could be used for more than just an autocomplete 😇

src/button.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  constructor(props, context) {
+    super(props, context)
+    this.autocomplete = this.context[AUTOCOMPLETE_CONTEXT]
+    this.handleKeyDown = compose(this.handleKeyDown, props.onKeyDown)
+  }
+
+  keyDownHandlers = {
+    ArrowDown(event) {
```
kentcdodds: What did you think of this?

> If we make clicking this button automatically focus the input, then we won't need these event handlers right? Should we do that? Or is that too opinionated?

src/button.js
```diff
@@ -0,0 +1,87 @@
...
+    const {isOpen} = this.autocomplete.state
+    return (
+      <button
+        role="button"
+        aria-label={isOpen ? 'close menu' : 'open menu'}
+        onClick={this.handleClick}
+        {...rest}
+        aria-expanded={isOpen}
+        aria-haspopup={true}
+        disabled={disabled}
```
kentcdodds: If all we're doing is just forwarding this do we even need to specify it here? This would get applied with `{...this.props}` if we just did that instead.

src/button.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  render() {
+    const {disabled, ...rest} = this.props
+    const {isOpen} = this.autocomplete.state
+    return (
+      <button
+        role="button"
+        aria-label={isOpen ? 'close menu' : 'open menu'}
+        onClick={this.handleClick}
```
kentcdodds: This should go below the spread of props so it doesn't get overridden. We compose this handler so users will be able to specify a custom `onClick`.

src/button.js
```diff
@@ -0,0 +1,87 @@
...
+        {...rest}
+        aria-expanded={isOpen}
+        aria-haspopup={true}
```
kentcdodds: These `aria-` props should probably go above the spread.

src/input.js
```diff
@@ -21,6 +22,7 @@ class Input extends Component {
   constructor(props, context) {
     super(props, context)
     this.autocomplete = this.context[AUTOCOMPLETE_CONTEXT]
+    this.autoFocus = typeof this.context[MENU_CONTEXT] !== 'undefined'
```
kentcdodds: This changes the behavior from before... Now it will almost _always_ be `false` because the menu is often constructed _after_ the `input`. Could we just do `autoFocus={Boolean(this.menu)}`?

src/button.js
souporserious: Yup, totally. I realized I was using disabled for tabIndex, but we don't need that since it's a real button so I removed it which means we can just forward all props 🎉

src/button.js
souporserious: I think it's beneficial since it allows you to interact with the list of items. Similar to how react-aria-menubutton does when it receives focus.

src/button.js
souporserious: This was on purpose. Things like `clearSelection` won't work properly. Should we document that?

src/button.js
souporserious: Wasn't sure if they should be able to be overridden or not. I'll put them back 👍

src/input.js
souporserious: I _think_ this should be fine since this will only evaluate true if it's a child of `Menu` which means `Menu` needs to render first right?

kentcdodds: Ah! Gotcha! You're right!

src/button.js
kentcdodds: Sounds good 👍

kentcdodds:
```
🐑  $ npm publish --tag beta
+ react-autocompletely@1.0.0-beta.2
```

souporserious:
🙌 🎉  thanks so much! Going to start making examples now 😎

kentcdodds:
Thank YOU!

PR #50, opened 2017-07-28
Missing fixes in #46

dashed:
Closes https://github.com/paypal/react-autocompletely/issues/32

Missing fixes in https://github.com/paypal/react-autocompletely/pull/46

Reference: https://github.com/paypal/react-autocompletely/pull/46#discussion_r130191005

### DIFF
--- a/src/input.js
+++ b/src/input.js
@@ -2,7 +2,7 @@ import React, {Component} from 'react'
 import PropTypes from 'prop-types'
 
 import {AUTOCOMPLETE_CONTEXT} from './constants'
-import {compose, moveCursorToTheEnd} from './utils'
+import {compose} from './utils'
 
 class Input extends Component {
   static contextTypes = {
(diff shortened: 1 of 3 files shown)


src/input.js
```diff
@@ -2,7 +2,7 @@ import React, {Component} from 'react'
 import PropTypes from 'prop-types'
 
 import {AUTOCOMPLETE_CONTEXT} from './constants'
-import {compose, moveCursorToTheEnd} from './utils'
```
kentcdodds: Could you also remove this method from `utils`? It's not used anywhere else :smile:

src/input.js
```diff
@@ -65,10 +65,6 @@ class Input extends Component {
     this.updateInputValue(event.target.value)
   }
 
-  reset = () => {
```
kentcdodds: Could you make sure that we're not calling `.reset()` on the input anywhere else? I think we may be...

src/input.js
dashed: Resolved.

src/input.js
dashed: Resolved.

dashed:
Sorry I jumped the gun on the PR. 😅

PR #52, opened 2017-07-30
feat(better-components): major cleanup

souporserious:
**What**:
Soo I was I fixing a bug with the menu PR from yesterday and I ended up going down a little rabbit hole 😇 I realized since we're letting the user control if this is in a menu or not, we shouldn't be opinionated about the menu rendering at all and leave it up to the user. This meant moving the item methods up into `Autocomplete` which actually helped out a ton and allowed me to clean a lot of things up. Then I realized `MenuStatus` could just render in `Autocomplete` now which got rid of all of the weird portal stuff 🎉 

**Why**:
This reduces the API surface and cleans things up a ton. By moving the menu logic into the highest level, coordination was way easier and I was able to get rid of the `mitt` dependency 🎉 I also think this makes it very clear how to grab and use state now since you only do that with the `Controller` component. 

**How**:

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

I understand this is quite a bit of changes, so please ask me any questions or let me know what you would like changed. If you look at the basic example, it really didn't change to much on the surface. We just got rid of the `Menu` component in place of `Controller`. I think overall it will allow for more complex components to be built, but still keep it simple to use at the same time. If everything looks good, I'll update the rest of the examples and then it should be ready to merge.


package.json
```diff
@@ -18,12 +18,8 @@
   "keywords": [],
   "author": "Kent C. Dodds <kent@doddsfamily.us> (http://kentcdodds.com/)",
   "license": "MIT",
-  "dependencies": {
-    "mitt": "^1.1.2"
```
dashed: yay! 🎉

kentcdodds:
Awesome! I can't wait to give this a good review, but I think I love it. I'm still driving (eating dinner now), but I can hopefully look at it tonight.

So this gets rid of the menu component entirely and just has the controller? What if the autocomplete component was just the controller? So the child was a function that accepted state and helpers? I've considered doing this in the past. What do you think?

souporserious:
Sounds good! Glad you like it, I was a little nervous since it's a lot of changes 😅 Hope the drive has been going well!

I thought `Autocomplete` should just be `Controller` as well, but after using it I think it's nice having it in a component. I like the separation of API's, I feel like it's easier to use and can actually be nicer if you're building you're own compose-able components.

kentcdodds:
That makes sense. What if we allow for both? You might remember I had that implemented for a little while...

souporserious:
I personally like having them seperate. I think it promotes use of the `Controller` component, which I think can make it easier to tell what's going on. But, it's your library, so I'm totally cool with whatever you want to do :)

kentcdodds:
Nah, I agree with you. Let's keep it this way 👍 can add it later without breaking things if we want.

src/autocomplete.js
```diff
@@ -1,86 +1,116 @@
...
     [AUTOCOMPLETE_CONTEXT]: PropTypes.object.isRequired,
   }
 
   static propTypes = {
-    ref: PropTypes.func,
+    children: PropTypes.node.isRequired,
+    defaultHighlightedIndex: PropTypes.number,
+    getA11yStatusMessage: PropTypes.func,
     onChange: PropTypes.func.isRequired,
+    ref: PropTypes.func,
```
kentcdodds: Actually, we should remove this proptype because we can't consume the `ref` prop anyway.

src/autocomplete.js
```diff
@@ -1,86 +1,116 @@
...
+  }
+
+  static defaultProps = {
+    defaultHighlightedIndex: null,
   }
 
   constructor(...args) {
     super(...args)
     this.ref = compose(node => (this._rootNode = node), this.props.ref)
```
kentcdodds: Unfortunately we can't get access to the `ref` prop. A real shame, so we can remove `compose` and replace this with:

```
class .. blah blah {

  // no constructor necessary anymore

  ref = node => (this._rootNode = node)

}
```

src/autocomplete.js
```diff
@@ -235,9 +242,30 @@ class Autocomplete extends Component {
...
-    return <div {...rest} ref={this.ref} />
+    const {
+      children,
+      // eslint-disable-next-line no-unused-vars
+      defaultHighlightedIndex,
+      getA11yStatusMessage,
+      // eslint-disable-next-line no-unused-vars
+      onChange,
+      // eslint-disable-next-line no-unused-vars
+      ref,
```
kentcdodds: Can get rid of this reference here too because we can't get access to `this.props.ref` if we wanted to. A real shame.

src/autocomplete.js
```diff
@@ -235,9 +242,30 @@ class Autocomplete extends Component {
...
+      ...rest
+    } = this.props
+    return (
+      <div {...rest} ref={this.ref}>
+        {children}
+        <Status
```
kentcdodds: Nice 👍

src/button.js
```diff
@@ -1,7 +1,7 @@
 import React, {Component} from 'react'
 import PropTypes from 'prop-types'
 
-import {AUTOCOMPLETE_CONTEXT} from './constants'
+import AUTOCOMPLETE_CONTEXT from './context'
```
kentcdodds: Let's leave it as `constants`, just in case we want to have more in the future for some reason. I see no real reason to change it.

src/autocomplete.js
souporserious: I thought about this when I saw it, we should still do `innerRef` though right?

src/button.js
souporserious: Sounds good 👍 I kept getting an eslint warning because it wanted it to be a single default export so I changed it to one export. I'll put it back though. Should I just add an ignore comment or something?

src/autocomplete.js
kentcdodds: Sure! Not a bad idea :+1:

kentcdodds: Just don't forget to document it ;-)

src/button.js
kentcdodds: Yeah, just add an ignore. And you can file an issue on eslint-config-kentcdodds to remove all the dumb rules I have enabled :P (seriously, been planning on disabling many of them recently)

souporserious:
Working on getting them in tonight 👍 just need to fix a bug in `scrollIntoView` and it should be good to go.

souporserious:
I added a `component` and `innerRef` prop to allow for easier integration with existing systems. I personally need to use things like `<Autocomplete.Button component={UIKit.Button} .../>` I'm open to some other way to allow this, maybe all elements accept a `render` prop and you pass the `ref` down? We do something similar in [react-popper](https://github.com/souporserious/react-popper#usage-w-child-function). Anyways, I'll have more time for this tomorrow so I can wrap whatever up :)

kentcdodds:
```
🦄  $ npm publish --tag beta
+ react-autocompletely@1.0.0-beta.3
```

souporserious:
Thanks!!! 🎉

PR #56, opened 2017-07-30
add TS definitions

vutran:
Initial draft of TS definitions as I'm planning to expand this more as I begin porting [`omnibar`](http://omnibar.now.sh/) over to use the primitive components offered by `react-autocompletely`.


vutran:
Of course. No expectations on your end. Hopefully the community can help build the TS definitions up alongside Flow in the other PR.

kentcdodds:
Awesome! Thanks! If you don't mind, I'd like to wait until the API has settled a bit before we merge TypeScript typings. I don't want you to have to deal with at churn while we figure out what's best 👍

kentcdodds:
Alright! I'm pretty confident the API has solidified, so if you'd like to update this PR that would be awesome :smile:

Thanks!

vutran:
@kentcdodds Not sure how you want tests for these definitions, but I made a [repo here](https://github.com/vutran/befit/tree/react-downshift) with the example provided in the README.

kentcdodds:
For testing, maybe you could look at the glamorous repo. Maybe @luke-john could help guide this?

luke-john:
Most projects simply test definitions by having a single test file that covers the api. ie.

```
// ./typings/add-numbers.d.ts
declare module 'add-two-numbers' {
    type AddTwoNumbers = (x: number, y: number) => x + y

    export default const addTwoNumbers: AddTwoNumbers
}

// ./test/add-numbers.ts
import addTwoNumbers from 'add-two-numbers'

const test: number = addTwoNumbers(1, 5)
```

With glamorous we took things a bit further to try and decrease the chance of regressions.

In addition to the above style test we also have a `test/should-fail.ts` which goes through the api and attempts to break things. ie.
```
// ./test/should-fail.ts
import addNumbers from 'add-numbers'

const test: number = addTwoNumbers(1)
const test: number = addTwoNumbers(1)
const test: number = addTwoNumbers('1', 5)
const test: number = addTwoNumbers(1, '5')
const test: number = addTwoNumbers(1, 5, 1)
const test: number = addTwoNumbers(1, 5, 1)
```

Given the current size of the typings just having a single test file is probably suitable.

It would be good to add a ~`test:ts` script to the `package.json` which runs typescript with the noEmit flag to test these.

Also if these typings are likely to grow  in complexity you may want to consider landing them under a `typings` folder and adding `"typings": "typings/index.d.ts",`.  This should make it easier to break into seperate files without polluting the top level folder.

luke-john:
Just looking in the typings, it appears there's still a number of `any`s.  This is fine, however you may want to add a note to the README that the typescript typings are still incomplete and contain `any`s to help set expectations for other users.

In Glamorous we added the following to a `other/TYPESCRIPT_USAGE.md` file and linked to this from the `README.md`.

```
# Typescript Usage

The current bundled typescript definitions are incomplete and based around the needs of the developers who contributed them.

Pull requests to improve them are welcome and appreciated. If you've never contributed to open source before, then you may find [this free video course](https://egghead.io/courses/how-to-contribute-to-an-open-source-project-on-github) helpful.
```

index.d.ts
```diff
@@ -0,0 +1,105 @@
...
+
+        // state
+        highlightedIndex: number;
+        inputValue: string;
+        isOpen: boolean;
+        selectedItem: any;
+    }
+
+    type ChildrenFunction = (options: ControllerStateAndHelpers) => React.ReactNode;
+    interface Downshift extends React.ComponentClass<Downshift.Props> { }
```
luke-john: Having an interface named `Downshift` inside a `Downshift` namespace which passes the namespaces Props as `Downshift.Props` seems like it may cause confusion.

It may be simpler to keep everything at the top level. ie.

```
interface Props {
 ...
}

const downshift: React.ComponentClass<Props>

export default downshift
```

vutran: Thanks for the suggestions! Will take note and address them :)

vutran:
> Most projects simply test definitions by having a single test file that covers the api. ie.

I was thinking of creating some test files based on the given examples but since it wouldn't make sense to place it into `src/__tests__`. I'm thinking `src/__ts_tests__` but that may be a bit confusing to some users.

> It would be good to add a ~test:ts script to the package.json which runs typescript with the noEmit flag to test these.

Will do.

> Just looking in the typings, it appears there's still a number of anys. This is fine, however you may want to add a note to the README that the typescript typings are still incomplete and contain anys to help set expectations for other users.

I left the `any` types for items to start off with but will gladly introduce a generic for items.

kentcdodds:
Thanks @luke-john!!

@vutran, if you could put the typescript test script in `package-scripts` under `test` that would be optimal. Then we can add it to the `validate` script to make it run concurrently :+1:

vutran:
@kentcdodds Just a quick update, will try to find some time this week to address the requested chances. (Apologies, trying to recover from outsidelands and work-related stuff in the same time).

kentcdodds:
Sounds good. Thank you!

vutran:
Okay, I think this is ready for a review. @luke-john 

- Removed `Downshift` namespace and have everything under the root namespace.
- Export all interface to allow importing during usage.
- Added `tsc --noEmit` test to `package-scripts` and added to the `validate` nps script.
- Added `other/TYPESCRIPT_USAGE.md`
- Moved typings to `typings/index.d.ts`

luke-john:
These are looking good to me @vutran. Great work 👍.

It looks like this project is using rollup in a similar way to glamorous so you may hit an issue with typescript not playing nice with the rollup compilation.  If so the following addition to the entry file should solve things. https://github.com/paypal/glamorous/pull/104

kentcdodds:
Awesome! We're not on auto-release yet, but I'll get this published soon. Thanks so much @vutran! :tada:

vutran:
> It looks like this project is using rollup in a similar way to glamorous so you may hit an issue with typescript not playing nice with the rollup compilation. If so the following addition to the entry file should solve things. paypal/glamorous#104

I can open a new PR for this if needed.

kentcdodds:
Feel free! I leave the TS stuff entirely up to other folks since I don't really use it 🤷‍♂️

kentcdodds:
```
🏎  $ npm publish --tag rc
+ downshift@1.0.0-rc.15
```

Let's try this out!

PR #60, opened 2017-07-30
add action prop to Button

souporserious:
**What**:
This adds an `action` prop to the `Button` component.

**Why**:
After messing with the example I realized `clearSelection` didn't work and figured we should bake this in since there is some proper coordination that needs to go on with key handlers and click events. This also widens the pit of success for accessibility 🎉 

**How**:
I added a map of how each action should be handled and then call the respective action with the prop that was passed. I think after this we can explore even better accessibility and expose the labels for internationalization 😁 

**Checklist**:

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged 

### DIFF
--- a/src/button.js
+++ b/src/button.js
@@ -11,17 +11,38 @@ class Button extends Component {
 
   static propTypes = {
     component: PropTypes.any,
+    action: PropTypes.oneOf(['open', 'close', 'toggle', 'clearSelection']),
     onKeyDown: PropTypes.func,
   }
 
   static defaultProps = {
     component: 'button',
+    action: 'toggle',
   }
 
   constructor(props, context) {
     super(props, context)
     this.autocomplete = this.context[AUTOCOMPLETE_CONTEXT]
     this.handleKeyDown = compose(this.handleKeyDown, props.onKeyDown)
+  }
+
+  actionTypes = {
+    open: {
+      label: 'open menu',
+      action: () => this.autocomplete.openMenu(),
+    },
+    close: {
+      label: 'close menu',
+      action: () => this.autocomplete.closeMenu(),
+    },
+    toggle: {
+      label: isOpen => (isOpen ? 'close menu' : 'open menu'),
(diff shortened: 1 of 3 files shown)


kentcdodds:
Thanks for this! After considering it I would like to come up with a different approach. I really don't like strings that have any significant meaning in a codebase. I think it makes for a fragile API. So instead what if you could do:

```js
action={Autocomplete.Button.open}
```

That would allow you to do:

```js
action={() => {
  Autocomplete.Button.open()
  console.log('opened')
}}
```

And in that case, what if we switch `action` with `onClick`:

```js
onClick={Autocomplete.Button.open}
```

And then what if we just expose the actions in the controller?

```javascript
<Autocomplete.Controller>
{({open}) => (
  <Autocomplete.Button onClick={open}>Blah</Autocomplete.Button>
)}
</Autocomplete.Controller>
```

Oh! I think that this already exists! :wink:

What do you think?

souporserious:
Totally. I was trying to figure out if there was an easier way to promote accessibility. Maybe just require an `aria-label`?

kentcdodds:
Yeah, that's probably best. Just have a prop-types for it maybe?

src/button.js
```diff
@@ -38,11 +59,13 @@ class Button extends Component {
     },
 
     Enter(event) {
-      event.preventDefault()
-      if (this.autocomplete.state.isOpen) {
-        this.autocomplete.selectHighlightedItem()
-      } else {
-        this.autocomplete.openMenu()
+      if (this.props.action !== 'clearSelection') {
```
souporserious: Any ideas how to handle this condition if we aren't using an action prop?

souporserious: @kentcdodds any ideas how to handle this condition if we aren't using an action prop?

souporserious:
This seems too opinionated and the API is weird like you said so I'm closing this.

PR #64, opened 2017-08-01
refactor(remove-item-prop): removed item prop in favor of indexOf

souporserious:
**What**:
Related to https://github.com/paypal/react-autocompletely/issues/63

**Why**:
Just trying to keep the API small if we can :)

**How**:
Removed the index prop in favor of a method in autocomplete that uses the instance and `indexOf` to determine the items index.

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Ready to be merged 

If this looks good, just let me know and I will update all of the examples 😎

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ Called when the user selects an item
 
 > `string`/`Component` | defaults to `'div'`
 
-This allows you to controll what is rendered as the root element.
+This allows you to control what is rendered as the root element.
 
 #### innerRef
 
@@ -193,13 +193,6 @@ This component renders a `button` tag and allows you to toggle the `Menu` compon
 
 Render your items inside this component. This renders a `div` and forwards all
 props.
-
-#### index
-
-> `number` | *required*
-
-this is how `react-autocompletely` keeps track of your item when updating the
-`highlightedIndex` as the user keys around.
 
 #### value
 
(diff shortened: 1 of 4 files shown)


kentcdodds:
I remember trying to do this and there was a problem with when things get resorted (the index of item instances doesn't change, but the order of them in the DOM does). If you sift through the recording you'll see when I had to deal with this problem. `index` must stay I'm afraid.

souporserious:
Dang 😞 I totally remember now. Any ideas on how to handle sections and passing the `index` easier? I'm working with something like this..

```
<Select>
  <Optgroup>
    <Option index={0}>...</Option>
    <Option index={1}>...</Option>
  </Optgroup>
  <Optgroup>
    <Option index={2}>...</Option>
    <Option index={3}>...</Option>
  </Optgroup>
</Select>
```

kentcdodds:
What if we make it optional? Might be a footgun though... 🤔

souporserious:
If you use `index` as your key I think it will always be in the same spot 🤔 could we remove the need for it and somehow require your key to be an index?

kentcdodds:
Unfortunately, react doesn't give us access to the `key` prop we put on elements (similar to the `ref` prop). So we can't access that :-(

souporserious:
Dang. Maybe making the `index` optional like you said and mentioning in the docs how controlling it yourself is beneficial in some cases might work?

PR #68, opened 2017-08-01
WIP: new more flexible API

### LINKED ISSUE #62: More flexibility, less API?
What do folks think about doing something like this:

```javascript
const ui = (
  <Autocomplete getValue={a => a.title}>
    {({getInputProps, getItemProps, isOpen}) =>
      <div>
        <input
          {...getInputProps({
            placeholder: 'hello world',
            ref: node => (this.myOwnReference = node),
          })}
        />
        {isOpen
          ? <div>
              {items.map(i =>
                <article {...getItemProps({key: i.id})}>
                  {i.articleDetails}
                </article>
              )}
            </div>
          : null}
      </div>}
  </Autocomplete>
)
```

The entire API consists of a handful of props you pass to `Autocomplete` and a handful of arguments you receive in that callback. No need for props like `component` because you render the component yourself and you call a function to get the props to apply to the component so it hooks up properly with the autocomplete. In addition, there's no need of using the `context` API anymore which make the implementation a little nicer.

I'm actually pretty excited by this idea (originally inspired by @jaredly a week ago or so). I tried it and it didn't work very well, but now I think I know how to make it work and I think it'll be great. Minimizing the API and enhance the flexibility/simplicity of the thing. I'm probably going to try to implement this as soon as I can. Just wanted to hear your thoughts :smile:

kentcdodds:
closes #62
No time to talk, please review. Will be back to add more context/comments


src/set-a11y-status.js
```diff
@@ -0,0 +1,47 @@
...
+    overflow: 'hidden',
+    padding: '0',
+    position: 'absolute',
+    width: '1px',
+  })
+  document.body.appendChild(statusDiv)
+  return statusDiv
+}
+
+export default setStatus
```
souporserious: LOVE THIS

souporserious:
Everything looks solid from what I can see 💯 I reallllly like this approach, seems super nice to integrate with other libraries and existing components. I feel like we can just expose utilities to help with whatever and it gives the end user full flexibility on how they want to use it.

src/autocomplete.js
```diff
@@ -1,93 +1,78 @@
...
-        itemIndex = index
-      }
-    })
-    return itemIndex
+  getIndexFromItem = itemValue => {
+    const item = this.items.find(({value}) => value === itemValue)
+    return item ? item.index : null
+  }
+
+  getItemNodeFromIndex = index => {
```
souporserious: 🙌

README.md
```diff
@@ -106,7 +116,7 @@ Available components and relevant props:
 
 ### Autocomplete
 
-This is the main component. It renders a `div` and forwards props. Wrap
+This is the only component. It renders a `div` and forwards props. Wrap
```
kentcdodds: :tada:

souporserious:
Functions FTW 🎉

kentcdodds:
```
😄  $ npm publish --tag beta
+ react-autocompletely@1.0.0-beta.5
```

src/autocomplete.js
```diff
@@ -267,43 +396,48 @@ class Autocomplete extends Component {
...
-        <Status
-          getA11yStatusMessage={getA11yStatusMessage}
-          getItemFromIndex={this.getItemFromIndex}
-          getValue={this.getValue}
-          highlightedIndex={this.state.highlightedIndex}
-          resultCount={this.items.length}
-          value={this.state.value}
-        />
-      </AutocompleteComponent>
+      <div {...rest} ref={this.rootRef} onClick={this.handleClick}>
```
souporserious: My bad, I missed this. Is there any reason we can't just return `rootRef` and somehow get around the `onClick` stuff? I'd love to be able to use a custom component here.

kentcdodds: Well, the problem is that if we don't render a div there's no reasonable way for us to find the `_rootNode` for some of the mousedown/up handlers we have. I think that rendering a single div is a nice tradeoff. What else would you render?

PR #70, opened 2017-08-02
Only return child function

souporserious:
**What**:
Removes the last opinion about the DOM. 

**Why**:
I already hit a snag where I wanted a custom top level component to integrate better in my UI kit and I think this might be better since it follows suite to the other functions like `getItemProps`.

**How**:
I was able to get around the `ref` issue by just allowing a prop called `refKey` to be passed when you need to forward the ref in a non-traditional way like `{ refKey: 'innerRef' }`. I'm not tied to the name of the prop, maybe `getRef` or something? After doing this I'm wondering if we can go back to just using refs for the other components. It's up to you though :)

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -1,6 +1,6 @@
 /* eslint camelcase:0 */
 
-import React, {Component} from 'react'
+import {Component} from 'react'
 import PropTypes from 'prop-types'
 import setA11yStatus from './set-a11y-status'
 import {cbToCb, composeEventHandlers, debounce, scrollIntoView} from './utils'
@@ -174,6 +174,7 @@ class Autocomplete extends Component {
       selectItemAtIndex,
       selectHighlightedItem,
       setHighlightedIndex,
+      getRootProps,
       getButtonProps,
       getInputProps,
       getItemProps,
@@ -191,6 +192,7 @@ class Autocomplete extends Component {
       openMenu,
       closeMenu,
       clearSelection,
+      getRootProps,
       getButtonProps,
       getInputProps,
       getItemProps,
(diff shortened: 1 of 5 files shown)


src/autocomplete.js
```diff
@@ -420,25 +431,8 @@ class Autocomplete extends Component {
...
-      onChange,
-      ...rest
-    } = this.props
-    return (
-      <div {...rest} ref={this.rootRef} onClick={this.handleClick}>
-        {children(this.getControllerStateAndHelpers())}
-      </div>
-    )
+    const {children} = this.props
+    return children(this.getControllerStateAndHelpers())
```
kentcdodds: This is great! I'm going to update this a little bit to super both APIs 👍

souporserious: Did you mean support both? Curious why not just one way? I always like one way to do things, if possible, since it's easier to teach 🤓

kentcdodds: I generally agree with you, but I think that `getRootProps` will only be necessary in some use cases, and I'd prefer to not require folks to use it :smile:

kentcdodds:
What do you think @souporserious?

kentcdodds:
I'm feeling pretty great about this API. So I'm going to merge this, publish a new beta, and then I'm going to start working on tests! :tada:

Feel free to comment. We can always make more changes :smile:

kentcdodds:
```
🦄  $ npm publish --tag beta
+ react-autocompletely@1.0.0-beta.6
```

souporserious:
Yay 🎉 thanks! I'm going to work on the multiple values stuff now 😁

PR #72, opened 2017-08-02
docs(contributors): add some folks who helped inspire things

kentcdodds:
**What**: Add @ryanflorence and @jaredly to the contributors:

screen shot 2017-08-02 at 11 12 53 am

Unfortunately, `all-contributors` doesn't have an `ideas` contribution type which is a shame. Something I might want to add in the future.

**Why**: Because they both shared feedback and inspired ideas that led to what we have now.

**How**: `all-contributors-cli`

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [x] Ready to be merged 

Ryan and Jared, are you both cool with this? Totally fine if you'd rather not be on the list.


kentcdodds:
I went ahead and merged this. @ryanflorence and @jaredly, if you'd rather not be on there, just let me know and I'll take you off. Just wanna show gratitude and give credit where it's due!

PR #74, opened 2017-08-03
feat(multiple-prop): added the ability to select multiple values

souporserious:
**What**:
Related to https://github.com/paypal/react-kadabra/issues/61

**Why**:
So we can build more complex menus that allow selecting multiple values.

**How**:
I followed how the browser handles `multiple` and `value`. When `multiple` is `true`, we handle value as an array. I switched the API from `selectedItem` -> `value` because I think this is a little more clear since you are technically selecting the value of your item[s]. This is also how other select menus like react-select handle this so it seemed like a safe change. I changed `value` -> `inputValue` which I think is more clear as well.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

I was able to build a select menu, a multi-select, an autocomplete, and a simple dropdown menu with these changes. It's pretty cool how powerful, but simple it is! I _think_ this is closing in on a pretty final API after this. But please, let me know what you think 😇  if all looks well, I'll update the rest of the examples. It's a pretty minimal breaking change. Mainly `value` -> `inputValue`.

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -289,11 +289,9 @@ class Autocomplete extends Component {
   //////////////////////////// BUTTON
 
   buttonKeyDownHandlers = {
-    ArrowDown: this.keyDownHandlers.ArrowDown,
+    ...this.keyDownHandlers,
 
-    ArrowUp: this.keyDownHandlers.ArrowUp,
-
-    Enter(event) {
+    ' '(event) {
       event.preventDefault()
       if (this.state.isOpen) {
         if (this.state.highlightedIndex === null) {
@@ -304,11 +302,6 @@ class Autocomplete extends Component {
       } else {
         this.openMenu()
       }
-    },
-
-    ' '(event) {
-      event.preventDefault()
-      this.toggleMenu()
     },
   }
 
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -292,10 +292,10 @@ class Autocomplete extends Component {
     Enter(event) {
       event.preventDefault()
       if (this.state.isOpen) {
-        if (this.state.highlightedIndex !== null) {
-          this.selectHighlightedItem()
-        } else {
+        if (this.state.highlightedIndex === null) {
           this.closeMenu()
+        } else {
+          this.selectHighlightedItem()
         }
       } else {
         this.openMenu()
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -8,18 +8,18 @@ import {cbToCb, composeEventHandlers, debounce, scrollIntoView} from './utils'
 class Autocomplete extends Component {
   static propTypes = {
     children: PropTypes.func.isRequired,
-    defaultSelectedItem: PropTypes.any,
     defaultHighlightedIndex: PropTypes.number,
+    defaultValue: PropTypes.any,
     getA11yStatusMessage: PropTypes.func,
     getValue: PropTypes.func,
+    multiple: PropTypes.bool,
     onChange: PropTypes.func.isRequired,
     onClick: PropTypes.func,
+    value: PropTypes.any,
   }
 
   static defaultProps = {
-    defaultSelectedItem: null,
     defaultHighlightedIndex: null,
-    getValue: i => String(i),
     getA11yStatusMessage({resultCount, highlightedItem, getValue}) {
       if (!resultCount) {
         return 'No results.'
@@ -30,15 +30,16 @@ class Autocomplete extends Component {
       }
       return getValue(highlightedItem)
     },
+    getValue: i => String(i),
   }
 
   constructor(...args) {
     super(...args)
     this.state = {
       highlightedIndex: null,
-      value: null,
+      inputValue: '',
       isOpen: false,
-      selectedItem: this.props.defaultSelectedItem,
+      value: this.props.defaultValue || this.props.multiple ? [] : '',
     }
     this.root_handleClick = composeEventHandlers(
       this.props.onClick,
@@ -58,7 +59,7 @@ class Autocomplete extends Component {
     })
   }
 
-  getIndexFromItem = itemValue => {
+  getIndexFromValue = itemValue => {
     const item = this.items.find(({value}) => value === itemValue)
     return item ? item.index : null
   }
@@ -84,7 +85,7 @@ class Autocomplete extends Component {
   }
 
   highlightSelectedItem = () => {
-    const highlightedIndex = this.getIndexFromItem(this.state.selectedItem) || 0
+    const highlightedIndex = this.getIndexFromValue(this.state.value) || 0
     this.setState({highlightedIndex}, () => {
       this.maybeScrollToHighlightedElement(highlightedIndex, true)
     })
@@ -123,8 +124,7 @@ class Autocomplete extends Component {
   clearSelection = () => {
     this.setState(
       {
-        selectedItem: null,
-        value: null,
+        value: this.multiple ? [] : '',
         isOpen: false,
       },
       () => {
@@ -137,9 +137,24 @@ class Autocomplete extends Component {
   }
 
   selectItem = itemValue => {
-    this.reset()
+    if (!this.props.multiple) {
+      this.reset()
+    }
     this.setState(
-      {selectedItem: itemValue, value: this.getValue(itemValue)},
+      state => {
+        if (this.props.multiple) {
+          const values = [...state.value]
+          const pos = values.indexOf(itemValue)
+          if (pos > -1) {
+            values.splice(pos, 1)
+          } else {
+            values.push(itemValue)
+          }
+          return {value: values}
+        } else {
+          return {value: itemValue, inputValue: this.getValue(itemValue)}
+        }
+      },
       () => {
         this.props.onChange(itemValue)
       },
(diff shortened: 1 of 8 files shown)


src/autocomplete.js
```diff
@@ -8,18 +8,18 @@ import {cbToCb, composeEventHandlers, debounce, scrollIntoView} from './utils'
...
     getA11yStatusMessage: PropTypes.func,
     getValue: PropTypes.func,
+    multiple: PropTypes.bool,
     onChange: PropTypes.func.isRequired,
     onClick: PropTypes.func,
+    value: PropTypes.any,
   }
 
   static defaultProps = {
-    defaultSelectedItem: null,
```
kentcdodds: Should this be `defaultValue: null`? Even though it doesn't actually make a difference, I like having defaults set for anything that's not required.

src/autocomplete.js
```diff
@@ -164,38 +179,43 @@ class Autocomplete extends Component {
...
+      toggleMenu,
       selectItem,
       selectItemAtIndex,
       selectHighlightedItem,
       setHighlightedIndex,
-      getRootProps,
+      clearSelection,
+    } = this
+    return {
+      // prop getters
```
kentcdodds: Thank you!

src/autocomplete.js
```diff
@@ -164,38 +179,43 @@ class Autocomplete extends Component {
...
-      clearSelection,
+      toggleMenu,
       selectItem,
       selectItemAtIndex,
       selectHighlightedItem,
       setHighlightedIndex,
-      getRootProps,
+      clearSelection,
+    } = this
+    return {
```
kentcdodds: Could we expose `getItemFromIndex` while we're at it? Ref: https://github.com/paypal/react-kadabra/issues/71

src/autocomplete.js
```diff
@@ -288,15 +308,19 @@ class Autocomplete extends Component {
...
-    const {value, selectedItem, isOpen} = this.state
+    const {isOpen} = this.state
     return {
       'data-autocomplete-input': true,
       role: 'combobox',
       'aria-autocomplete': 'list',
       'aria-expanded': isOpen,
       autoComplete: 'off',
-      value: (value === null ? this.getValue(selectedItem) : value) || '',
+      value: this.state.inputValue,
```
kentcdodds: Let's destructure this above. I like to destructure stuff at the top of the closure so it's easy to tell what this function uses.

src/autocomplete.js
```diff
@@ -308,14 +332,12 @@ class Autocomplete extends Component {
     if (event.key && this.keyDownHandlers[event.key]) {
       this.keyDownHandlers[event.key].call(this, event)
     } else if (!['Shift', 'Meta', 'Alt', 'Control'].includes(event.key)) {
-      this.openMenu(() => {
-        this.highlightIndex()
-      })
+      this.openMenu()
```
kentcdodds: Why this change?

src/autocomplete.js
```diff
@@ -338,11 +360,10 @@ class Autocomplete extends Component {
   //\\\\\\\\\\\\\\\\\\\\\\\\\\\\\ ITEM
 
   reset = () => {
-    this.setState(({selectedItem}) => ({
+    this.setState({
       isOpen: false,
       highlightedIndex: null,
-      value: this.getValue(selectedItem),
```
kentcdodds: The `inputValue` should be reset to the `getValue` of the currently selected item right?

src/autocomplete.js
```diff
@@ -371,10 +396,6 @@ class Autocomplete extends Component {
     this.toggleMenu(false, cb)
   }
 
-  getValue = item => {
```
kentcdodds: Why is this removed?

src/autocomplete.js
souporserious: done ✅

src/autocomplete.js
souporserious: Oops! On it.

src/autocomplete.js
souporserious: I moved that logic into `toggleMenu` so anytime the menu is opened it will handled highlighting.

src/autocomplete.js
souporserious: Oops, that was a mistake, putting it back 😁

souporserious: Ohh that wasn't a mistake. I handle that in `selectItem` now. I actually need to handle `multiple` and `inputValue`.

src/autocomplete.js
souporserious: I moved it closer to the input methods :)

src/autocomplete.js
souporserious: My bad. Done 👍

souporserious:
I added some other things I found 😁 it looks like the recent highlighting changes got lost in the new API from today. I added that back, as well as the button API we had before, which turned out nice because we now have access to the same handlers as Input uses 🎉

souporserious:
Do you think `value` should be `selectedValue` so it's a little more explicit? 🤔

kentcdodds:
> Do you think value should be selectedValue so it's a little more explicit? 🤔

Yes 👍

src/autocomplete.js
```diff
@@ -257,6 +285,30 @@ class Autocomplete extends Component {
 
   //////////////////////////// BUTTON
 
+  buttonKeyDownHandlers = {
+    ArrowDown: this.keyDownHandlers.ArrowDown,
+
+    ArrowUp: this.keyDownHandlers.ArrowUp,
+
+    Enter(event) {
```
kentcdodds: Actually, I think I prefer "Enter" on the button to select the highlighted item.

I'd love to hear what @marcysutton has to say on the subject of accessibility. Right now you can hit the spacebar on the button to toggle the menu and the up/down arrow keys to change the highlighted item, then the enter key to select an item. If we add this event handler, I feel like the Enter key has too much going on:

1. It can open the menu
2. It can close the menu
3. It can select the highlighted item.

I'd rather just have the enter key select the highlighted item if one is highlighted (like it is currently). [Here's the interaction](https://codesandbox.io/s/9rWEmzEv8) as it stands today: 

a11y

souporserious: Totally 👍 I did that wrong. I meant for space to handle that functionality :) that's seems to be how the default browser select works.

souporserious:
Alright. I _think_ I have everything done. I was wondering in `onChange` whether we should pass the `selectedValue` there as well for convenience. That way you can see what value was just selected/deselected as well as look at the current state of selections.

kentcdodds:
> I was wondering in onChange whether we should pass the selectedValue there as well for convenience.

Yes. This sounds like a good idea. Though what if an item is removed in a multi-select scenario?

souporserious:
This is why I think we almost need to pass it now. That way you can compare what was just selected with the current values to determine that. I ended up with [something like this for selectly](https://github.com/souporserious/selectly/blob/master/src/Select.jsx#L146-L150) that worked well at our work. I don't _think_ we need to pass as much as I was passing though.

kentcdodds:
Got it so the API would be:

`onChange(currentValue, previousValue)`?

kentcdodds:
I kinda like what you have there actually. Maybe we should do an object...

```js
onChange({selectedValue, previousValue})
```

souporserious:
I like that. I think that's enough info to do what you need 🤔 maybe this will be good enough for now and we can revisit this part if we need to.

kentcdodds:
I wonder if we could also provide what kind of change it was: `onChange({selectedValue, previousValue, type: Autocomplete.changeTypes.removal})`

And in `changeTypes` we'd have `removal`, `addition`

Would that be useful? Should we concern ourselves with that?

tannermares:
@souporserious: I'm torn on wether or not the select should be that smart too. I have needed that value a couple times but wasn't super annoyed I had to do the subtraction myself to figure it out. I'm afraid it'll start a trend of random helper props that just bloats things. I love where this is going either way!

kentcdodds:
> I'm afraid it'll start a trend of random helper props that just bloats things

Yeah, let's skip the `type` thing and just stick with `selectedValue` and `previousValue`. We can always add it later. Harder to remove it ;-)

souporserious:
Cool :) that's what I was thinking. It's always easier to add than remove.

souporserious:
I landed on this. Thoughts?
image

kentcdodds:
Cool. So `itemValue` would be the item that was just selected? What would this look like the context of a non-multiselect? What would `itemValue` be if an item is removed?

souporserious:
Haha I just deleted that comment 😆 I think I was doing pretty much what we talked about above. Let's see how `selectedValue` and `previousValue` play out.

src/autocomplete.js
marcysutton: Yep, space bar is the default for buttons.

souporserious:
Alright, I _think_ it's finally done 😁

src/autocomplete.js
```diff
@@ -137,11 +140,36 @@ class Autocomplete extends Component {
...
+          const values = [...state.selectedValue]
+          const pos = values.indexOf(itemValue)
+          if (pos > -1) {
+            values.splice(pos, 1)
+          } else {
+            values.push(itemValue)
+          }
+          return {
+            selectedValue: values,
+            inputValue: values.map(value => this.getValue(value)).join(', '),
```
kentcdodds: In the future I expect we'll need to make this a prop as well `getInputValue`... Either that or instead of mapping all the values to `this.getValue` we just call `this.getValue(values)`... Seems like that'd be reasonable... Should we do that now?

kentcdodds:
Sorry, I'm going to push forward with the rename. I want to get that done to avoid confusing people longer! Hopefully there will be no merge conflicts.

src/autocomplete.js
souporserious: Ah yes, that's true. The end user might want to handle arrays in a different way or maybe they wouldn't want to, I'm not sure 🤔

souporserious:
Looks like no merge conflicts 🙌

src/autocomplete.js
souporserious: Maybe ok to push through and see how people react? Keeps the API a little easier IMO

kentcdodds: Sounds fine.

kentcdodds:
I'll merge this now. Release coming when I finish with my late lunch

kentcdodds:
```
🐑  $ npm publish --tag beta
+ downshift@1.0.0-beta.9
```

PR #84, opened 2017-08-04
feat(stateProps): add the concept of state props

### LINKED ISSUE #80: Controlled vs uncontrolled
Right now this is totally uncontrolled (you can't pass the value in and can only get the value out). We should support controlled (passing the value in as well).

Someone brought up their use case to me in gitter and this is important I think.

kentcdodds:
This adds the concept of `state props` to enable a totally controlled component experience. It's actually quite awesome if I say so myself. 🕶

Feedback welcome!

Closes #80

**Checklist**:

- [x] Documentation
- [ ] Tests N/A (we don't have tests yet. Still solidifying the API)
- [x] Ready to be merged 

cc @jgoux, this is what you're looking for I think :smile: I'd really appreciate a review!

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,8 +122,8 @@ Available components and relevant props:
 
 ### Autocomplete
 
-This is the only component. It doesn't render anything itself, it just calls the child function and renders that. Wrap
-everything in this.
+This is the only component. It doesn't render anything itself, it just calls
+the child function and renders that. Wrap everything in this.
 
 #### getValue
 
@@ -163,6 +163,42 @@ A default `getA11yStatusMessage` function is provided that will check `resultCou
 > `function({selectedValue, previousValue})` | *required*
 
 Called when the user selects an item
+
+#### onStateChange
+
+> `function({highlightedIndex, inputValue, isOpen, selectedValue})` | not required, no useful default
+
+This function is called anytime the internal state changes. This can be useful
+if you're using downshift as a "controlled" component, where you manage some or
+all of the state (e.g. isOpen, selectedValue, highlightedIndex, etc) and then
+pass it as props, rather than letting downshift control all its state itself.
+
+#### highlightedIndex
+
+> `number` | **state prop** (read more below)
+
+The index that should be highlighted
+
+#### inputValue
+
+> `string` | **state prop** (read more below)
+
+The value the input should have
+
+#### isOpen
+
+> `boolean` | **state prop** (read more below)
+
+Whether the menu should be considered open or closed. Some aspects of the
+autocomplete component respond differently based on this value (for example, if
+`isOpen` is true when the user hits "Enter" on the input field, then the
+item at the `highlightedIndex` item is selected).
+
+#### `selectedValue`
+
+> `any`/`Array(any)` | **state prop** (read more below)
+
+The currently selected value.
 
 #### children
 
@@ -249,6 +285,17 @@ translations:
   'aria-label': translateWithId(isOpen ? 'close.menu' : 'open.menu'),
 })} />
 ```
+
+### State Props
+
+You can pass some props which normally the `downshift` will manage for you. If
+you pass these props, then they become "controlled" props. In this situation,
+`downshift` will no longer update them directly and will instead call your
+`onStateChange` handler and expect you to update them. This can be useful if
+you want to control the component externally (like selecting an item
+from another part of the UI).
+
+State Props are labeled above with **state prop**
 
 ## Examples
 
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -14,7 +14,16 @@ class Autocomplete extends Component {
     getValue: PropTypes.func,
     multiple: PropTypes.bool,
     onChange: PropTypes.func.isRequired,
+    onStateChange: PropTypes.func,
     onClick: PropTypes.func,
+    // things we keep in state for uncontrolled components
+    // but can accept as props for controlled components
+    /* eslint-disable react/no-unused-prop-types */
+    selectedValue: PropTypes.any,
+    isOpen: PropTypes.bool,
+    inputValue: PropTypes.string,
+    highlightedIndex: PropTypes.number,
+    /* eslint-enable */
   }
 
   static defaultProps = {
@@ -31,6 +40,13 @@ class Autocomplete extends Component {
       return getValue(highlightedItem)
     },
     getValue: i => String(i),
+    onStateChange: () => {},
+  }
+
+  // this is an experimental feature
+  // so we're not going to document this yet
+  static stateChangeTypes = {
+    mouseUp: '__autocomplete_mouseup__',
   }
 
   constructor(...args) {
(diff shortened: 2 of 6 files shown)


src/autocomplete.js
```diff
@@ -490,36 +572,49 @@ class Autocomplete extends Component {
...
+      // in a "controlled" component manner if you need to
+      selectedValue,
+      isOpen,
+      inputValue,
+      highlightedIndex,
       ...rest
     } = this.props
-    const element = children(this.getControllerStateAndHelpers())
-    if (this.getRootProps.called) {
+    // doing React.Children.only for Preact support ⚛️
```
kentcdodds: So I may have thrown in one or two extras in here (cc @zouhir @developit) :wink:

stories/examples/controlled.js
```diff
@@ -0,0 +1,199 @@
+import React, {Component} from 'react'
```
kentcdodds: Pull this down and run it. It's actually a pretty slick experience.

```
git clone https://github.com/paypal/downshift.git
cd downshift
npm install
npm run storybook
```

Then open `http://localhost:6006` and choose `Examples => controlled` :+1:

README.md
```diff
@@ -164,6 +164,41 @@ A default `getA11yStatusMessage` function is provided that will check `resultCou
 
 Called when the user selects an item
 
+#### onStateChange
+
+> `function({highlightedIndex, inputValue, isOpen, selectedValue})` | not required, no useful default
+
+This function is called anytime there's a state change internally. This can be
+useful if you're using the component as a "controlled" component (passing the
+state in rather than letting the component control the state itself).
```
BurntCaramel: I think this is worded in a confusing way. ‘The component’ in this case is the Downshift component? But sounds like it could be talking about your component too.

Maybe:
“This function is called anytime the internal state changes. This can be useful if you're using Downshift as a "controlled" component, where you manage the state (e.g. `isOpen`, `selectedValue`, `highlightedIndex`, etc) and then pass it as props, rather than letting Downshift control all its state itself.”

sprjr:
I'm not familiar with this repo, or concept, so just skimming here. I'm curious why the *state props* concept though. It seems like a bit of overhead to grok. I feel like you could (maybe naively) simply add a single prop, like *stateUpdateTransform*, with a default value that modifies state in the non-controlled fashion. Then, if users want, they can simply override a single function, which is already the recommended `onStatecChange` prop here if they don't want the default behavior.

Maybe I'm just getting hung up on the naming of state props? It seems like it's some weird mixture of two concepts that already work so nicely for components.

src/autocomplete.js
```diff
@@ -50,6 +58,13 @@ class Autocomplete extends Component {
   input = null
   items = []
 
+  getState() {
+    return Object.keys(this.state).reduce((state, key) => {
+      state[key] = this.props[key] == null ? this.state[key] : this.props[key]
+      return state
+    }, {})
+  }
```
johannes-scharlach: When I saw this, I was first wondering if there was maybe an elegant way of generally defaulting to `state` if a prop wasn't available in `props` and what I had in mind could be something like this:

```js
getPropsOrState() {
  return {
    ...state,
    ...props
  }
}
```
This nicely overwrites with `props` if defined (the details are a little different to this method, because if a prop is available and set to `null` it would still overwrite the state, but I believe that's what a user would expect)

src/autocomplete.js
```diff
@@ -187,11 +193,82 @@ class Autocomplete extends Component {
...
+        // if the selectedValue changed
+        // then let's call onChange!
+        if (Object.keys(onChangeArg).length) {
+          this.props.onChange(onChangeArg)
+        }
+        // We call this function whether we're controlled or not
+        // It's mostly useful if we're controlled, but it can
+        // definitely be useful for folks to know when something
+        // happens internally.
+        this.props.onStateChange(onStateChangeArg)
```
johannes-scharlach: I'm wondering why you're doing this in the callback rahter than within the `setState` call... I imagine this would lead to a change of props right after we've changed the state, so two rerenders, whereas doing this within `setState` would allow react to merge the state and prop changes together and all be executed in one rerender.

I can elaborate some more if you'd like me to give more context? 😸

README.md
kentcdodds: Great suggestion! Will use that. Thanks!

src/autocomplete.js
kentcdodds: Great suggestion, really simplifies things. Though I would prefer to not spread all the props on that object if I can avoid it. The only use case I care about is the state for this function.

But you do make a great point here:

> This nicely overwrites with props if defined (the details are a little different to this method, because if a prop is available and set to null it would still overwrite the state, but I believe that's what a user would expect)

So I'm going to change that to use `this.props[key] === undefined` so the user can set it to null if they wish.

src/autocomplete.js
kentcdodds: Thanks for the suggestion there. I worry that doing that could lead to some interesting effects (in a negative way). What we're doing when we call these functions is saying: "The state has changed" however, if we do it in the `setState` call, then we're saying: "The state will probably change in the future"

I don't believe that React actually makes any guarantee about what the state will end up looking like after our state changes due to batching `setState` calls. So I think it'll be better to leave it as is. Thanks though!

src/autocomplete.js
```diff
@@ -31,6 +40,13 @@ class Autocomplete extends Component {
       return getValue(highlightedItem)
     },
     getValue: i => String(i),
+    onStateChange: () => {},
+  }
+
+  // this is an experimental feature
+  // so we're not going to document this yet
+  static stateChangeTypes = {
```
kentcdodds: See the controlled example for a use case for this.

src/autocomplete.js
```diff
@@ -489,33 +591,23 @@ class Autocomplete extends Component {
...
-      getA11yStatusMessage,
-      // eslint-disable-next-line no-unused-vars
-      defaultHighlightedIndex,
-      // eslint-disable-next-line no-unused-vars
-      multiple,
-      // eslint-disable-next-line no-unused-vars
-      onClick,
-      // eslint-disable-next-line no-unused-vars
-      onChange,
-      ...rest
```
kentcdodds: Decided we don't need to forward props. Because you're rendering the child component/element yourself, you can apply the props you want directly to that element.

stories/examples/controlled.js
```diff
@@ -0,0 +1,207 @@
...
+  items = ['Black', 'Red', 'Green', 'Blue', 'Orange', 'Purple']
+
+  changeHandler = changes => {
+    let {
+      selectedValue = this.state.selectedColor,
+      isOpen = this.state.isOpen,
+      inputValue = this.state.inputValue,
+      type,
+    } = changes
+    isOpen =
```
kentcdodds: Here's where the `type` comes in handy. This allows us to know whether the open/close happened due to a mouseup interaction, in which case we want to leave it open because we're controlling that state ourselves. Otherwise you'd click the button and that would count as a mouseup.

kentcdodds:
```
🚀  $ npm publish --tag beta
+ downshift@1.0.0-beta.11
```

src/autocomplete.js
johannes-scharlach: Hmmm, if I'm following your argument correctly, you're trying to avoid passing anything to the parent that was never actually the state of this component.

However we're doing exactly that: passing properties to the parent that maybe never actually were part of the state.

Also: if this is a totally controlled component, it's possible that we never actually change the state, right?

Currently I'm not sure if state and props give us all the tools necessary to avoid such interesting effects and we'd need to keep the "state" in a separate property of the component. Does that make sense?

kentcdodds: > However we're doing exactly that: passing properties to the parent that maybe never actually were part of the state.

Yes, that is what we're doing, however, downshift knows when those things should be changed (the user hits the down arrow on the input for example) and we need to tell the parent that the `isOpen` state should change because downshift can't control that value itself.

If you pull down the repo and try out the example you'll see that this API gives us everything we need and no more.

Issue #79, opened 2017-08-03
Accessibility feedback


marcysutton:
Here are some things I found when I tested the component!

## What Works
- It works great from the keyboard. Very intuitive, and the clear results button is actionable. 👍
- Great job announcing the number of items available!

## What Needs Work
- [x] The input needs a label, as `placeholder` isn't reliable in every assistive technology. You could use a unique `aria-label`, or ideally, a `<label>` element with text wrapping the input (which increases the hit area).
- [x] In Voiceover, you can't navigate with the arrow keys alone so the directions announcing "navigate with the up and down arrow keys" is inaccurate. You could try adding `aria-activedescendant` to the INPUT pointing to the ID of the currently highlighted item as you navigate. Here's a [similar thread](https://github.com/accdc/tsg/issues/1) you could reference.
- [x] When an item is selected, the a11y-status-message says "no results" (see attached screenshot). It should say which item is selected. Same thing in NVDA and Voiceover.

Note: I can't even get codesandbox.io to load in IE, so I was unable to test it with JAWS.

![no results when Anakin Solo is selected](https://user-images.githubusercontent.com/1045233/28936137-9c427dde-783b-11e7-818a-104269165812.png)

marcysutton:
Note: I wanted to test it on mobile, too, but I can't get codesandbox.io to work on a small screen!

kentcdodds:
Wow @marcysutton, this is super nice of you to take time out of your busy schedule to help make this accessible! Thank you very much!

On mobile you might try this URL: https://codesandbox.io/embed/9rWEmzEv8?hidenavigation=1&view=preview

You could also download [the example](https://codesandbox.io/s/9rWEmzEv8): 

<img width="266" alt="screen shot 2017-08-03 at 12 19 16 pm" src="https://user-images.githubusercontent.com/1500684/28936783-ff57802c-7845-11e7-8f47-0a1e9da58138.png">

Then cd into the directory and run `npm install && npm start`. That'll pull it up locally so you can test it in JAWS (which would be super awesome of you too).

Thank you again. We'll make sure to address these things before we release 1.0.0 :tada:

souporserious:
I think `activedescendant` should be relatively easy to add since we have `highlightedIndex`. I did some of this work in [react-aria](https://github.com/souporserious/react-aria/blob/master/src/Select/Option.jsx#L48). I'll try and get it done tomorrow 😁

kentcdodds:
[...]

PR #86, opened 2017-08-04
fix(a11y): improve `getA11yStatusMessage`

kentcdodds:
Ref: #79

**What**: Fixes some things in the default implementation of `getA11yStatusMessage`

**Why**: So the default is more accessible

**How**: Magic ✨

**Checklist**:

- [x] Documentation
- [ ] Tests N/a
- [x] Ready to be merged 

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -36,15 +36,31 @@ class Autocomplete extends Component {
   static defaultProps = {
     defaultHighlightedIndex: null,
     defaultValue: null,
-    getA11yStatusMessage({resultCount, highlightedItem, getValue}) {
+    // eslint-disable-next-line complexity
+    getA11yStatusMessage({
+      isOpen,
+      highlightedValue,
+      selectedValue,
+      resultCount,
+      previousResultCount,
+      getValue,
+    }) {
+      if (!isOpen) {
+        if (selectedValue) {
+          return getValue(selectedValue)
+        } else {
+          return ''
+        }
+      }
+      const resultCountChanged = resultCount !== previousResultCount
       if (!resultCount) {
         return 'No results.'
-      } else if (!highlightedItem) {
+      } else if (!highlightedValue || resultCountChanged) {
         return `${resultCount} ${resultCount === 1 ?
           'result is' :
           'results are'} available, use up and down arrow keys to navigate.`
       }
-      return getValue(highlightedItem)
+      return getValue(highlightedValue)
     },
     getValue: i => String(i),
     onStateChange: () => {},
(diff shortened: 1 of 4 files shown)


src/autocomplete.js
```diff
@@ -274,16 +291,16 @@ class Autocomplete extends Component {
       () => {
         // call the provided callback if it's a callback
         cbToCb(cb)()
-        // if the selectedValue changed
```
kentcdodds: As I was working with things it seemed to me to be a bit better to have this happen after `onStateChange` because `onStateChange` happens _all_ the time, so I felt it better to have `onChange` win (because it comes second) in the event of a state change conflict for the user.

src/autocomplete.js
```diff
@@ -607,15 +629,8 @@ class Autocomplete extends Component {
...
-    // TODO: what do we need to do for this when the
-    // autocomplete is a controlled component?
-    if (
-      prevState.highlightedIndex !== this.state.highlightedIndex ||
-      this.state.selectedValue !== prevState.selectedValue
-    ) {
-      this.updateStatus()
-    }
+  componentDidUpdate() {
+    this.updateStatus()
```
kentcdodds: Turns out we want to update the status of the state anytime there's an update to the state 😅

src/set-a11y-status.js
```diff
@@ -13,7 +13,7 @@ function setStatus(status) {
     statuses = [status]
   }
   const div = getStatusDiv()
-  div.innerHTML = `${statuses.map(getStatusHtml).join('')}`
+  div.innerHTML = `${statuses.filter(Boolean).map(getStatusHtml).join('')}`
```
kentcdodds: I don't care to have empty strings as the status.

stories/examples/controlled.js
```diff
@@ -34,6 +34,13 @@ class Examples extends Component {
   items = ['Black', 'Red', 'Green', 'Blue', 'Orange', 'Purple']
 
   changeHandler = changes => {
+    this.setState({
+      selectedColor: changes.selectedValue,
+      isOpen: false,
```
kentcdodds: This is so now when you select a value, the controller (the user of the downshift component) will close the menu, which is neat.

src/autocomplete.js
souporserious: What if... we got rid of `onChange` and just pointed people to something like `onStateChange({currentState, prevState})`.

souporserious: Or maybe we don't even worry about `prevState` since that should be fairly easy for someone to store 🤔

souporserious:
Other than my `onChange` comments, this all looks great 🎉

src/autocomplete.js
kentcdodds: I want to have a discussion about what we can remove in another issue before we release 1.0.0 :smile:

PR #88, opened 2017-08-05
Document/fix example

nndung179:


kentcdodds:
I think the git history in your branch is a bit messed up. You probably need to rebase it...

nndung179:
I think you are right. Please close this PR. I will make another one

PR #91, opened 2017-08-05
changed param of onChange

nndung179:
Currently, `onChange` will print out "undefined". As I discovered that `onChange` takes normal param instead of `object` param.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ function App() {
   return (
     <BasicAutocomplete
       items={['apple', 'orange', 'carrot']}
-      onChange={({selectedValue}) => console.log(selectedValue)}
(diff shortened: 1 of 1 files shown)


kentcdodds:
Hmmm... No, [it accepts an object](https://github.com/paypal/downshift/blob/cfd0e8e98e5e7ab69d38aaf3243334592f11e2f1/src/autocomplete.js#L302) 🤔

kentcdodds:
Are you sure you're using the latest version? Currently the version is `1.0.0-beta.13`

Please comment if I'm wrong :)

nndung179:
This issue happened when `1.0.0-beta.8` was latest version. I will test again later with a new version.

PR #69, opened 2017-08-02
Example/material autocompletely

nndung179:
Hi @kentcdodds 
This repo related to [https://github.com/paypal/react-autocompletely/pull/57](https://github.com/paypal/react-autocompletely/pull/57). Only add example link to README file

PR #92, opened 2017-08-05
added link to Material DownShift Example

nndung179:
This PR related to https://github.com/paypal/downshift/pull/69. I have updated to new version. Also refactor code. Now it looks cleaner.


nndung179:
Hi @kentcdodds !

Please review it for me.

Thanks!

kentcdodds:
Looks awesome! One thing. It doesn't appear to be using `match-sorter` which I think is great for something like this. More importantly, it doesn't visually show when items are highlighted as I use the up and down arrow keys. I'll merge this, but could you update it so it does that?

nndung179:
Sure. I will update

nndung179:
Hi @kentcdodds 
I added 2 things like you recommended:
+ match sorter
+ highlight item when using key up & down

PR #95, opened 2017-08-06
feat(activedescendant): add aria-activedescendant support

souporserious:
**What**:
Pertaining to https://github.com/paypal/downshift/issues/79.

**Why**:
This adds ARIA support for activedescendant.

**How**:
I used this [article from w3](https://www.w3.org/WAI/GL/wiki/Using_aria-activedescendant_to_allow_changes_in_focus_within_widgets_to_be_communicated_to_Assistive_Technology) as a guide. I ended up creating a unique `id` that could be used throughout the component, which ended up simplifying some things, so I was able to remove some code 🎉 let me know if I wasn't supposed to remove something, everything looked ok to remove. I went with a smaller id name as well, `downshift` abbreviated to `ds`. I personally like smaller id's so it keeps devtools cleaner and easier to digest, but I can change this to whatever, just let me know.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -195,7 +195,7 @@ class Autocomplete extends Component {
         isOpen: false,
       },
       () => {
-        const inputNode = document.getElementById(this.id)
+        const inputNode = document.getElementById(this.inputId)
         inputNode && inputNode.focus && inputNode.focus()
       },
     )
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -8,8 +8,7 @@ import {
   composeEventHandlers,
   debounce,
   scrollIntoView,
-  generateInputId,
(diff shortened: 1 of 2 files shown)


src/autocomplete.js
```diff
@@ -196,9 +195,7 @@ class Autocomplete extends Component {
         isOpen: false,
       },
       () => {
-        const inputNode = this._rootNode.querySelector(
-          '[data-autocomplete-input]',
-        )
+        const inputNode = document.getElementById(this.id)
```
kentcdodds: This should be referencing `this.inputId`

kentcdodds:
Love these changes. Very well done!

Issue #83, opened 2017-08-03
A note on equlity testing


ferdinandsalis:
We should add a note somewhere about equality testing for props like `isSelected`. This will bite people (me just now ☺️) since objects might not be the 'same' across different render cycles. Maybe in the examples or in a notes section.

kentcdodds:
Sounds great! Why don't you create a "Common Pitfalls" section in the README and add this to it :+1: Thanks!

ferdinandsalis:
Will do.

souporserious:
I was wondering about this. I was thinking we could treat `selectedValue` as `selectedValue: this.props.getValue(itemValue)` this way we are only working with strings or an array of strings so it's easier to do these things and avoid pitfalls. But that also might make it more complicated since you probably will need to find the item with helper like `getItemFromValue` or something.

kentcdodds:
I think that the `selectedValue` should be `item.value` personally...

souporserious:
I like it because you avoid the pitfalls. Do you think we should provide a helper like `getItemFromValue`?

kentcdodds:
Maybe I'm confused by what you're suggesting....

ferdinandsalis:
To be honest I am a bit confused about the different 'values'. This is my understanding of each of the 'values'.

* `selectedValue` — `selectedItem`, item is already a term used in the code
* `inputValue` — value of the input (this one is obvious) which is always a `string`
* `value` — the value from `getValue(selectedValue)` it kinda serves the purpose of an id, right?
* `defaultValue` — the `initialValue` of `value`

Also there is `getValue` which operates on `selectedValue`. I think of it as `getValueFromItem`. What are your thoughts. Hope this makes sense 😅. The gist of it, we should think hard about the names.

souporserious:
Some old API's in there, sorry 😅  `value` is no longer a thing. It seems `selectedValue` might not make sense, maybe `selectedItemValue` is better? And then maybe `defaultValue` to `defaultSelectedValue` or something.

ferdinandsalis:
[...]

PR #97, opened 2017-08-07
feat(autocomplete): store simple string values

souporserious:
**What**:
This deals with the issues brought up in https://github.com/paypal/downshift/issues/83.

**Why**:
This stores whatever `this.getValue` evaluates to so it truly is the `selectedValue`. I think this makes things a little bit easier to work with since we are sometimes dealing with multiple values. This also mimics a native select even more now which I like. We get to store complex values and retrieve them with simple string values 🎉 

**How**:
Instead of storing the item as the `selectedValue`, we now store `selectedValue: this.getValue(item)` which allows things like `defaultValue` to be very simple and you don't need to pass a complex object around. I also added a helper function called `getItemFromValue` to help grab the selected items for a trigger or whatever you might need them for.

I also added `isHighlighted` and `isSelected` to `getItemProps`.  I like that you have to make a custom component or pull them off like `const { isSelected, ...itemProps } = getItemProps({ value: option, index })` since it forces you to either pull them off or create a custom component. I personally think we should take this burden so people will style these things, plus it's going to be the same in every use case, so I think it's ok, would love to hear your thoughts on this.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

[Here is an example](https://gist.github.com/souporserious/248747f1381aacb4c3453de46947d0da) of a select menu that I am using at work that I built with these changes. It keeps the common, somewhat tedious logic away from the user and allows `downshift` to take care of it. Here is another example I was able to build:
```jsx
function StyledAutocomplete({ items, ...restProps }) {
  return (
    <Autocomplete {...restProps}>
      {({
        getButtonProps,
        getItemProps,
        getItemFromValue,
        getRootProps,
        highlightedIndex,
        isOpen,
        selectedValue
      }) =>
        <Box {...getRootProps({ refKey: 'innerRef' })}>
          <Button
            {...getButtonProps()}
            iconRight={{ name: isOpen ? 'arrow-up' : 'arrow-down' }}
            children={selectedValue.length > 0
              ? getItemFromValue(selectedValue, items).name
              : 'Select an item'}
          />
          {isOpen &&
            <Card maxHeight={12} overflow="auto">
              {items.map((item, index) => {
                const {
                  isSelected,
                  isHighlighted,
                  ...itemProps
                } = getItemProps({ value: item, index })
                return (
                  <Flex
                    {...itemProps}
                    key={item.code}
                    shrink={0}
                    backgroundColor={
                      isSelected
                        ? 'orange'
                        : isHighlighted
                          ? 'pink'
                          : 'white'
                    }
                  >
                    {item.name}
                  </Flex>
                )})
              }
            </Card>
          }
        </Box>
      }
    </Autocomplete>
  )
}
```

I'll be out of town for two weeks, so feel free to close this if it doesn't make sense 😇 or if you have any questions, I'll have my phone so I should be able to respond. I wasn't able to get to docs or updating examples 😞

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -15,7 +15,7 @@ class Autocomplete extends Component {
   static propTypes = {
     children: PropTypes.func.isRequired,
     defaultHighlightedIndex: PropTypes.number,
-    defaultValue: PropTypes.any,
+    defaultValue: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
     getA11yStatusMessage: PropTypes.func,
     getValue: PropTypes.func,
     multiple: PropTypes.bool,
@@ -25,7 +25,7 @@ class Autocomplete extends Component {
     // things we keep in state for uncontrolled components
     // but can accept as props for controlled components
     /* eslint-disable react/no-unused-prop-types */
-    selectedValue: PropTypes.any,
+    selectedValue: PropTypes.string,
     isOpen: PropTypes.bool,
     inputValue: PropTypes.string,
     highlightedIndex: PropTypes.number,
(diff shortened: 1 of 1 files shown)


kentcdodds:
Hey @souporserious! Thanks for this, unfortunately we can't really do this because the `getValue` of two items could result in the same string (let's say you have a list of contacts and two of the contacts are named "John Smith" for example).

So I wont be merging this one. Sorry!

PR #114, opened 2017-08-08
fix(getItemProps): clicking anywhere in the item selects the item

### LINKED ISSUE #113: Relying on the id attribute for selecting an item fails for compound components
- `downshift` version: 1.0.0-beta.21
- `node` version: 8.2.1
- `npm` (or `yarn`) version: 5.3.0

**What you did:**
Render a compound component as Item.

**What happened:**
Item does not get selected.

**Problem description:**
Item get selected only when the target node has the id attribute. With compound components this is clearly a problem.

**Suggested solution:**
We could walk up the dom tree to the nearest id attribute (maybe use data-item-id) instead.

I think this problem was also mentioned in https://github.com/paypal/downshift/issues/110

kentcdodds:
Closes #113

**What**: This looks up the tree to find the closest parent with the item id.

**Why**: This fixes an issue where if you have items with a nested structure the item would not be selected because the `target` did not have the `id` on it.

**How**: Refactor `getClosestScrollParent` into a generic `findParent` function and use that. This actually also fixed a bug in `getClosestScrollParent` where the `rootNode` was not passed recursively and the search would stop prematurely.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,13 +192,13 @@ properties:
 
 #### onChange
 
-> `function({selectedItem, previousItem})` | *required*
+> `function({selectedItem, previousItem})` | optional, no useful default
 
 Called when the user selects an item
 
 #### onStateChange
 
-> `function({highlightedIndex, inputValue, isOpen, selectedItem})` | not required, no useful default
+> `function({highlightedIndex, inputValue, isOpen, selectedItem})` | optional, no useful default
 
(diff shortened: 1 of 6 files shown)


kentcdodds:
```
😎  $ npm publish --tag beta
downshift@1.0.0-beta.22
```

Issue #119, opened 2017-08-09
RC release feedback


jgoux:
`downshift` version: 1.0.0-rc.1

Hello,

As I'm working on a [material-ui / virtual-list example](https://codesandbox.io/s/LWMA9QZj) using downshift I have a little list of remaining issues : 

- [x] The [`onMouseUp` listener](https://github.com/paypal/downshift/blob/master/src/downshift.js#L583) seems to force a rerender everytime we click outside of the component, even if there is no state / props changes. You can see the log on my codesandbox demo, try to click outside of the 2 components multiple times.
- [x] The `selectedItem` prop works great to make the component controlled, but the `inputValue` gets updated only when we finally click outside of the component (onMouseUp again I think). In my demo, try to select a value in the first component, then you'll notice the second one isn't updated fully until you click outside.
- [ ] The scroll management done internally in the lib seems to be in conflict with windowing libraries. I suspect the `scrollIntoView` and other functions which need to access the items nodes to be the problem. As the list becomes virtual, it tries to scroll to them before they are rendered. I had to do weird tricks to make it somehow workable, like adding this to my input component : 
```javascript
onKeyDown: e => {
  if (
    !autocompleteProps.isOpen ||
    !["ArrowUp", "ArrowDown"].includes(e.key)
  ) {
    return;
  }
  const top =
    e.key === "ArrowUp" && autocompleteProps.highlightedIndex === 0;
  const bottom =
    e.key === "ArrowDown" &&
    autocompleteProps.highlightedIndex === filteredItems.length - 1;

  if (!top && !bottom) {
    return;
  }

  const scrollToIndex = top ? filteredItems.length - 1 : 0;

  // go to the top or bottom of the virtual-list
  scroll({
    scrollToIndex
  });

  // allows to wait that the node to be highlighted is mounted
  setTimeout(() => {
    autocompleteProps.setHighlightedIndex(scrollToIndex);
  }, 0);
}
```

- [ ] The mouse interaction with the virtual-list is fine, but the keyboard one is buggy. When you hold `ArrowUp` or `ArrowDown` and the scroll gets too fast, the highlight / focus on item is lost.

- [ ] Still virtual-list related, the `Enter` event listener seems to be active only on initial mounted nodes. When you try to hit `ArrowUp`, and go at the bottom of the list, you can't select an item hitting `Enter`.

Don't hesitate to play with my demo to verify all of these issues.

We are almost there, thanks **a lot** to all contributors for the great work! 👍

[...]

PR #121, opened 2017-08-09
fix: update `inputValue` when `selectedItem` is controlled

kentcdodds:
Ref #119


kentcdodds:
publishing `1.0.0-rc.2` right now :+1:

PR #127, opened 2017-08-10
chore(precommit): integrate `remove-lockfiles`

luftywiranda13:
**What**:
Integrate [remove-lockfiles](https://github.com/luftywiranda13/remove-lockfiles)

**Why**:
Currently this project ignores lockfiles (package-lock.json & yarn.lock). This can lead to harder debugging process because contributors might be locked to different versions of deps

**How**:
- Install `remove-lockfiles` then integrate it in `precommit`
- Remove lockfiles from `.gitignore`

**How did I test this?**
- There are tests in `remove-lockfiles` repo
- Also tested in this project by installing deps using `yarn` and `npm`. Lockfiles were generated but not included in commits.
- `remove-lockfiles` not only unstages but also removes the lockfiles before commits

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 


kentcdodds:
After thinking more on this I've decided that this is only really useful for folks who use `yarn` because if you use `npm` you have the `package-lock=false` in the `.npmrc` (which you noted). Because this is only useful for folks who use `yarn`, _and_ it would only be a problem _very_ rarely even in that case, I think it's not worth the extra dependency.

Thanks anyway!

PR #129, opened 2017-08-10
fix(inputValue): initilize the inputValue correctly

### LINKED ISSUE #119: RC release feedback

kentcdodds:
**What**: This initializes the `inputValue` based on the `selectedItem`

**Why**: So if `selectedItem` is given as a prop, the `inputValue` is set based on `this.props.itemToString`. Closes #119

**How**: Updated the constructor and fixed a bug in `getState`.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/src/__tests__/downshift.lifecycle.js
+++ b/src/__tests__/downshift.lifecycle.js
@@ -69,6 +69,20 @@ test('props update causes the a11y status to be updated', () => {
   expect(setA11yStatus).toHaveBeenCalledTimes(1)
 })
 
+test('inputValue initializes properly if the selectedItem is controlled and set', () => {
+  const childSpy = jest.fn(() => null)
+  mount(
+    <Downshift selectedItem={'foo'}>
+      {childSpy}
+    </Downshift>,
+  )
+  expect(childSpy).toHaveBeenCalledWith(
+    expect.objectContaining({
+      inputValue: 'foo',
+    }),
+  )
+})
+
 test('props update of selectedItem will update the inputValue state', () => {
   const childSpy = jest.fn(() => null)
   const wrapper = mount(
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -58,12 +58,16 @@ class Downshift extends Component {
   constructor(...args) {
     super(...args)
     this.id = generateId('downshift')
-    this.state = {
+    const state = this.getState({
       highlightedIndex: this.props.defaultHighlightedIndex,
-      inputValue: this.props.defaultInputValue,
       isOpen: this.props.defaultIsOpen,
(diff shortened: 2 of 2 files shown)


kentcdodds:
```
🏎  $ npm publish --tag rc
+ downshift@1.0.0-rc.5
```

PR #131, opened 2017-08-10
feat(onChange): remove previousItem from onChange callback

### LINKED ISSUE #130: change the onChange API
Right now the `onChange` callback is called with `{selectedItem, previousItem}` but I'm starting to think we should just call it with `selectedItem`. I'm thinking this because the `onStateChange` is only called with the `changes, allState` and doesn't include the previous state.

I just built the [geniejs example](https://codesandbox.io/s/jRLKrxwgl) and found I had to preserve the `previousInputValue` myself. It wasn't difficult at all and made total sense that I would have to do that to support my use case. So I don't think I want to add supporting passing the previous state to the `onStateChange` callback.

I want to have consistency between these two APIs. And I don't think that it's too much to ask folks to keep track of the previous state if they need it. In fact, I think it's odd that people have to destructure out the `selectedItem` in the `onChange` handler. So for that reason, I'm going to make this change unless someone can convince me otherwise... :smile:

kentcdodds:
**What**: Removes `previousItem` from the `onChange` callback and instead calls it with just `selectItem, allState`.

**Why**: Closes #130

**How**: simplify things a bit and update examples

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ function App() {
   return (
     <BasicAutocomplete
       items={['apple', 'orange', 'carrot']}
-      onChange={({selectedItem}) => console.log(selectedItem)}
+      onChange={selectedItem => console.log(selectedItem)}
     />
   )
 }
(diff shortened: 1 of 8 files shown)


kentcdodds:
```
🏎  $ npm publish --tag rc
+ downshift@1.0.0-rc.6
```

PR #134, opened 2017-08-11
Make onChangeArg more strict

### LINKED ISSUE #133: onChange not trigerred when calling clearSelection
`downshift` version: 1.0.0-rc.8

Hi,

Since the last RC release, I'm unable to clear the selection when using the controlled component.

The `onChange` function isn't trigerred anymore when calling `clearSelection`

jgoux:
**What**: Fix #133

**Why**: So the onChange function will be called when passing a null selectedItem

**How**: By making a check for onChange's argument more strict (null being falsy in JS)

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -261,7 +261,7 @@ class Downshift extends Component {
         if (Object.keys(onStateChangeArg).length) {
           this.props.onStateChange(onStateChangeArg, this.getState())
         }
-        if (onChangeArg) {
+        if (onChangeArg !== undefined) {
           this.props.onChange(onChangeArg, this.getState())
         }
       },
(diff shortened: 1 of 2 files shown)


src/downshift.js
```diff
@@ -633,7 +633,7 @@ class Downshift extends Component {
     if (!uiDescriptor) {
       return null
     }
-    // doing React.Children.only for Preact support ⚛️
+    // doing React.Children.only for Preact support
```
kentcdodds: Awe 😕 why no emoji?

jgoux: Sorry as it was a really easy change I edited via Github directly, seems like the emoji goes away for some reason?

kentcdodds:
This is great! Thanks! Do you think you could add a test so this doesn't happen again?

kentcdodds:
Probably should go in the downshift.misc.js file

kentcdodds:
```
🏎  $ npm publish --tag rc
+ downshift@1.0.0-rc.9
```

PR #135, opened 2017-08-11
fix(preact): fix issues with preact support

### LINKED ISSUE #132: Fix preact support
It's kinda broken right now. I'm pretty sure it'll be easy to support, just need to dig into it a little bit. I'm thinking we can do away with `React.Children.only` and just do a little bit of logic. We need to make sure that if we get back an array from the `childre()` call we only pick the first time to return though...

kentcdodds:
This was a bit harder than I thought, but I think I got it!

cc @zouhir, @developit

**What**: Fix issues with preact support

**Why**: Fixes #132 (I really want this to work with preact-habitat because I think that'd be slick)

**How**:
- Add some tests (that was tough...)
- Make things compatible and not fail
- Test things out manually locally and they work! 🎉

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 


kentcdodds:
```
🏎  $ npm publish --tag rc
+ downshift@1.0.0-rc.10
```

PR #137, opened 2017-08-11
fix(onChange): only call `onChange` when the selection changes

kentcdodds:
Before, it was calling `onChange` any time there was a
`internalSetState({selectedItem})`


kentcdodds:
```
🏎  $ npm publish --tag rc
+ downshift@1.0.0-rc.11
```

PR #145, opened 2017-08-20
Add the-simian back to the "all-contributers" list

the-simian:
- I was accidentally removed when the repo was made public


the-simian:
This is the only change:
image

kentcdodds:
Hey! Thanks for this! That table is actually generated using a tool (see the CONTRIBUTING.md). I've updated it myself so we're all good! Thanks again for all your contributions! :tada:

Issue #53, opened 2017-07-30
Implement react-autosuggest API


kentcdodds:
Anyone wanna take on this task? By doing this it would really help us know whether react-autocompletely is powerful enough to support these use cases and if there's anything we can do to improve the API.

kentcdodds:
To be clear, this would probably be best as a [Codesandbox.io example](http://kcd.im/rac-example) (could be open sourced later if that's desired, but doesn't need to be).

bernard-lin:
Are you looking for something similar to the react-instantsearch example?

bernard-lin:
^bump 🙃

kentcdodds:
Oh, sorry about that! 😅 

Yeah, something like that would be fine :+1:

kentcdodds:
You really have the liberty to do whatever you'd like honestly :)

kentcdodds:
What I'm really looking for is some feedback on the API. How straightforward is it? Does it support your use cases? That kind of stuff.

bernard-lin:
Sounds good! 👍

kentcdodds:
FYI, pushed a pretty significant API change. Probably the last one. It's really great :smile:

bernard-lin:
I saw! Thanks for the update 😄

bernard-lin:
I've been weighed down by work so if anyone wants to take a shot at this in the meantime, please go for it!

souporserious:
@kentcdodds any idea how to achieve the functionality where the input is filled in with the suggested value like in react-autosuggest? I'm guessing you need to hold on to two values. The suggested one as well the current inputValue 🤔 it's an interesting interaction where keyboard highlighting is the only way to get a suggestion, whereas hovering highlighting does not pre fill the input.

kentcdodds:
I think that we'd just need to control the `inputValue` ourselves right?

souporserious:
I _think_ so. I haven't tried yet, mainly because I'm a little unsure how to handle it. It seems like we would need to know if highlights are coming from a mouse or keyboard.

kentcdodds:
>  It seems like we would need to know if highlights are coming from a mouse or keyboard.

Yeah, it does seem that way. Maybe we'll need to add to the [`stateChangeTypes`](https://github.com/paypal/downshift/blob/be71c80318d6412fd89716c7b6d27e5d72551f57/src/downshift.js#L57-L61) after all. I didn't want to do that, but I think that'd be the best way to support this...

[...]

PR #147, opened 2017-08-22
feat(stateChangeTypes): add a state change type to everything

kentcdodds:
This is an undocumented API, but I think it'll be really useful for some
use cases so we'll try to support it.

I also kinda broke a few APIs in the process. Shouldn't impact anyone,
but it is technically different. Please see the changes in the README.

**What**: add a state change type to everything

**Why**: To support a few edge cases #53 

**How**: Changing the APIs of several actions to accept `otherStateToSet` which can have a `type` property.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 


kentcdodds:
```
🏎  $ npm publish --tag rc
+ downshift@1.0.0-rc.17
```

PR #149, opened 2017-08-22
feat: add `onUserAction` to be notified whenever state could be updated

### LINKED ISSUE #148: Add onUserAction
The recent change in #147 made it so `onStateChange` is called even though there are no state changes (because there's a `type` property on most `internalSetState` calls now).

So I'm wondering whether we should add an `onUserAction` which would be called with `type`, `changes`, and `allState`. Then we'd change `onStateChange` to only have `changes` (no `type` anymore).

Thoughts here?

I'm pretty sure that `onStateChange` being called repeatedly isn't a big deal, but I may be missing something here. I think that @jgoux had some issues with it being called too frequently...

I'm hoping to publish 1.0.0 by tomorrow so quick feedback would be appreciated!

kentcdodds:
**What**: add `onUserAction` to be notified whenever state could be updated

**Why**: This allows us to gate `onStateChange` to only be called when state actually changed. Closes #148 

**How**:
- Pass a `type` 100% of the time (default to `unknown`)
- gate calling `onStateChange`
- always call `onUserAction`.

Not sure I love the name `onUserAction` because this will be called even if things happen programmatically. Hence this is not documented yet...

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/stories/examples/react-autosuggest.js
+++ b/stories/examples/react-autosuggest.js
@@ -15,38 +15,40 @@ class Examples extends Component {
     this.setState({selectedColor})
   }
 
-  onStateChange = changes => {
-    let {selectedColor, inputValue, itemsToShow} = this.state
-    const isClosingMenu = changes.hasOwnProperty('isOpen') && !changes.isOpen
-    if (
-      changes.type === Downshift.stateChangeTypes.keyDownEscape &&
-      !isClosingMenu
-    ) {
-      selectedColor = null
-    }
-    if (changes.hasOwnProperty('inputValue')) {
-      if (changes.type === Downshift.stateChangeTypes.keyDownEscape) {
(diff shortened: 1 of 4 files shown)


stories/examples/react-autosuggest.js
```diff
@@ -15,38 +15,40 @@ class Examples extends Component {
     this.setState({selectedColor})
   }
 
-  onStateChange = changes => {
```
kentcdodds: Changes here aren't as dramatic as it looks. Just wrapped everything in a `this.setState` callback.

kentcdodds:
```
🏎  $ npm publish --tag rc
+ downshift@1.0.0-rc.18
```

PR #152, opened 2017-08-25
Improve typings

thomhos:
**What**:
Minor improvements in typings for ```getButtonProps, getInputProps, getItemProps```.

**Why**:
When using downshift with typescript and following examples, I've found the get props functions have a lot of optional properties that are marked mandatory in the typings.

**How**:
I've updated the index.d.ts file and ts test file.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged 


test/basic.test.tsx
```diff
@@ -49,7 +53,7 @@ export default class App extends React.Component<Props, State> {
                                       )
                                       .map((item: any, index: number) =>
                                           <div
-                                              {...getItemProps({ item, index })}
+                                              {...getItemProps({ item, index, isSelected: selectedItem === item })}
```
kentcdodds: This is going to put an attribute on the div called isSelected. I realize this is just a test, but it's probably best to make sure that the test code is correct 👍

vutran: Agreed, we may want to remove user-defined props from the basic test. Perhaps a secondary test for cases like this would work?

.all-contributorsrc
```diff
@@ -205,6 +205,13 @@
       "contributions": [
         "tool"
       ]
+    },
+    {
+      "login": "thomhos",
+      "name": "thom",
+      "avatar_url": "https://avatars1.githubusercontent.com/u/11661846?v=4",
+      "profile": "http://thom.kr",
+      "contributions": []
```
vutran: Mind adding me too (if it's okay with @kentcdodds)? I had forgotten to add myself to this list. 😃

test/basic.test.tsx
pbomb: `isSelected` is not a valid DOM element prop, so would generate a warning from React, right? Tests shouldn't generate React warnings, so if this would, I think it would be better off removed.

I'm new to downshift and trying to understand why `getItemProps` returns any key/value pairs it doesn't recognize. In this example, it would be easy enough to get the same behavior using this code:
```js
                                         <div
                                              {...getItemProps({ item, index })}
                                              isSelected={selectedItem === item}
```
Is there a different scenario where this approach is not possible and props need to be passed through `getItemProps`? If not, then why return `...rest` inside that function?

morajabi: @pbomb is right I suppose.

thomhos: I've added a new test that shows the above with a custom component for a list item. Adding isSelected separately is also possible, but since the ```...rest``` is in the js, I figured to have the typings reflect that as well.

.all-contributorsrc
thomhos: done!

PR #159, opened 2017-08-29
feat(index): make index optional

kentcdodds:
**What**: This makes indexes optional

**Why**: So @souporserious doesn't feel like he needs to make another dropdown implementation because [grouping is hardish with downshift](https://twitter.com/souporserious/status/902243143843700736) :wink:

**How**: default `index` to the `indexOf` after pushing an item onto the `items`. Going with `indexOf` rather than `this.items.length - 1` because who knows 🤷‍♂️

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ function BasicAutocomplete({items, onChange}) {
                 )
                 .map((item, index) => (
                   <div
-                    {...getItemProps({item, index})}
+                    {...getItemProps({item})}
                     key={item}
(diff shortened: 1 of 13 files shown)


souporserious:
Appreciate it! I'm trying to solve some other problems as well. But maybe we can merge some of my stuff in after I'm done if it makes sense 😊

PR #165, opened 2017-08-31
Types 1.2.0

tansongyang:
**What**: Update typings.

**Why**: Some types did not match the docs.

**How**: Edit index.d.ts

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

I tried to make only non-breaking changes (adding types, making required types optional, etc.) with one exception. `GetButtonPropsOptions` was requiring incorrect properties—in fact, it shouldn't be requiring any at all. I completely removed them.

There are a couple things that probably should change, but I didn't want to make them for fear of introducing a breaking change. These are:

* The props `onUserAction` and `onClick` in `DownshiftProps` are not in the docs. Have they been removed? If so, they should be removed from the type definitions as well.
* `A11StatusMessageOptions` should probably be renamed to `A11yStatusMessageOptions` (there is a missing "y")
* `ChangeOptions` is completely unused, and I could not find any mention of it in the docs. However, it is exported, so I did not remove it in case anyone is importing it.


kentcdodds:
ping @pbomb and @vutran could you review this?

typings/index.d.ts
```diff
@@ -1,25 +1,28 @@
 import * as React from 'react';
 
+type CB = () => void
```
pbomb: The callbacks are optional. Additionally, the code checks if they are functions and only returns them if they are. So, I think it would be more correct to go with either `Function | undefined` or `() => any | undefined`. I think `void` is too restrictive - we don't really care what the callback function returns.

The TypeScript types for React's `setState` function (which is how these callbacks are used) is typed as `() => any | undefined`

typings/index.d.ts
```diff
@@ -1,25 +1,28 @@
...
-    defaultHighlightedIndex?: number;
+    defaultHighlightedIndex?: number | null;
     defaultSelectedItem?: any;
     defaultInputValue?: string;
     defaultIsOpen?: boolean;
     getA11yStatusMessage?: (options: A11StatusMessageOptions) => any;
     itemToString?: (item: any) => string;
     onChange?: (selectedItem: any, stateAndHelpers: ControllerStateAndHelpers) => void;
-    onStateChange?: (options: StateChangeOptions, stateAndHelpers: ControllerStateAndHelpers) => void;
+    onStateChange?: (changes: StateChangeOptions, stateAndHelpers: ControllerStateAndHelpers) => void;
```
pbomb: I prefer to keep this named as `options` to be consistent with everything else in here.

typings/index.d.ts
```diff
@@ -85,18 +72,20 @@ export interface ControllerStateAndHelpers {
...
-    toggleMenu: () => void;
-    selectItem: (item: any) => void;
-    selectItemAtIndex: (index: number) => void;
-    selectHighlightedItem: (index: number) => void;
-    setHighlightedItem: (index: number) => void;
-    clearSelection: () => void;
+    openMenu: (cb: CB) => void;
+    closeMenu: (cb: CB) => void;
+    toggleMenu: (cb: CB) => void;
+    selectItem: (item: any, otherStateToSet: any, cb: CB) => void;
```
pbomb: `otherStateToSet` and `cb` should be optional types. Additionally, if `otherStateToSet` is defined, it should be an object as it gets spread into another object.

typings/index.d.ts
```diff
@@ -85,18 +72,20 @@ export interface ControllerStateAndHelpers {
...
+    selectItemAtIndex: (index: number, otherStateToSet: any, cb: CB) => void;
```
pbomb: `otherStateToSet` and `cb` should be optional types. Additionally, if `otherStateToSet` is defined, it should be an object as it gets spread into another object.

typings/index.d.ts
```diff
@@ -85,18 +72,20 @@ export interface ControllerStateAndHelpers {
...
+    selectHighlightedItem: (otherStateToSet: any, cb: CB) => void;
```
pbomb: `otherStateToSet` and `cb` should be optional types. Additionally, if `otherStateToSet` is defined, it should be an object as it gets spread into another object.

typings/index.d.ts
```diff
@@ -85,18 +72,20 @@ export interface ControllerStateAndHelpers {
...
+    setHighlightedItem: (index: number, otherStateToSet: any, cb: CB) => void;
```
pbomb: `otherStateToSet` should be optional and there is no `cb` argument to this function.

typings/index.d.ts
```diff
@@ -85,18 +72,20 @@ export interface ControllerStateAndHelpers {
...
+    clearSelection: (cb: CB) => void;
```
pbomb: `cb` is optional

typings/index.d.ts
```diff
@@ -85,18 +72,20 @@ export interface ControllerStateAndHelpers {
...
+    reset: (otherStateToSet: any, cb: CB) => void;
```
pbomb: `cb` is optional

typings/index.d.ts
tansongyang: Are you talking about `setHighlightedItem` here? If so, I think that, first of all, this must be a typo, as I don't see any `setHighlightedItem` in the docs. However, there is `setHighlightedIndex`. If that is the case, I believe it does have a `cb` argument.

typings/index.d.ts
tansongyang: Is `otherStateToSet` also optional here?

tansongyang:
@pbomb First of all, thanks for reviewing the PR! I have a local commit with the changes, but I'm waiting to push until we resolve my two follow-up questions. Also, do you have any thoughts on my additional comments (in the original PR post)? If you think any of those are worth doing, we can also do them in this branch so that it all gets merged in on one go.

pbomb:
@tansongyang sorry, i missed your questions in the initial PR posting. Here's what my thoughts are.

* You're right, there does seem to be an optional `onClick` prop that isn't documented. @kentcdodds does documentation need to be added for this prop? It looks to me like the user can pass `onClick` both to the `<Downshift/>` component and to the top-level element returned by the function-as-child and that they will both be called as the click handlers are all composed together. Does that sound right to you?
* Yes, `A11StatusMessageOptions` should be renamed to `A11yStatusMessageOptions` as it describes the argument to the `getA11yStatusMessage` function prop.
* You should remove `ChangeOptions` as it's out of date. That used to be the type of the first argument passed to the `onChange` callback prop, but was changed to just be the `selectedItem` in commit `620a3ad5e4f50dece867ae9989174a8b35104fb3` and `previousItem` was removed in that commit as well.

typings/index.d.ts
pbomb: Hmm, I was talking about `setHighlightedItem`, but now I don't see it anywhere in the repo except in the TypeScript definitions, so I think you can remove this line.
As for the line above, with `selectHighlightedItem`, you're right. `otherStateToSet` should be optional and there is also an optional `cb` argument.

kentcdodds:
> You're right, there does seem to be an optional onClick prop that isn't documented. @kentcdodds does documentation need to be added for this prop? It looks to me like the user can pass onClick both to the <Downshift/> component and to the top-level element returned by the function-as-child and that they will both be called as the click handlers are all composed together. Does that sound right to you?

That's a mistake actually. We should remove the prop. It's undocumented so I don't think we need a breaking change for it. Sorry about that!

tansongyang:
I've made the changes that we discussed. Let me know if anything else needs to change.

kentcdodds:
There are merge conflicts. Could you fix them? Thanks!

tansongyang:
@kentcdodds Thanks for getting that README merge conflict! It was a pleasure working with you all.

kentcdodds:
You're welcome!

The build is broken right now due to a problem with TypeScript. Could someone look into that please?

pbomb:
I'm taking a look now

tansongyang:
Found the issue. After I removed `ChangeOptions`, I forgot to update the tests, which reference `ChangeOptions`. That's my bad. Working on a fix right now.

tansongyang:
In my local environment, I believe I've fixed the TS issue, as it now successfully compiles. However, `npm start validate` reports lint issues with `downshift\stories\examples\multiple.js`. What would you like to do about that?

Also, the problem in the test files was `custom.test.tsx` and `basic.test.tsx` were importing `ChangeOptions`, but they weren't using it. Removing it from the imports fixes the issue.

kentcdodds:
Go ahead and fix whatever you can and I'll fix what's left. Thanks!

tansongyang:
New pull request: https://github.com/paypal/downshift/pull/175

Apologies for the inconvenience. Thanks for being flexible.

PR #172, opened 2017-09-04
Auto-open menu on focus when input is non-empty

srph:
**What**:

This auto-opens the menu when the input with value is focused. Previously, user has to type something (albeit he already had) to open up the menu.

**Why**:

A number of autocomplete components implement this feature, namely react-select. Google does this as well. I also think that this is a convenient UX.

This is easily doable outside Downshift, but I think it should be part of the core. In addition, we can add a prop that allows users to disable this option.

**How**:

By adding a new focus event listener.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

Not yet, I'd like to make this as a discussion starter - it's a pretty simple change anyway.

If this sounds good, I'd be happy to push tests.

Will put up GIFs within the hour (haha, never done them with Windows).

### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -528,6 +528,7 @@ class Downshift extends Component {
         onInput,
         this.input_handleChange,
       ),
+      onFocus: this.input_handleFocus,
       onKeyDown: composeEventHandlers(onKeyDown, this.input_handleKeyDown),
       onBlur: composeEventHandlers(onBlur, this.input_handleBlur),
       ...rest,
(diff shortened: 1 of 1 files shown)


kentcdodds:
Thanks for the contribution! However I'm not going to merge this as it's really easy to do yourself: https://codesandbox.io/s/k2k73rw7lo

(the important part):

```js
{...getInputProps({
  isOpen,
  placeholder: 'Enter a name',
  onFocus: openMenu,
})}
```

srph:
TIL. Looks like I wasn't paying enough attention to the docs. Thanks!

PR #180, opened 2017-09-08
Flow definitions

nitin42:
**What**: Flow definitions

**Why**: Support for flow types

**How**: Added flow types and modified config (`package.json` and `.babelrc`) of the project

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 


nitin42:
Done! Sorry, I've been busy with maintaining some of my projects (also interviews) and also tweaking experimenting with glamorous.

kentcdodds:
Sweet! Could you rebase your changes onto master? And add `flow` to the list in the `validate` script here: https://github.com/paypal/downshift/blob/master/package.json#L22 :)

pbomb:
There are some differences between these definitions and the ones we just released for TypeScript. A couple of them are just out of date, but some are missing optional items (like a few of the callbacks). @nitin42 once you rebase from master, can you compare your flow file and the TypeScript one and make sure they match?

andreiglingeanu:
Any way to move this forward? I could probably get some help, if needed, just let me know how!

kentcdodds:
I think @nitin42 had some ideas about this...

andreiglingeanu:
Would love to hear them!

package.json
```diff
@@ -50,12 +51,14 @@
     "babel-plugin-transform-react-jsx": "^6.24.1",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.8",
     "babel-preset-env": "^1.6.0",
+    "babel-preset-flow": "^6.23.0",
```
TrySound: Flow preset is not required for definitions.

nitin42:
Woaah! Finally, I've updated the type definition file for `Flow`. @andreiglingeanu @pbomb You can still review the type definitions before @kentcdodds blends this into master 😄

package.json
```diff
@@ -8,6 +8,12 @@
   "module": "dist/downshift.esm.js",
   "typings": "typings/index.d.ts",
   "scripts": {
+    "start": "nps",
+    "test": "nps test",
+    "prerelease": "npm start validate",
+    "release": "npm publish && npm dist-tag add downshift@$npm_package_version rc",
+    "precommit": "lint-staged && opt --in pre-commit --exec \"npm start validate\"",
+    "flow": "flow"
```
TrySound: ,

typings/downshift.flow.js
```diff
@@ -0,0 +1,113 @@
+import * as React from 'react'
+
+// @flow
```
TrySound: `// @flow` should be placed before any statements.

package.json
TrySound: missed comma I mean

package.json
```diff
@@ -8,6 +8,12 @@
   "module": "dist/downshift.esm.js",
   "typings": "typings/index.d.ts",
   "scripts": {
+    "start": "nps",
+    "test": "nps test",
```
pbomb: "test" is already defined as "kcd-scripts test"

package.json
```diff
@@ -8,6 +8,12 @@
   "module": "dist/downshift.esm.js",
   "typings": "typings/index.d.ts",
   "scripts": {
+    "start": "nps",
```
pbomb: why is `nps` being added as part of this PR to add flow definitions? For example, why would someone execute `npm start validate` instead of `npm run validate`?

package.json
```diff
@@ -8,6 +8,12 @@
...
+    "test": "nps test",
+    "prerelease": "npm start validate",
+    "release": "npm publish && npm dist-tag add downshift@$npm_package_version rc",
```
pbomb: I'm also worried that this PR affects how downshift is being released. @kentcdodds do these scripts need to be added? I figured the releasing was already being taken care of by `kcd-scripts`.

package.json
pbomb: Shouldn't `flow` be added to the `validate` script? I'm not sure if that can be done via `kcd-scripts` or if it has to be a separate command in that script. @kentcdodds, thoughts?

package.json
```diff
@@ -50,6 +56,30 @@
     "react": ">=15"
   },
   "devDependencies": {
+    "@storybook/react": "^3.2.3",
+    "all-contributors-cli": "^4.3.0",
```
pbomb: I believe most of these dependencies are handled by kcd-scripts. @kentcdodds, can you verify which dependencies need to be added. For example, there are already storybook scripts in `package.json` so I wouldn't expect that dependency to need to be added here.

typings/downshift.flow.js
```diff
@@ -0,0 +1,113 @@
...
+type ChildrenFunction = (options: ControllerStateAndHelpers) => React.Node<*>
+
+type downshiftProps = {
+  children: ChildrenFunction,
+  defaultHighlightedIndex?: number,
+  defaultSelectedItem?: any,
+  defaultInputValue?: string,
+  defaultIsOpen?: boolean,
+  getA11yStatusMessage?: (options: A11StatusMessageOptions) => any,
+  itemToString?: (item: any) => any,
```
pbomb: this function should return `string` instead of `any`

typings/downshift.flow.js
```diff
@@ -0,0 +1,113 @@
...
+  // state
+  highlightedIndex: number,
+  inputValue: string,
+  isOpen: boolean,
+  selectedItem: any,
+}
+
+type ChildrenFunction = (options: ControllerStateAndHelpers) => React.Node<*>
+
+type downshiftProps = {
```
pbomb: `onOuterClick?: () => void` is missing

pbomb: Can you capitalize `downshiftProps` to match the other type definitions?

typings/downshift.flow.js
```diff
@@ -0,0 +1,113 @@
...
+  itemToString?: (item: any) => any,
+  onChange?: (
+    options: ChangeOptions,
+    stateAndHelpers: ControllerStateAndHelpers,
+  ) => void,
+  onStateChange?: (
+    options: StateChangeOptions,
+    stateAndHelpers: ControllerStateAndHelpers,
+  ) => void,
+  onClick?: Function,
```
pbomb: It looks like the TypeScript definitions are missing this prop. However, I also don't understand how Downshift is using the root `onClick` prop. @kentcdodds should downshift have the `onClick` prop? How is it getting used? For instance, I don't see how `root_handleClick` is being used.

typings/downshift.flow.js
pbomb:
```js
  onUserAction?: (
    options: StateChangeOptions,
    stateAndHelpers: ControllerStateAndHelpers,
  ) => void
``` is also missing

Comment (pbomb) [reply typings/downshift.flow.js]: `itemCount: number | void` should be added

Comment (pbomb) [reply typings/downshift.flow.js]: `id?: string` is missing

Comment (pbomb) [reply typings/downshift.flow.js]: `environment?: Environment` also needs to be added. You can get the definition for Environment from the TypeScript definition file

File: typings/downshift.flow.js [hunk 11]
```diff
@@ -0,0 +1,113 @@
+import * as React from 'react'
+
+// @flow
+
+type A11StatusMessageOptions = {
+  highlightedIndex: number,
```
pbomb: this can also be `null`

typings/downshift.flow.js
```diff
@@ -0,0 +1,113 @@
...
+  inputValue: string,
+  isOpen: boolean,
+  selectedItem: any,
+}
+
+type GetInputPropsOptions = React.InputHTMLAttributes
+
+type GetLabelPropsOptions = React.LabelHTMLAttributes
+
+type GetButtonPropsOptions = {
```
pbomb: Can this not use something like `React.ButtonHTMLAttributes`? The TypeScript definitions are doing something like this.

typings/downshift.flow.js
```diff
@@ -0,0 +1,113 @@
...
+  getButtonProps: (options: GetButtonPropsOptions) => any,
+  getLabelProps: (options: GetLabelPropsOptions) => any,
+  getInputProps: (options: GetInputPropsOptions) => any,
+  getItemProps: (options: GetItemPropsOptions) => any,
+
+  // actions
+  openMenu: () => void,
+  closeMenu: () => void,
+  toggleMenu: () => void,
+  selectItem: (item: any) => void,
```
pbomb: This can take an optional `otherStateToSet` object and optional callback function

typings/downshift.flow.js
```diff
@@ -0,0 +1,113 @@
...
+  selectItemAtIndex: (index: number) => void,
```
pbomb: This can take an optional `otherStateToSet` object and optional callback function

typings/downshift.flow.js
```diff
@@ -0,0 +1,113 @@
...
+  selectHighlightedItem: (index: number) => void,
```
pbomb: This property def is incorrect. It takes an optional `otherStateToSet` object and optional callback function.

typings/downshift.flow.js
```diff
@@ -0,0 +1,113 @@
...
+  setHighlightedItem: (index: number) => void,
```
pbomb: `setHighlightedItem` doesn't exist. This should be:
```js
selectHighlightedIndex: (index?: number, otherStateToSet: object) => void
```

typings/downshift.flow.js
```diff
@@ -0,0 +1,113 @@
...
+  clearSelection: () => void,
+
+  // state
+  highlightedIndex: number,
```
pbomb: I believe this can also be `null`

typings/downshift.flow.js
```diff
@@ -0,0 +1,113 @@
...
+type ControllerStateAndHelpers = {
+  // prop getters
+  getRootProps: (options: GetRootPropsOptions) => any,
+  getButtonProps: (options: GetButtonPropsOptions) => any,
+  getLabelProps: (options: GetLabelPropsOptions) => any,
+  getInputProps: (options: GetInputPropsOptions) => any,
+  getItemProps: (options: GetItemPropsOptions) => any,
+
+  // actions
+  openMenu: () => void,
```
pbomb: The functions from here to `clearSelection` can all take an optional callback function as their last argument.

typings/downshift.flow.js
```diff
@@ -0,0 +1,113 @@
...
+type GetRootPropsOptions = {
+  refKey: string,
+}
+
+type GetItemPropsOptions = {
+  index: number,
+  item: any,
+}
+
+type ControllerStateAndHelpers = {
```
pbomb: this type is also missing these props:
```js
reset: (otherStateToSet?: object, cb?: Function) => void,
itemToString: (item: any) => string // this is incorrectly defined in the TypeScript defs
id: string // this is missing in the TypeScript defs

```

typings/downshift.flow.js
smdern: You can do something like `React$ElementProps<'button'>`

kentcdodds:
I'm going to go ahead and close this PR. If anyone feels motivated to pick it back up please do so in another PR. Thanks :)

Issue #182, opened 2017-09-09
Console errors when using 'toggleMenu' action prop


ikovic:
- `downshift` version: 1.3.1
- `node` version: 8.2.1
- `npm` (or `yarn`) version: yarn 0.27.5

Relevant code or config

```javascript
export default class LanguageDropdown extends Component {
  itemToString = item => item.name;

  toggleDropdown = () => this.setState(state => !state.isOpen);

  render() {
    return (
      <Downshift itemToString={this.itemToString}>
        {({ toggleMenu, isOpen, ...rest }) =>
          <div className={s.inputContainer}>
            <Title onClick={toggleMenu} />
            {isOpen ? <Menu {...{ items: languages, ...rest }} /> : null}
          </div>}
      </Downshift>
    );
  }
```

What you did:

I've tried using the `toggleMenu` action and `isOpen` props to control the visibility of a dropdown menu.

What happened:

`isOpen` value changes as expected, but the console shows a bunch of these errors:
<img width="860" alt="screen shot 2017-09-09 at 15 08 34" src="https://user-images.githubusercontent.com/1204170/30240438-d0f1f080-9570-11e7-8a29-3c1a1e397d14.png">

Problem description:

Console gets a couple of above errors on each execution of `toggleMenu` action.

Suggested solution:

I could easily get around this by managing the state of `isOpen` on my own, but it would be better if I could use the `downshift` actions for that simple use case.

tansongyang:
I'm seeing the same thing. I made a [minimal reproduction](https://codesandbox.io/s/81q4voppl9) and I get the same errors.

kentcdodds:
Hi! Yeah, so the problem is that `onClick` handlers are invoked with an `event` and `toggleMenu` accepts an argument called `otherStateToSet`. So if it's passed an `event`, then downshift tries to set state with the `event`.

So the solution is simple: `onClick={() => toggleMenu()}`...

You know... I think we could make this nicer for folks. In [`toggleMenu`](https://github.com/paypal/downshift/blob/a1490f070575ffcd71dac8c063fe7e092f840a67/src/downshift.js#L610-L623), we could check if `otherStateToSet` is an `event`, and if it is, then we don't spread it on the state we return in our `internalSetState` callback. This would be tested [here](https://github.com/paypal/downshift/blob/4fd6ae258948c70f214d7f90ac03490d854eb39a/src/__tests__/downshift.misc.js#L58).

Who wants to [do this](http://makeapullrequest.com/)?

[...]

PR #183, opened 2017-09-10
feat(pickState): only pick relevant state

tansongyang:
Make `toggleMenu` and friends more user-friendly by only picking out
state that is relevant to downshift, allowing users to use them as event
event handlers. See #182.

**What**: Make `toggleMenu` and friends more user-friendly.

**Why**:  See #182.

**How**: Add `pickState` and call it whenever we are about to spread `otherStateToSet` into `internalSetState`

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] Ready to be merged 

Additional comments:

1. When I initially pulled the most recent from master, `npm run validate` failed.

	    FAIL  other\__tests__\build.js
  	     ● Test suite failed to run

	     Cannot find module '../../dist/downshift.es' from 'build.js'

	       at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:179:17)
	       at Object.<anonymous> (other/__tests__/build.js:14:18)

  	    FAIL  other\__tests__\preact.js
	     ● Test suite failed to run

	     Cannot find module '../../dist/downshift.preact.cjs' from 'preact.js'

	       at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:179:17)
	       at Object.<anonymous> (other/__tests__/preact.js:22:24)

    I investigated a bit. Only the `build-and-test` task is failing. `lint`, `test:cover`, and `test:ts` succeed. I decided to ignore it while developing because I could still run the other test tasks, thinking it could just be a problem with my machine. *Please let me know if we should address this first.*
2. I wanted to add direct tests on `toggleMenu` and others to `downshift.misc.js` (in addition to adding `utils.pick-state.js`), but there wasn't a way to test what object was being passed to `internalSetState`. No matter what the internal state is, we always pass the correct keys to `render` (https://github.com/paypal/downshift/blob/master/src/downshift.js#L336). So, with the current test setup, I couldn't find a way to see what the internals were. I decided it was OK, thinking the new test was sufficient, but let me know if you think otherwise.
3. Locally, I added a new story to test the changes. It didn't seem appropriate to include it in the commit, but let me know if you think otherwise.

### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -18,6 +18,7 @@ import {
   getElementProps,
   noop,
   requiredProp,
+  pickState,
 } from './utils'
 
 class Downshift extends Component {
@@ -154,6 +155,7 @@ class Downshift extends Component {
     highlightedIndex = this.props.defaultHighlightedIndex,
     otherStateToSet = {},
   ) => {
+    otherStateToSet = pickState(otherStateToSet)
     this.internalSetState({highlightedIndex, ...otherStateToSet}, () => {
       const node = this.getItemNodeFromIndex(this.getState().highlightedIndex)
       const rootNode = this._rootNode
@@ -213,6 +215,7 @@ class Downshift extends Component {
   }
 
   selectItem = (item, otherStateToSet, cb) => {
+    otherStateToSet = pickState(otherStateToSet)
     this.internalSetState(
       {
         isOpen: false,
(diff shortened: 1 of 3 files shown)


src/downshift.js
```diff
@@ -154,6 +155,7 @@ class Downshift extends Component {
     highlightedIndex = this.props.defaultHighlightedIndex,
     otherStateToSet = {},
   ) => {
+    otherStateToSet = pickState(otherStateToSet)
```
kentcdodds: Actually, I think this is the only place we really need this function called because all the other code paths lead here. Try removing the other changes and see if this works. I think it will.

tansongyang: I just tried this, and it didn't work for me. I also tried to move this further down into `internalSetState`, and that didn't remove the warnings either.

PR #195, opened 2017-09-21
Reset highlightedIndex to defaultHighlightIndex after selecting item

### LINKED ISSUE #194: `defaultHighlightedIndex ` is set back to `null` after first selection
- `downshift` version: 1.7.0
- `node` version: 8.3.0
- `npm` (or `yarn`) version: 5.4.2

What you did:

I set `defaultHighlightedIndex={0}` on the `<Downshift />` component.

What happened:

When selecting an item for the first time, it correctly highlights the first item. After changing the selection for a second time however, there is no default highlight on the first item anymore. This is because `highlightedIndex` is reset to `null` after the first selection. 

Reproduction repository:

https://codesandbox.io/s/vnkkw205kl

Problem description:

Most "fancy select" components automatically select the first item in the list, so the user can just press `<enter>`. E.g. react-select does this. I tried to imitate this functionality by using the `defaultHighlightedIndex` prop, but it did not behave like I expected.

Suggested solution:

Don't reset `highlightedIndex` to `null`, but reset it to the value in `defaultHighlightedIndex` after selecting an item. I don't know if that makes sense though, so perhaps what I'm asking is if there's another easy way to achieve this?

-----

Btw: I really like Downshift! This finally let me to understand why packages that only provide primitives are so powerful 👍

SpaceK33z:
**What**:
Fixes #194

**Why**:

As a user I'd expect `highlightedIndex` to always have the value that I initially set it to, after selecting an item. I.e. react-select also works like this. It allows the user to quickly hit enter when the right item is the first in the list.

**How**:

`highlightedIndex` was always set to `null` when selecting an item. I changed that to be set to the `defaultHighlightedIndex`prop.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/src/__tests__/downshift.get-item-props.js
+++ b/src/__tests__/downshift.get-item-props.js
@@ -50,6 +50,19 @@ test('on mouseenter of an item updates the highlightedIndex to that item', () =>
   )
 })
 
+test('after selecting an item highlightedIndex should be reset to defaultHighlightIndex', () => {
+  const {Component, childSpy} = setup()
+  const wrapper = mount(<Component defaultHighlightedIndex={1} />)
+  const firstButton = wrapper.find('button').first()
+  childSpy.mockClear()
+  firstButton.simulate('click')
+  expect(childSpy).toHaveBeenCalledWith(
+    expect.objectContaining({
+      highlightedIndex: 1,
+    }),
+  )
+})
+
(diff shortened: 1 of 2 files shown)


kentcdodds:

PR #202, opened 2017-09-25
Update README.md

briandipalma:
Typo in readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ and the new state of `downshift`. (see `onStateChange` for more info on
 `stateAndHelpers`).
 
 - `selectedItem`: The item that was just selected
-- `stateAndHelpers`: This is the same thing your `children` prop
+- `stateAndHelpers`: This is the exact same thing you're `children` prop
   function is called with (see [Child Callback Function](#child-callback-function))
 
 ### onStateChange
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -218,15 +218,15 @@ and the new state of `downshift`. (see `onStateChange` for more info on
...
+- `stateAndHelpers`: This is the exact same thing you're `children` prop
   function is called with (see [Child Callback Function](#child-callback-function))
 
 ### onStateChange
 
 > `function(changes: object, stateAndHelpers: object)` | optional, no useful default
 
 This function is called anytime the internal state changes. This can be useful
-if you're using downshift as a "controlled" component, where you manage some or
+if your using downshift as a "controlled" component, where you manage some or
```
kentcdodds: The use of `you're` here is correct.

README.md
```diff
@@ -218,15 +218,15 @@ and the new state of `downshift`. (see `onStateChange` for more info on
 `stateAndHelpers`).
 
 - `selectedItem`: The item that was just selected
-- `stateAndHelpers`: This is the same thing your `children` prop
+- `stateAndHelpers`: This is the exact same thing you're `children` prop
```
tansongyang: This should be "your". Also, this line is inconsistent with line 238. We should either keep "exact" or take it out, but it should be consistent.

README.md
briandipalma: Sorry, I think I changed the wrong line. I don't even remember making the change below (I tend to delete redundant text)!. So I'm going to close this PR.

Issue #200, opened 2017-09-25
Rendering in iframe - clicks don't register


paul-veevers:
#### What you did:
Rendering downshift inside a different window/document to where the Javascript is running - e.g. inside an iframe.

#### What happened:
Clicks on items don't register.

#### Reproduction repository:
https://codesandbox.io/s/x2ynm9mojw

#### Problem description:
It's due to ` window.addEventListener('mousedown'` and `window.addEventListener('mouseup'` in `componentDidMount`. Referencing the "correct" window (the one the component is in) solves the problem.

This may sound like a crazy use case, but it's quite common for 3rd party widgets, etc. I haven't tested but I assume the same is happening in a shadow dom.

#### Suggested solution:
Accept a prop called `context` or similar that defaults to `window`. I don't think it's possible to automatically get the parent window of a component?

As a note, there's also a reference to `document` in `getItemNodeFromIndex`, which will also need to be changed.

Happy to do a PR if that will help :)

kentcdodds:
Hi @paul-veevers!

Yeah, I'd be fine with your suggestion solution! Might even help push #185 along :)

paul-veevers:
See https://github.com/paypal/downshift/issues/200

Rendez:
I have the impression that this bug still exists within ShadowDOM. Can someone confirm this?

Issue #185, opened 2017-09-12
Implement support for React Native


gricard:
- `downshift` version: 1.5.0
- `node` version: 8.3.0
- `npm` (or `yarn`) version: yarn 1.0.1

Relevant code or config

```javascript
import React from 'react';
import { Input } from 'native-base';
import Downshift from 'downshift'

export default function AutocompleteInput({items, onChange}) {
    return (
        <Downshift onChange={onChange}>
            {({
                getInputProps,
                getItemProps,
                getRootProps,
                isOpen,
                inputValue,
                selectedItem,
                highlightedIndex
              }) => {
                return (
                    <Input {...getRootProps({refKey: "categoryTesting"})} {...getInputProps()} />
                );
            }}
        </Downshift>
    )
}
```

What you did:
Added an autocomplete input component to my React Native project and attached Downshift to it. 

What happened:

window.addEventListener is not a function. (In 'window.addEventListener('mousedown', onMouseDown)', 'window.addEventListener' is undefined)

componentDidMount
    downshift.cjs.js:699:30
measureLifeCyclePerf
    ReactNativeStack-dev.js:1610:15
<unknown>
    ReactNativeStack-dev.js:1657:33
notifyAll
    ReactNativeStack-dev.js:2121:107
close
    ReactNativeStack-dev.js:2138:8
closeAll
    ReactNativeStack-dev.js:1412:101
perform
    ReactNativeStack-dev.js:1388:52
batchedMountComponentIntoNode
    ReactNativeStack-dev.js:2004:24
perform
    ReactNativeStack-dev.js:1382:99
renderComponent
    ReactNativeStack-dev.js:2032:45
renderApplication
    renderApplication.js:34:4
runApplication
    AppRegistry.js:191:26
__callFunction
    MessageQueue.js:266:47
<unknown>
    MessageQueue.js:103:26
__guard
    MessageQueue.js:231:6
callFunctionReturnFlushedQueue
    MessageQueue.js:102:17

![image1](https://user-images.githubusercontent.com/4550801/30304109-cecbc3a6-9739-11e7-9d9f-72b0db31024f.PNG)

Reproduction repository:
https://github.com/gricard/downshift-reactnative-repro

Problem description:
Downshift does not currently have support for the React Native environment, which has some minor differences to running in the browser. 

Suggested solution:
I had a brief discussion about the issue with Kent, and he provided these three links which are areas where there will need to be changes in order to support React Native:

[...]

PR #204, opened 2017-09-26
Multi window context render

paul-veevers:
**What**:
Rendering into a different window context, e.g. an iframe. See https://github.com/paypal/downshift/issues/200

**Why**:
Clicks on items don't register.

**How**:
Accept an `eventTarget` prop that defaults to `window`

**Checklist**:
- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

Notes:
- Any ideas on how to test this in a nice way?
- I haven't touched the `set-a11y-status` file. Do you want me to change that as well? As it stands, the wrong `document` will be rendered into.
- Are you ok with the prop name? I couldn't think of anything both more flexible and descriptive.

ps. something that stumped me at the beginning was unresolved import errors from eslint. Turns out I needed to `npm i` from inside the `stories` folder.


src/downshift.js
```diff
@@ -148,7 +153,7 @@ class Downshift extends Component {
   }
 
   getItemNodeFromIndex = index => {
-    return document.getElementById(this.getItemId(index))
+    return this._rootNode.querySelector(`#${this.getItemId(index)}`)
```
kentcdodds: This won't work for folks rendering the menu outside the autocomplete root node I'm afraid. Need to set this as the eventTarget. Due to that fact, maybe eventTarget is not quite the right name... Maybe simply "window" ?

tansongyang: We could also consider "context", since the new docs say that the type can be `Element`/`document`/`window`.

kentcdodds: I'd be fine with that too, tough I hesitate because I worry folks might confuse that with React's context feature 🤔

tansongyang: Agreed. I realized right after I typed it that React also has "context".

Looked at the [thesaurus entry for "context"](http://www.thesaurus.com/browse/context). I'll throw out "environment".

I'll also suggest "renderContext".

kentcdodds: I'm good with either of those. Like the docs said, most people shouldn't be using this prop anyway.

paul-veevers: "environment" works well for other use cases like React Native (https://github.com/paypal/downshift/issues/185)

paul-veevers:
I've made it so `environment` is must be a `window` object. It makes assumptions in the code much easier, rather than having to do some ternary magic depending on whether it's a window or document or element.

tansongyang:
Are you comfortable with TypeScript? If so, would you mind also updating `typings/index.d.ts`?

paul-veevers:
Definitely not comfortable, but I gave it a go. Let me know if I got anything wrong.

typings/index.d.ts
```diff
@@ -26,6 +26,17 @@ export interface DownshiftProps {
   inputValue?: string
   highlightedIndex?: number
   id?: string
+  environment?: Environment
+}
+
+export interface Environment {
+  addEventListener: (type: string, listener: (event: Event) => void) => void
+  removeEventListener: (type: string, listener: () => void) => void
```
tansongyang: For both `addEventListener` and `removeEventListener`, I suggest just using `Function`  as the callback type (as the rest of this file does). This gives us the most flexibility.

typings/index.d.ts
```diff
@@ -26,6 +26,17 @@ export interface DownshiftProps {
...
+  document: Document
+}
+
+export interface Document {
+  getElementById: (id: String) => HTMLElement
```
tansongyang: Use the lowercase `string` here instead of `String`.

paul-veevers:
Thanks for the comments @tansongyang !

tansongyang:
@paul-veevers Thanks for the hard work!

bliitzkrieg:
This commit breaks SSR because window does not exist in node

tansongyang:
@bliitzkrieg What would you recommend? Is there a way to get this functionality while maintaining support for SSR? Also, are there any automated tests we could add that would catch this in the future? (I know very little about SSR.)

bliitzkrieg:
I believe the fix can be:

```
  static defaultProps = {
    defaultHighlightedIndex: null,
    defaultSelectedItem: null,
    defaultInputValue: '',
    defaultIsOpen: false,
    getA11yStatusMessage,
    id: generateId('downshift'),
    itemToString: i => (i == null ? '' : String(i)),
    onStateChange: () => {},
    onUserAction: () => {},
    onChange: () => {},
    environment: typeof window === 'undefined' ? {} : window,
  }
```

I tested locally and the tests all pass and my app renders on the server fine.

kentcdodds:
Your suggestion works fine for me.

I'll work on the SSR issue. I have some ideas. :+1:

kentcdodds:
See https://github.com/paypal/downshift/pull/205 :)

paul-veevers:
Thank you @kentcdodds and @tansongyang for making my first OSS PR a great experience :)

kentcdodds:
Sure thing! I'm glad you were able to start here!

bliitzkrieg:
Thanks @kentcdodds for looking into it so quickly! I really appreciate the hard work

Issue #206, opened 2017-09-29
handleStateChange confusion around controlling isOpen


jole78:
**Problem description:**
I'm trying to create a dropdown that contains a couple of choices that the user can select (a checkbox list one could say). It's basically a multiple select thingy.

So I wan't to keep the dropdown menu open while the user is interacting with it. 
So, I know that I then have to to take "control" over the isOpen prop and supply that to downshift. That all works fine.

The issue that I'm facing is that if the user clicks outside the downshift component (the root component) the menu should close.

I found this piece of code in the example:

```javascript
  handleStateChange = changes => {
    const {isOpen, type} = changes
    if (type === Downshift.stateChangeTypes.mouseUp) {
      this.setState({isOpen})
    }
  }
```
That seems to do the trick...but...for one it seems a little bit "complicated" (or how should I put it) and I really don't know why I need this and how it all works together.

**Suggested solution:**
Either better documentation about this type of situation or some other means to tell downshift that when the user clicks outside just close it, ignoring isOpen prop.
React Bootstrap has a prop called rootClose on its overlays that handles this for example.

kentcdodds:
Thanks for the issue @jole78, I knew we'd have to deal with this eventually.

Here's how things work... There are 4 parts of state in downshift: `isOpen`, `highlightedIndex`, `selectedItem`, and `inputValue`. We have an API that allows you to control the internal state of the component. When you're controlling any part of the state, then we need to only call `setState` for the parts of the state that downshift is controlling and we call `onStateChange` with anything that changed (whether or not it's controlled).

As noted above, when you control any state, we don't call `setState` internally and instead you're responsible for updating the state. In fact, we _can't_ control any state that you're controlling or we'll get into really confusing situations. This is consistent with controlled form fields. Due to this fact, you need to make sure you update your own version of the state based on whatever criteria you determine. This is why `onStateChange` exists. It's suggestions to you. You can use the information to update your own state or not.

There are [a lot of reasons state can change](https://github.com/paypal/downshift/blob/24134f38bec09cf839ba402528c23f9210fdebd1/src/downshift.js#L72-L94) and it's not easily determinable without the `type` property on the `changes` object. The `isOpen` state in particular has at least 5 ways to change (probably more, I didn't actually check). So that's why the `stateChangeTypes` exists.

It's undocumented because I didn't love the API and most folks can get a long way without needing to know it exists. But I think that things have been around long enough to say that no better API will materialize and we can set this one in stone.

[...]

PR #208, opened 2017-10-02
Enhance internal state control by providing onOuterClick prop

divyenduz:
This PR is not ready to be merged yet. When merged, this will fix https://github.com/paypal/downshift/issues/206

**What**:

Add an optional prop `onOuterClick` to better handle the internal state or anything else with downshift

**Why**:

This PR is required to enhance the developer experience around outer click and internal state handling with downshift with issues like https://github.com/paypal/downshift/issues/206

**How**:

This PR does the following (Will appreciate early feedback and comments):-

1. Add onOuterClick description to README. I can remove the link to a specific issue from there

2. Add onOuterClick to propTypes with type PropTypes.func

3. Call this.props.onOuterClick as a callback to reset function when it is called under onMouseUp event

4. Add a test case for newly added functionality. This is still incomplete. I started with the test case expecting it to fail in its current state where I open the menu and simulate an outside click with a click on document.body. With the current state of the onOuterClick callback, this should have resulted in isOpen to be false there by failing the test but it is not the case. Exploring it further but this makes this PR not "ready to be merged".

5. Also, to simulate the click on document.body. I borrowed the function `mouseDownAndUp` from `downshift.lifecycle.js`. I can either move this test over there or expose that function in a better way than duplicating.

6. I observed that `package.json` got formatted as soon as I run `npm install`. I have not checked that in with this PR. Not sure why that happened, I have validated that it happens with `npm install` and not via other action. Any pointers on that?

Let me know your feedback on the approach? Also, please point out if you believe that I understood the problem incorrectly.

**Checklist**:

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged 


tansongyang:
If you're comfortable with TypeScript, I would also suggest updating the [typings](https://github.com/paypal/downshift/blob/master/typings/index.d.ts). If not, don't worry; someone else can do it.

kentcdodds:
Ok, I've made a bunch of updates in the GitHub editor. We'll see if the build passes 😅

divyenduz:
I noticed that changes you made and they make sense to me. However, build fails because `onOuterClick` doesn't seem to be getting called. Do you have an idea of why that is happening? 

You made the documentation look so much better 🥇 

Also, let me know if I need to make edits to `typings` as well as suggested by @tansongyang. I noticed that `onClick` is also not defined in `typings`. I have no experience in TypeScript but this is pretty basic and I can do it. 

Let me know what should I do next though. 

Thanks

README.md
```diff
@@ -619,7 +619,7 @@ Thanks goes to these people ([emoji key][emojis]):
...
...umentation") | [<img src="https://avatars3.githubusercontent.com/u/746482?v=4" width="100px;"/><br /><sub>Divyendu Singh</sub>](https://divyendusingh.com)<br />[💡](#example-divyenduz "Examples") [💻](https://github.com/paypal/downshift/commits?author=divyenduz "Code") [📖](https://github.com/paypal/downshift/commits?author=divyenduz "Documentation") [⚠️](https://github.com/paypal/downshift/commits?author=divyenduz "Tests") | [<img src="https://avatars1.githubusercontent.com/u/841955?v=4" width="100px;"/><br /><sub>Muhammad Salman</sub>](https://github.com/salmanmanekia)<br />[💻](https://github.com/paypal/downshift/commits?author=salmanmanekia "Code") | [<img src="https://avatars3.githubusercontent.com/u/10820159?v=4" width="100px;"/><br /><sub>João Alberto</sub>](https://twitter.com/psicotropidev)<br />[💻](https://github.com/paypal/downshift/commits?author=psicotropicos "Code") | [<img src="https://avatars0.githubusercontent.com/u/16327281?v=4" width="100px;"/><br /><sub>Bernard Lin</sub>](https://github.com/bernard-lin)<br />[💻](https://github.com/paypal/downshift/commits?author=bernard-lin "Code") [📖](https://github.com/paypal/downshift/commits?author=bernard-lin "Documentation") |
```
kentcdodds: You're not going to want to edit this directly. It's automatically generated. If something needs to be updated, then update the `.all-contributorsrc` file :)

divyenduz: Noted, reverting and doing it via `.all-contributorsrc` 👍

kentcdodds:
What's really odd is the test that's failing isn't the new one but [this one](https://github.com/divyenduz/downshift/blob/enhance_on_outer_click/src/__tests__/downshift.lifecycle.js#L9-L49) and it's unclear to me how this simple change could have made that impact 🤔

kentcdodds:
Can you reproduce this breakage locally

README.md
divyenduz: This is now done via `npm run add-contributor` 👍

divyenduz:
Yes, I am able to get the breakage locally as well. Here is a small snipped from my terminal on running `npm run validate`

```
[test:cover]  FAIL  src/__tests__/downshift.lifecycle.js[test:cover]   ● Console[test:cover][test:cover]     console.error node_modules/react-dom/cjs/react-dom.development.js:8305
[test:cover]       The above error occurred in the <Downshift> component:[test:cover]           in Downshift[test:cover]           in Unknown (created by WrapperComponent)
[test:cover]           in WrapperComponent
[test:cover]
[test:cover]       Consider adding an error boundary to your tree to customize error handling behavior.
[test:cover]       You can learn more about error boundaries at https://fb.me/react-error-boundaries.
[test:cover]
[test:cover]   ● handles mouse events properly to reset state
[test:cover]
[test:cover]     expect(jest.fn()).toHaveBeenCalledTimes(1)
[test:cover]
[test:cover]     Expected mock function to have been called one time, but it was called zero times.
```

divyenduz:
Reading code with ☕️ to pinpoint the issue 👍

divyenduz:
I observed that if I call reset like this - callback without parameters and wrapper. 

```
this.reset(
          {type: Downshift.stateChangeTypes.mouseUp},
          this.props.onOuterClick,
        )
```

And remove the following from the new test case - since we are not passing any object after above change. 

```
expect(onOuterClick).toHaveBeenCalledWith(
    expect.objectContaining({
      // just verify that it's the controller object
      isOpen: false,
      getItemProps: expect.any(Function),
    }),
  )
```

Everything falls into place. 

So, I think this has to do with the way we are passing the callback with the object. 
I tried various configurations to pass the object with bind, not using implicit return with arrow function but nothing worked thus far breaking a different set of test cases on different configurations. 

Let me see if I can get around this in some way.

kentcdodds:
The problem is that there was no default for the `onOuterClick`. The way that `reset` works is it will default the given callback to an `noop` function if it's not provided, but since we provided one (`() => this.props.onOuterClick()`) that default wasn't being used and we tried to call a function that didn't exist. I'm not sure why it didn't give the full stacktrace though.

Should be fixed now though.

kentcdodds:
Thanks a ton @divyenduz!

kentcdodds:

divyenduz:
Thanks @kentcdodds - It is an honor to join the team. Time to read most of the code and watch the screencasts you had made while creating downshift with some ☕️

kentcdodds:
🎉

PR #215, opened 2017-10-04
add editable multi select example

slavab89:
**What**: Add a multi editable select tags example

**Checklist**:
- [X] Documentation
- [ ] Tests - N/A
- [X] Ready to be merged

In case some changes need to be made in the example, please let me know.
I'm not sure if its somehow possible to make the example editable by other people (In case something needs to be fixed / changed in the example in the future).. Or we'll just need to put a different link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -536,6 +536,7 @@ Examples exist on [codesandbox.io][examples]:
 - [Section/option group example](https://codesandbox.io/s/zx1kj58npl)
 - [Integration with `fuzzaldrin-plus` (Fuzzy matching)](https://codesandbox.io/s/pyq3v4o3j)
 - [Dropdown/select implementation with Bootstrap](https://codesandbox.io/s/53y8jvpj0k)
+- [Multiple editable tag selection](https://codesandbox.io/s/o4yp9vmm8z)
 
 <!--
 


mikeebee:
Hey @slavab89 / @kentcdodds. This example doesn't appear to be working for creation of new tags. The selection works fine but when I type something and press enter, nothing happens.

Any ideas?

kentcdodds:
That's because the example doesn't support enter, only tab. I've added enter support here. Was a pretty simple addition to `MultiDownshift.js` and I've added a comment to call it out :)

https://codesandbox.io/s/pk0kv34zm7

mikeebee:
Haha solid comment in that file, Sir! :D 

Thanks for that! Out of curiosity, was there a particular reason for favouring tab over enter? I'd not seen that as a convention for things like this.

kentcdodds:
I didn't make the example, but I'm assuming it's so the enter key can submit the form or something 🤷‍♂️ 

Could probably have both though (ie only use `event.preventDefault()` if the menu is open).

mikeebee:
Yeah I _think_ that's what most users would expect. Thanks for getting back to me, and thanks @slavab89 for the example.

slavab89:
😄 Just wanted to show an example of how to use it but yeah i believe Enter should be actionable as it seems natural to me.
btw, the example is not really updated to the latest version of the library and quite a lot has changed since (22+ versions) so the whole implementation may somewhat change based on new features and fixes.

@kentcdodds I saw that there are stories as well, should i try to migrate that example to a story (while basing the structure and solution on your other more recent examples?)

kentcdodds:
The stories are also useful examples, the problem is that it's harder for people contribute to and "fork" those examples and make changes to them than if we use codesandbox. So I prefer codesandbox generally.

PR #223, opened 2017-10-13
feat(disabled): if a disabled prop is passed, we won't return callback props

### LINKED ISSUE #191: disabled prop?
Greetings. I am building a Dropdown component based on Downshift and need to provide a way to disable it.  Disabled styles, disabled event handlers, etc.

Have you given any consideration to including a `disabled` prop to Downshift?  If true, the event handlers to open the menu would then be suppressed.

I can do this on my own by simply excluding the eventHandlers after calling `getButtonProps` or `getInputProps`.  It seems like something that might be useful in the core though.

kentcdodds:
**What**: This adds logic to `getButtonProps` and `getInputProps` to not
return callback props (like `onClick`) when the `disabled` prop is
passed.

**Why**: Otherwise IE11 will still fire the `onClick`, even if the
button's disabled (shakes fist in IE's general direction). Closes #191

**How**:
- Update and refactor some tests
- Add logic for the callback props to only be applied when the given
props don't include `disabled`.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 
- [x] Added @brentertz to contributors table 👍


src/downshift.js
```diff
@@ -441,13 +441,18 @@ class Downshift extends Component {
 
   getButtonProps = ({onClick, onKeyDown, ...rest} = {}) => {
     const {isOpen} = this.getState()
+    const callbacks = rest.disabled
```
pbomb: nitpicky, but `eventHandlers` would better describe the intent of this variable

src/downshift.js
```diff
@@ -516,6 +521,18 @@ class Downshift extends Component {
       ? 'onInput'
       : 'onChange'
     const {inputValue, isOpen, highlightedIndex} = this.getState()
+    const callbacks = rest.disabled
```
pbomb: change to `eventHandlers`?

kentcdodds: Yep :+1: good call

Issue #225, opened 2017-10-14
prop getters should return all current prop values


ellismarkf:
- `downshift` version: 1.11.1
- `node` version: 6.9.1
- `npm` (or `yarn`) version: 0.27.5

*Relevant code or config*

```javascript

import React from 'react';
import Downshift from 'downshift';

function AutocompleteWindowItem(props) {
  const {
    getItemProps,
    highlightedIndex,
    selectedItem,
  } = props.downshiftProps;
  return (
    <div
      className="JPMC__Autocomplete__item"
      {...getItemProps({ item: props.item })}
      style={{
        backgroundColor: highlightedIndex === props.index ? 'gray' : 'white',
        fontWeight: selectedItem === props.item ? 'bold' : 'normal',
      }}
    >
      {props.item}
    </div>
  );
}

function AutocompleteWindow(props) {
  const Item = props.renderItem;
  return (
    <div className="JPMC__Autocomplete__window">
      {props.items
        .map((item, index) => (
            <Item
              key={item}
              item={item}
              index={index}
              downshiftProps={props.downshiftProps}
            />
          )
        )}
    </div>
  )
}

function Nullable(props) {
  return props.isNull ? null : props.children;
}

function AutocompleteDOM(items, inputProps) {
  return downshift => (
    <div className="JPMC__Autocomplete__container">
      <input {...downshift.getInputProps(inputProps) } />
      <Nullable isNull={!downshift.isOpen}>
        <AutocompleteWindow
          items={items}
          renderItem={AutocompleteWindowItem}
          downshiftProps={downshift}
        />
      </Nullable>
    </div>
  );
}

function Autocomplete(props) {
  return (
    <Downshift {...props.downshiftProps}>
      {AutocompleteDOM(props.items, props.inputProps)}
    </Downshift>
  );
}

function TypeAhead(props) {
  return forwarded => (
    <Autocomplete
      inputProps={{name: forwarded.name, ...props.inputProps}}
      items={props.items}
      downshiftProps={props.downshiftProps}
    />
  );
}

function LabeledFormPart({ label, name, width, children, extraClasses, ...props }) {
  const formattedWidth = width
  ? `${width.value}${width.unit}`
  : null;
const style = width ? {width: formattedWidth} : null;
const baseClass = 'LabeledFormPart';
const classes = extraClasses ? `${baseClass} ${extraClasses}` : baseClass; 
  return (
    <div className={classes} style={style}>
      <label htmlFor={name}>{label}</label>
[...]

PR #228, opened 2017-10-16
Forward all props passed to downshift to the getStateAndHelpers function

Dajust:
This is the implementation of the idea detailed in this issue #225


pbomb:
I'd prefer that we not make this change given that the solution to #225, like @kentcdodds mentioned, is (in most cases?) simple to handle in the consumer with an arrow function to re-pass the desired props to the event handlers. Additionally, -- and this is not a great argument, but -- spreading all the props adds a little extra overhead for all consumers who don't need this functionality. Handling this in a reusable component consumer could look something like this:

```js
class Autocomplete extends React.Component {
  render() {
    return (
      <Downshift name={this.props.name} onChange={ (value) => this.props.onChange(name, value) }>{ ... }</Downshift>
    );
  }
}
```

If we do decide to allow this change, then this Pull Request will also need to add tests and TypeScript type definitions. The type definition changes are more difficult than usual due to the dynamic nature of forwarding on all props.

kentcdodds:
Well, we could avoid passing all the props if we pass the `id` in addition to the `itemToString`. Then @Dajust's desired outcome will be satisfied.

This would require a change to the section on props that I'll not inline.

README.md
```diff
@@ -518,6 +517,10 @@ These are values that represent the current state of the downshift component.
 | `isOpen`           | `boolean`         | the menu open state                            |
 | `selectedItem`     | `any`             | the currently selected item input              |
 
+### props
+
+As a convenience, all props you pass to <Downshift /> are forwarded along to you. You can access all props via the `getStateAndHelpers` function.
```
kentcdodds: The `getStateAndHelpers` function is not exposed, so we shouldn't mention that. Just remove this sentence:

> You can access all props via the `getStateAndHelpers` function.

You could change this section to say:

```
### props

As a convenience, the `id` and `itemToString` props which you pass to `<Downshift />` are available here as well.
```

src/downshift.js
```diff
@@ -332,7 +332,6 @@ class Downshift extends Component {
 
   getStateAndHelpers() {
     const {highlightedIndex, inputValue, selectedItem, isOpen} = this.getState()
-    const {itemToString} = this.props
```
kentcdodds: Let's restore this line and destructure out the `id` from here too

src/downshift.js
```diff
@@ -352,6 +351,7 @@ class Downshift extends Component {
     } = this
     return {
       // prop getters
+      ...this.props,
```
kentcdodds: Let's remove this.

src/downshift.js
```diff
@@ -369,7 +369,6 @@ class Downshift extends Component {
       setHighlightedIndex,
       clearSelection,
       clearItems,
-      itemToString,
```
kentcdodds: Let's restore this and add the `id` below it.

Then above this line let's leave a blank line and then a comment that says: `// props` like the other sections.

PR #229, opened 2017-10-17
fix: cursor jumping to end of input

### LINKED ISSUE #217: Cursor jumps to end of input when inputValue is a controlled prop
- `downshift` version: 1.11.0
- `node` version: 6.9 (and codesadbox.io, could not find out, what they're using)
- `npm` (or `yarn`) version: yarn 1.1.0 (and whatever codesandbox.io uses)

Problem description:
- Visit this [codesandbox](https://codesandbox.io/s/3qnr4j4lnq)
- type text
- move cursor to the left
- type a letter
- the letter will be added at the correct position, but the cursor will move to the end of the input instead of staying where you have typed

Relevant code or config

```javascript
class Demo extends React.Component {
  state = {
    inputValue: ""
  };

  render() {
    return (
      <div>
        <Downshift
          onStateChange={data => {
            return (
              data.hasOwnProperty("inputValue") &&
              this.setState({ inputValue: data.inputValue })
            );
          }}
          inputValue={this.state.inputValue}
        >
          {({ getInputProps }) => (
            <div>
            </div>
          )}
        </Downshift>
      </div>
    );
  }
}
```

robin-drexler:
#217

**What**:
Provide an onInputValueChange callback to allow to update controlled prop before its old value is used to re-render the children, causing the cursor in the input to jump to the end in the next render.

**Why**:
fixes #217 

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged -> docs missing 

@kentcdodds I'd like to get your opinion on this change before adding documentation for it.
Do you think this is a direction worth pursuing?

### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -31,6 +31,7 @@ class Downshift extends Component {
     itemToString: PropTypes.func,
     onChange: PropTypes.func,
     onStateChange: PropTypes.func,
+    onInputValueChange: PropTypes.func,
     onUserAction: PropTypes.func,
     onClick: PropTypes.func,
     onOuterClick: PropTypes.func,
@@ -62,6 +63,7 @@ class Downshift extends Component {
     id: generateId('downshift'),
     itemToString: i => (i == null ? '' : String(i)),
     onStateChange: () => {},
+    onInputValueChange: () => {},
     onUserAction: () => {},
     onChange: () => {},
     onOuterClick: () => {},
@@ -268,6 +270,14 @@ class Downshift extends Component {
   internalSetState(stateToSet, cb) {
     let onChangeArg
     const onStateChangeArg = {}
+
+    if (stateToSet.hasOwnProperty('inputValue')) {
+      this.props.onInputValueChange(
+        stateToSet.inputValue,
+        this.getStateAndHelpers(),
+      )
+    }
+
     return this.setState(
       state => {
         state = this.getState(state)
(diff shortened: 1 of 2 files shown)


src/downshift.js
```diff
@@ -268,6 +270,14 @@ class Downshift extends Component {
   internalSetState(stateToSet, cb) {
     let onChangeArg
     const onStateChangeArg = {}
+
+    if (stateToSet.hasOwnProperty('inputValue')) {
```
kentcdodds: Better be careful here. `stateToSet` can be a function. Also, `this.getStateAndHelpers()` will get us the _old_ state, not any of the new state. So let's move this down to right before `return nextState` and then we can do:

```javascript
this.props.onInputValueChange(
  stateToSet.inputValue,
  {...this.getStateAndHelpers(), ...nextState},
)
```

Thoughts?

robin-drexler: Wouldn't this cause `inputValue` being read from stale props again?

kentcdodds: Nope, because `nextState` will have the `inputValue`.

But actually, it should probably be this instead:

```javascript
this.props.onInputValueChange(
  stateToSet.inputValue,
  {...this.getStateAndHelpers(), ...stateToSet},
)
```

And we could put it right below `stateToSet.type = ...`

robin-drexler:
Sorry, I need to close this PR. github is not updating new commits in here. I can't figure out why. Will open a new one.

src/downshift.js
robin-drexler: [Tried that here](https://github.com/paypal/downshift/pull/230/commits/c7d998fc81835898437bae97d5b6ab308ba2352f), but placing the `onInputValueChange` inside the setState function, causes the children to re-render with the old value once, still leading to the cursor jumping to the end.

PR #230, opened 2017-10-17
Fix cursor jumping to end of input

### LINKED ISSUE #217: Cursor jumps to end of input when inputValue is a controlled prop

robin-drexler:
#217

Continuation of https://github.com/paypal/downshift/pull/229

**What**:
Provide an onInputValueChange callback to allow to update controlled prop before its old value is used to re-render the children, causing the cursor in the input to jump to the end in the next render.

**Why**:
fixes #217 

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged -> docs missing 

@kentcdodds I'd like to get your opinion on this change before adding documentation for it.
Do you think this is a direction worth pursuing?

### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -270,9 +270,14 @@ class Downshift extends Component {
   internalSetState(stateToSet, cb) {
     let onChangeArg
     const onStateChangeArg = {}
-    const isSetToStateFunction = typeof stateToSet === 'function'
+    const isStateToSetFunction = typeof stateToSet === 'function'
 
-    if (!isSetToStateFunction && stateToSet.hasOwnProperty('inputValue')) {
+    // we want to call `onInputValueChange` before the `setState` call
+    // so someone controlling the `inputValue` state gets notified of
+    // the input change as soon as possible. This avoids issues with
+    // preserving the cursor position.
+    // See https://github.com/paypal/downshift/issues/217 for more info.
+    if (!isStateToSetFunction && stateToSet.hasOwnProperty('inputValue')) {
       this.props.onInputValueChange(stateToSet.inputValue, {
         ...this.getStateAndHelpers(),
         ...stateToSet,
@@ -281,7 +286,7 @@ class Downshift extends Component {
     return this.setState(
       state => {
         state = this.getState(state)
-        stateToSet = isSetToStateFunction ? stateToSet(state) : stateToSet
+        stateToSet = isStateToSetFunction ? stateToSet(state) : stateToSet
         // this keeps track of the object we want to call with setState
         const nextState = {}
         // this is just used to tell whether the state changed
@@ -319,7 +324,9 @@ class Downshift extends Component {
           }
         })
 
-        if (isSetToStateFunction && stateToSet.hasOwnProperty('inputValue')) {
+        // if stateToSet is a function, then we weren't able to call onInputValueChange
+        // earlier, so we'll call it now that we know what the inputValue state will be.
+        if (isStateToSetFunction && stateToSet.hasOwnProperty('inputValue')) {
           this.props.onInputValueChange(stateToSet.inputValue, {
             ...this.getStateAndHelpers(),
             ...stateToSet,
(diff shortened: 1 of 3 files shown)


robin-drexler:
I've added a story (to be removed again) to showcase the issue and to tinker around with.
I tried to move the `onInputValueChange` callback into the setState function, but that caused the initial issue to show up again.

src/downshift.js
```diff
@@ -287,6 +290,14 @@ class Downshift extends Component {
           onChangeArg = stateToSet.selectedItem
         }
         stateToSet.type = stateToSet.type || Downshift.stateChangeTypes.unknown
+
+        if (stateToSet.hasOwnProperty('inputValue')) {
```
kentcdodds: Unfortunately, this wont work with [when we call `reset`](https://github.com/paypal/downshift/blob/1385fc9dc5bc5e93ef68c863c971946a3e29db63/src/downshift.js#L621) which users can do themselves _or_ when the user clicks outside the menu... 🤔 This is because `stateToSet` in that case is a function, but it does set the `inputValue`. So I'm thinking that we can do this, but we _also_ have to handle the case where `stateToSet` is a function. I think the best thing to do would be to test this by calling `reset` and make sure that `onInputValueChange` is called in that case (in addition to the other tests you already have). Then we'll figure out the best way to implement that.

robin-drexler: Did you see my comment here? https://github.com/paypal/downshift/pull/229#discussion_r145216365
I'm not sure yet that invoking the `onInputValueChange` callback inside the `setState` function works at all to fix the issue at hand. I'd like to prove that first before addressing issues like the different incarnations of  `stateToSet`.

Sorry for the confusion with two opened PRs. :/

kentcdodds: I did see your comment. But as I said, the existing code will _never_ call `onInputValueChange` when `reset` is called. We definitely need to call `onInputValueChange` any time the `inputValue` is called.

I think that the solution is good as is, then we add _more_ code for the case when the `stateToSet` is a function (which only applies in the `reset` case). Luckily for us, when `reset` happens, the original issue won't be a problem.

robin-drexler: The issue I'm having is that calling `onInputValueChange` inside `setState` function does not appear to fix the issue with the cursor jumping to the end of the input. 

Or am I misunderstanding and with "the existing code" you were referring to my original version, where the `onInputValueChange` call was placed above the `setState`?

Sorry, I'm a bit confused right now.

kentcdodds: Ok, so the code you have already is great. We're not going to change it.

The problem is that the code you have doesn't cover the `reset` case because `stateToSet` is a function in that case (and it has to be that way). So if we merged this PR as is it would work great for everything except for when the user clicks outside the menu or tabs out of the menu (which is when `reset` is called).

To deal with this issue, we need more code. We need to handle the case where `stateToSet` is a function. If it is, then we need to do that inside of `setState`.

Luckily, when `stateToSet` is a function, then we wont have the problem of the cursor jumping to the end of the input because the focus will be elsewhere anyway.

Does that make sense?

robin-drexler: Yes, it does. Thanks for clearing things up! :)

robin-drexler:
Implemented the function version of `stateToSet` as well and added a test for it.

kentcdodds:

robin-drexler:
Thank you so much for your help. :)
I'll open a new PR for the contributions table.

PR #231, opened 2017-10-17
fix(a11y-status): escape HTML in statuses

usandfriends:
**What/Why**:
Potential XSS attack in generation of a11y status messages. Need to escape input before rendering those.

**How**:
Escape input by building the a11y status div using DOM API and fill in status message using `textContent`, which automatically escapes text (wanted to use `innerText` but it's not supported in Enzyme tests).

When using the DOM API, previous children will need to removed manually. I opted for using a loop and `Node.removeChild` over `Node.innerHTML = ''` because the former is more performant.

**Checklist**:
- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged>


kentcdodds:
This has been released. **Everyone should upgrade to `downshift@1.13.1` as soon as possible.**

Issue #232, opened 2017-10-18
onChange not called when selectedItem is already set to the inputs value


robin-drexler:
- `downshift` version: 1.13
- `node` version: codesandbox
- `npm` (or `yarn`) version: codesandbox

Problem description:

Go to the [typeahead example](https://codesandbox.io/s/82m2px40q9)

 - Type "orange"
- Wait for the selection list to appear.
- Select "orange"

The alert "You selected orange" does not appear, because `onChange` is not triggered.
Probably caused by: https://github.com/paypal/downshift/blob/master/src/downshift.js#L283

Suggested solution:
Would anything speak against calling `onChange` even if the value did not change, as long as `selectedItem` is part of `stateToSet`? Or maybe allow by flag?

kentcdodds:
Thanks @robin-drexler,
Hmm... The `onChange` shouldn't happen if the selected item is the same (it doesn't change). I don't think I want to allow it by a flag either. I think that would add unnecessary complexity. You should be able to get it from `onStateChange` though. Do you think you could workaround this issue?

robin-drexler:
thanks @kentcdodds for the quick reply! :)

Maybe if `stateChangeTypes` were considered stable and clicking on an item would produce a type that is not unknown, it might be possible to workaround, but I'm not sure yet (especially considering the `inputValue` might already be updated at this point once #230 is merged).

Regardless of this particular issue: 
Have you thought about switching from `onChange` to `onSelect`? onSelect could always be called when an item is selected, regardless of prior values. If yes, what were the reasons to go against it? It's one of the things that came as a surprise to me when I first discovered downshift.

kentcdodds:
Hmmm... What if you added a `onClick` handler to your items? `<div {...getItemProps({item, onClick: handleItemSelection})} />`

I'd just really prefer to workaround this issue as I don't think it's a very common situation.

robin-drexler:
I guess that could work if I also added a `keyDown` handler for the input field to also account for the user picking by keyboard. I'd probably call `selectItemAtIndex` with `highlightedIndex`. But that feels like re-implementing a large portion of the downshift api. Will need to think about it, but we can close the issue here if a fix/change won't be accepted to address the issue directly in downshift.

robin-drexler:
Would you be open to some sort of `onResetRequested` callback that would be called in `reset` and return state to be merged with Downshift's defaults?

This way one could override `inputValue`, but keep the rest as is.

Something like:

> downshift.js

```js
[...]

PR #239, opened 2017-10-23
feat: add onSelect callback

robin-drexler:
**What**:
callback is called whenever an item is selected, regardless of prior selected item.

https://github.com/paypal/downshift/issues/232

**Why**:
Allows to react to every selection, even when the value does not change: https://github.com/paypal/downshift/issues/232

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ and the new state of `downshift`. (see `onStateChange` for more info on
 
 > `function(selectedItem: any, stateAndHelpers: object)` | optional, no useful default
 
-Called when the user selects an item, regardless of the previous selected item. `onChange` is only called when the selected item is different from the previously selected one.
+Called when the user selects an item, regardless of the previous selected item.
  Called with the item that was selected
 and the new state of `downshift`. (see `onStateChange` for more info on
 `stateAndHelpers`).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -215,7 +215,20 @@ properties:
...
+
+- `selectedItem`: The item that was just selected
+- `stateAndHelpers`: This is the same thing your `children` prop
+  function is called with (see [Child Callback Function](#child-callback-function))
+
+### onSelect
+
+> `function(selectedItem: any, stateAndHelpers: object)` | optional, no useful default
+
+Called when the user selects an item, regardless of the previous selected item. `onChange` is only called when the selected item is different from the previously selected one.
```
tansongyang: Consider putting docs about `onChange` under `onChange`. I don't think it needs to be here in the `onSelect` section.

robin-drexler: I thought it could make sense since it might not be obvious what the difference between `onChange` and `onSelect` is at a glance. But I don't mind removing it again. (it's already documented in `onChange`

tansongyang: I think that's a valid point. I thought things were clear enough without it. But that's why I said "consider" instead of "please do this", so you can do what seems best.

robin-drexler:
That was a quick merge and release. Thanks a lot! :)

Issue #235, opened 2017-10-20
validateGetRootPropsCalledCorrectly does not correctly check if the root props are applied


yp:
- `downshift` version: `1.14.0`
- `node` version: 
- `npm` (or `yarn`) version:

Reproduction repository:
https://codesandbox.io/s/23yzon246r

Problem description:
I would like to use a non-DOM component as root component returned by the children function. I pass to it a prop `rootProps` containing the result of `getRootProps()` and it spreads the object to the inner root `div`. In such a way, the root props that Downshift needs are correctly applied. However the `validateGetRootPropsCalledCorrectly` throws an error.

Suggested solution:
I would suggest to either remove the check or to add a prop `excludeSafetyChecks` that allows to bypass such a check.

tansongyang:
I believe the problem is that you are not using `refKey`. Please see the documentation on [getRootProps](https://github.com/paypal/downshift#getrootprops).

Change `MyDiv` like so:

```jsx
const MyDiv = ({innerRef, ...rest}) => <div ref={innerRef} {...rest} />
```

And then change your call site:

```jsx
<MyDiv {...getRootProps({refKey: 'innerRef'})}>
   {/* ... */}
</MyDiv>
```

Here's a working example: https://codesandbox.io/s/m4n2416l1x

Closing now, as I believe this resolves the issue. If not, let me know and we can keep discussing.

yp:
In my case I cannot change the definition of MyDiv. Actually, the real case is more complicated than the example (that I greatly simplified in the codesandbox).

tansongyang:
OK. Is there no way to make `refKey` work in your scenario?

yp:
What about transforming the `throw new Error`s into `console.error` and surrounding the function call by a `if (NODE_ENV!=='production') `?

kentcdodds:
If we did that then the component would just not work at all. Failing early and obviously is better. Any reason you couldn't wrap `<MyDiv />` in a `<div />` [like this](https://codesandbox.io/s/k09j74mp5)?

yp:
I understand your position, but the check is too simplicistic and it is failing even if the component would work because the ref is set correctly to the root component.
IMHO a failure when the component should not be failing is an issue, but if you do not agree, I will find a workaround.

kentcdodds:
But the component doesn't work. With the way refs work, downshift can't get the root DOM node without being able to pass a ref prop to the node itself. It's not enough to pass a ref to a composite component.

[...]

PR #241, opened 2017-10-24
Add the option `suppressRefError` to `getRootProps`.

yp:
Fixes paypal/downshift#235.

**What**:
The option `suppressRefError` has been added to `getRootProps` to bypass the ref check when it erroneously raises an error.

**Why**:
As discussed in paypal/downshift#235, a ref could be correctly forwarded to the root DOM element of a composite child but the ref check could still fail.
The new option allows to bypass such a check.

**How**:
A new optional argument of `getRootProps` has been introduced.
This argument must be an object and if it has the key `suppressRefError` and if the corresponding value evaluates to `true` the ref check is bypassed (but the check that `getRootProps` has been called is still performed).
The introduction of the second argument of `getRootProps` instead of using a key in the first argument is intentional in order to clarify that it is not a prop forwarded to the root component.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 


kentcdodds:
I'm going to go ahead and ask that we wait for another maintainer to review this as well 👌 thank you!

kentcdodds:

.all-contributorsrc
```diff
@@ -370,6 +370,16 @@
...
       ]
+    },
+    {
+      "login": "yp",
+      "name": "yp",
+      "avatar_url": "https://avatars1.githubusercontent.com/u/275483?v=4",
+      "profile": "http://algolab.eu/pirola",
+      "contributions": [
+        "bug",
+        "code"
```
kentcdodds: You could add `"test"` to this too if you want. Just update the file and commit. It'll update the readme in a commit hook 👍

kentcdodds:
Sorry, I'm not sure what happened with this PR! Do you mind resolving conflicts and I'll get it merged.

Easiest way to resolve all-contributors conflicts is to just fix the `.all-contributorsrc` by hand, then commit, the README will fix itself!

PR #242, opened 2017-10-25
fix(selectItemAtIndex): allow selecting empty string

tobiasandersen:
**What**:
I'm currently unable to select an item if it's only an empty string.

**Why**:
I'm creating a select input using downshift (which I really enjoy btw, so big props!), and want it to work similar to the native html5 select input. But in downshift I can't select an item if it's an empty string. I could work around it in my implementation, but to me it seems reasonable for downshift to allow the empty string as a value.

Here's a codesandbox for comparison: 
https://codesandbox.io/s/r0kyyx0y1q

**How**:
Only escape from selecting a new item if `item == null`.

**Checklist**:

- [ ] Documentation "N/A"
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/src/__tests__/downshift.misc.js
+++ b/src/__tests__/downshift.misc.js
@@ -75,8 +75,15 @@ test('toggleMenu can take no arguments at all', () => {
 })
 
 test('clearItems clears the all items', () => {
-  const items = ['Chess']
-  const {wrapper, clearItems} = setup({items})
+  const item = 'Chess'
+  const children = ({getItemProps}) => (
+    <div>
+      <div key={item} {...getItemProps({item})}>
+        {item}
+      </div>
+    </div>
+  )
+  const {wrapper, clearItems} = setup({children})
   clearItems()
   expect(wrapper.instance().items).toEqual([])
 })
--- a/src/__tests__/downshift.misc.js
+++ b/src/__tests__/downshift.misc.js
@@ -32,6 +32,24 @@ test('selectItemAtIndex does nothing if there is no item at that index', () => {
   expect(childSpy).not.toHaveBeenCalled()
 })
 
+test('selectItemAtIndex can select item that is an empty string', () => {
+  const items = ['Chess', '']
+  const children = ({getItemProps}) => (
+    <div>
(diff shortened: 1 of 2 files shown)


kentcdodds:
Thanks so much!

kentcdodds:

tobiasandersen:
Thanks for the quick review! I've added a test, let me know if there's anything else needed.

tobiasandersen:
I noticed that [another test](https://github.com/paypal/downshift/blob/master/src/__tests__/downshift.misc.js#L59-L64) in that file isn't actually testing anything:

```js
test('clearItems clears the all items', () => {
  const items = ['Chess']
  const {wrapper, clearItems} = setup({items})
  clearItems()
  expect(wrapper.instance().items).toEqual([])
})
```

The [setup](https://github.com/paypal/downshift/blob/master/src/__tests__/downshift.misc.js#L109-L117) function doesn't handle the `items` passed in, so no items will ever get created (that's why the test is still successful). 

**Solutions:**

1. Make the `setup` function handle items, similar to what you do in [this setup function](https://github.com/paypal/downshift/blob/master/src/__tests__/downshift.get-input-props.js#L282-L299). If you want this, I could update my test in this PR, along with the `setup` function. But there's another test that currently passes `children`, so that would also have to be handled still.

2. Make the `clearItems` test do what I did here, and just pass `children` (since `setup` handles that). Would you want that fix in this, or in a new, PR?

kentcdodds:
Hey @tobiasandersen! Thank you for noticing that! I'm good with either of your solutions :) Those `setup` functions each got written at different times with different use cases, so this kind of mistake was easy for someone to make and a reviewer to miss 😅 

Feel free to fix it in this PR. Thank you again!

tobiasandersen:
Okay I opted with the second solution since it changed less.

kentcdodds:
Would you like to [add yourself](../blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor) to the [contributors table](../#contributors)? Feel free to do so in another PR :+1:

PR #247, opened 2017-10-31
fix(scrollIntoView): update scroll position when controlled prop highlightedIndex changes

arahansen:
**What**: update internal highlightedIndex when controlled prop changes

**Why**: if highlightedIndex changes, then scroll position will be updated with the previous controlled prop value

**How**: add logic to `componentDidUpdate` lifecycle hook

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/src/__tests__/downshift.lifecycle.js
+++ b/src/__tests__/downshift.lifecycle.js
@@ -99,6 +99,22 @@ test('props update of selectedItem will update the inputValue state', () => {
   )
 })
 
+test('props update of highlightedIndex will update highlighted index', () => {
+  const childSpy = jest.fn(() => null)
+  const wrapper = mount(
+    <Downshift highlightedIndex={1}>
+      {childSpy}
+    </Downshift>,
+  )
+  childSpy.mockClear()
+  wrapper.setProps({highlightedIndex: 0})
+  expect(childSpy).toHaveBeenCalledWith(
+    expect.objectContaining({
+      highlightedIndex: 0,
+    }),
+  )
+})
+
 function mouseDownAndUp(node) {
   node.dispatchEvent(new window.MouseEvent('mousedown', {bubbles: true}))
   node.dispatchEvent(new window.MouseEvent('mouseup', {bubbles: true}))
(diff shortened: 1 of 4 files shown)


src/downshift.js
```diff
@@ -622,6 +622,11 @@ class Downshift extends Component {
         inputValue: this.props.itemToString(this.props.selectedItem),
       })
     }
+
+    if (this.props.highlightedIndex !== prevProps.highlightedIndex) {
```
kentcdodds: We should probably add a check for whether `highlightedIndex` is a state prop as well (like we do with `selectedItem` above).

src/downshift.js
```diff
@@ -622,6 +622,11 @@ class Downshift extends Component {
...
+      this.setHighlightedIndex()
```
kentcdodds: Rather than calling `setHighlightedIndex`, I'd prefer to extract [this callback](https://github.com/arahansen/downshift/blob/2286e064cad768b43754c42b195430b0d5e8b09a/src/downshift.js#L133-L137) into a new function called "scroll highlighted item into view" and call that here. Makes it more explicit.

src/__tests__/downshift.lifecycle.js
```diff
@@ -99,6 +99,22 @@ test('props update of selectedItem will update the inputValue state', () => {
   )
 })
 
+test('props update of highlightedIndex will update highlighted index', () => {
```
kentcdodds: I think this test passes without your changes. I think we should change this test to make sure that the scroll position changes properly. Might take a little more work..

src/downshift.js
arahansen: I agree, that seems better! Will update.

arahansen:
not sure why my diff is so large? From what I can tell the contribtutors that are being "added" in this PR already show up in master. Is this github diffing going haywire?

kentcdodds:
If you look at your commit, it actually has those changes in it. I suggest creating a brand new branch based on this upstream repo's `master` and making your changes to that one, then you can push that to this PR branch. Should update without trouble.

PR #251, opened 2017-11-04
fix(onInputValueChange): only update internal state if inputValue has changed

arahansen:
**What**: check that the actual intended value of `inputValue` has changed

**Why**: If `selectedItem` is always a new object, then the check in componentDidUpdate will always return true, resulting in `onInputValueChange` being called in an endless loop. [See repro here](https://codesandbox.io/s/wqky51ql68)

**How**: update `componentDidUpdate`

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 


arahansen:
My understanding was if the itemToString value didn’t change, then the inputValue state shouldn’t change. Is that not the case?

arahansen:
For what it’s worth, the common use case for my project is to generate the list of items during render. So all items, and selectedItem would be a new object on each render.

This in my experience is a pretty common pattern in general.

arahansen:
Thinking more on what you said @kentcdodds, I agree. What if we split the `internalSetState` between both checks? Something like this: 

```js
componentDidUpdate(prevProps) {
    const itemString = this.props.itemToString(this.props.selectedItem)
    const prevItemString = this.props.itemToString(prevProps.selectedItem)

    if (
      this.isControlledProp('selectedItem') &&
      this.props.selectedItem !== prevProps.selectedItem
    ) {
      this.internalSetState({
        type: Downshift.stateChangeTypes.controlledPropUpdatedSelectedItem,
        
      })

      if (itemString !== prevItemString) {
        this.internalSetState({
          inputValue: itemString,
        }
      }
    }

    this.updateStatus()
  }
```

kentcdodds:
Hmmm... I was thinking something more like this:

```javascript
componentDidUpdate(prevProps) {
  const itemString = this.props.itemToString(this.props.selectedItem)
  const prevItemString = this.props.itemToString(prevProps.selectedItem)

  if (
    this.isControlledProp('selectedItem') &&
    this.props.selectedItemChanged(this.props.selectedItem, prevProps.selectedItem)
  ) {
    this.internalSetState({
      type: Downshift.stateChangeTypes.controlledPropUpdatedSelectedItem,
      inputValue: this.props.itemToString(this.props.selectedItem),
    })
  }

  this.updateStatus()
}
```

With a `defaultProps` entry for `selectedItemChanged` that would be the current behavior (because I think that's more common and to avoid a breaking change):

```js
selectedItemChanged: (selectedItem, previousItem) => selectedItem !== previousItem
```

Thoughts?

arahansen:
I don't mind adding this, but it's unfortunate that it's pretty easy to fall into an infinite update loop with such a trivial example. It would be nice to have something in place to protect against this.

kentcdodds:
> it's unfortunate that it's pretty easy to fall into an infinite update loop with such a trivial example

Not sure I follow. Could you explain?

arahansen:
If you [look at my repro here](https://codesandbox.io/s/wqky51ql68), you'll see that it only takes a selectedItem that consists of an object defined in render, coupled with a onInputValueChange function to enter into an endless update loop. 

It seems wrong to me that `onInputValueChange` is called if the inputValue didn't actually change.

arahansen:
I would be ok with merging this as-is, since I do think its an improvement, and unblocks me for my use case.

If you're interested, we can keep the discussion on this going after the PR is merged.

src/__tests__/downshift.lifecycle.js
```diff
@@ -92,6 +92,22 @@ test('props update of selectedItem will update the inputValue state', () => {
...
 
+test('props update of selectedItem will not update inputValue state', () => {
+  const childSpy = jest.fn(() => null)
+  const wrapper = mount(
+    <Downshift
+      onInputValueChange={childSpy}
+      selectedItemChanged={(prevItem, item) => prevItem.id !== item.id}
+      selectedItem={{id: '123', value: 'wow'}}
+    >
+      {() => null}
```
kentcdodds: Shouldn't this be childSpy?

src/__tests__/downshift.lifecycle.js
```diff
@@ -92,6 +92,22 @@ test('props update of selectedItem will update the inputValue state', () => {
   )
 })
 
+test('props update of selectedItem will not update inputValue state', () => {
+  const childSpy = jest.fn(() => null)
+  const wrapper = mount(
+    <Downshift
+      onInputValueChange={childSpy}
```
kentcdodds: Oh... This should be called something more accurate I think.

Issue #253, opened 2017-11-08
Replace scrollIntoView with scroll-into-view-if-needed


kentcdodds:
I'd like to reduce the amount of code to maintain and `scrollIntoView` is probably the most complicated code in this project. [Richard pointed me](https://twitter.com/richardszalay/status/928128167117701120) to [`scroll-into-view-if-needed`](https://github.com/stipsan/scroll-into-view-if-needed). Looks like it could be a good replacement.

Anyone want to make a PR to [replace `scrollIntoView`](https://github.com/paypal/downshift/blob/22ed3d7e9be53320a94d41500ebccd611d1cac53/src/utils.js#L50) and see how it impacts the examples?

Drapegnik:
It would be interesting to me, but at the moment there is no time. I can try to do it during the week

razh:
Note that [`scroll-into-view-if-needed`](https://unpkg.com/scroll-into-view-if-needed@1.2.8/dist/) (`bundle.js`is 9.51 kB unminified) includes the [amator](https://github.com/anvaka/amator) animation library [(8.13 kB unminified)](https://unpkg.com/amator@1.1.0/dist/).

kentcdodds:
Ah, that's good to know @razh! Thanks for bringing that up! We definitely don't want that :-/

Maybe we should just take ours and extract it to an isolated package if we can't find something else? I feel like there's gotta be a good, small solution to this!

Drapegnik:
@kentcdodds how you see the best way:
1. fork `scroll-into-view-if-needed`, get rid of `amator` and:
    1. use it by direct link on github fork repo?
    2. publish new package and use it?
2. completely extract source code as new package

In first case we'll still be able to update ours fork from parent

kentcdodds:
Well, if we can't find anything better than what we have built ourselves, then I say let's make a new package out of what we have.

Alternatively we could just not worry about it and leave things as they are 😅 Maybe I'm just complicating things...

Drapegnik:
How about use [`patch-package`](https://github.com/ds300/patch-package)?

kentcdodds:
Interesting! I've never heard of that before! That is _definitely_ overkill though. I thought about this as a passing idea that would make downshift a little easier to maintain. It's not really a big sticking point for me, and I feel like doing something like that would overcomplicate things...

stipsan:
[...]

PR #259, opened 2017-11-17
WIP: use scroll-int-view-if-needed

kentcdodds:
**What**: See #253 

**Why**: Reduce maintenance cost of this area of the code by using an existing solution

**How**: swap our implementation out with `scroll-into-view-if-needed/dist/calculate`

**Status/Thoughts**:
- I think this will increase bundle size. Not sure by how much, but I'm thinking we probably wont do it if it it's by much...
- The build and tests fail currently

Build errors:

```
(!) `this` has been rewritten to `undefined`
https://github.com/rollup/rollup/wiki/Troubleshooting#this-is-undefined
node_modules/scroll-into-view-if-needed/dist/calculate.js
1: var __assign = (this && this.__assign) || Object.assign || function(t) {
                   ^
2:     for (var s, i = 1, n = arguments.length; i < n; i++) {
3:         s = arguments[i];
...and 1 other occurrence
```

Test errors:

```
/Users/kdodds/Developer/downshift/node_modules/scroll-into-view-if-needed/dist/calculate.js:14
export default function calculate(target, options) {
^^^^^^
    
SyntaxError: Unexpected token export

```

Some discussion around this on: https://github.com/stipsan/scroll-into-view-if-needed/pull/152

> Ok, so just FYI, I just tried:
> 
> ```javascript
> import calculate from 'scroll-into-view-if-needed/dist/calculate'
> ```
> 
> And jest failed to run because babel doesn't (and shouldn't) apply to `node_modules`. So here's what I suggest:
> 
> Create a `calculate/package.json` file where you have entries for `main` and `module` that point to a CJS version of the calculate file and the ESM version of the calculate file. Then users can simply do:
> 
> ```javascript
> import calculate from 'scroll-into-view-if-needed/calculate'
> ```
> 
> And it will work in node _and_ with tools like rollup/webpack.
> 
> Thoughts?

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests WIP
- [ ] Ready to be merged

cc @stipsan


stipsan:
Hi! 

I've just published `scroll-into-view-if-needed@next` with changes that enables proper tree-shaking support.

It should not be necessary to:
```js
import {calculate} from 'scroll-into-view-if-needed/dist/calculate'
```

For the tree shaking to work properly. I haven't enabled the package.json trick you mentioned yet as I wanted to test this first. It should work well with both jest, rollup and webpack without nasty workarounds like running babel on node_modules and similar 😄 

This is tested to tree shake properly on rollup@0.50.0:

```js
import {calculate} from 'scroll-into-view-if-needed'

console.log(calculate)
```

Also: it currently weighs around 4kB but there's still a few things that can be done to cut down the size (like avoiding `{...spread}` syntax so that it's not necessary to generate es5 helpers for it). I will keep publishing on `next` until we're happy with it.

kentcdodds:
Ok, so I've updated things with the new version and the tests run! The build error is still a problem though and the `utils.scroll-into-view.js` tests fail which leads me to believe that the behavior would change based on this change which makes me worried. I think at this point I'm going to stop working on this. If anyone would like to pick up where I left off in another PR feel free, but I'm going to go ahead and close this PR and the associated issue.

Thanks!

stipsan:
Sure no worries. It could mean that there's a bug in `scroll-into-view-if-needed` so I'll investigate it.
I can take over from here and open a new PR after I've made sure your testing suite pass. 

Thanks for your help, this will improve `scroll-into-view-if-needed` for all our users no matter if the end game is you guys ending up using it or not 🙂

kentcdodds:
Thanks @stipsan!

PR #261, opened 2017-11-22
feat(render): add render prop

### LINKED ISSUE #258: Support a render prop
I'll provide more info as soon as I'm able, but in the next breaking change (whenever that happens, no plans any time soon), I want to switch from the `children` prop to a new prop called `render`. I have reasons for this (one big one being teachability, also the `children` prop is special in react so it kinda doesn't make sense to use it).

In any case, we can allow folks to opt-into this with the new `breakingChanges` prop that was just added (see the docs). So if you opt-in, then we'll call the `render` function, otherwise we'll call the `children` function like we do today.

kentcdodds:
**What**: This also moves all examples and tests to use render prop and adds one test to verify the children prop still works

**Why**: Closes #258 

I'm trying to bring some cohesion to the render prop idea.

**How**: the source code change was as simple as `this.props.render ||`. Most of the changes are updating the stories, tests, and docs to use the render prop instead.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

NOTE: The children prop still works fine. I don't know whether I'll ever drop support for it. Can't think of a reason why I would...

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,24 +7,22 @@
 <p align="center" style="font-size: 1.2rem;">Primitives to build simple, flexible, WAI-ARIA compliant React
 autocomplete/dropdown/select/combobox components</p>
 
-> See [the intro blog post](https://blog.kentcdodds.com/introducing-downshift-for-react-b1de3fca0817)
+> See
+> [the intro blog post](https://blog.kentcdodds.com/introducing-downshift-for-react-b1de3fca0817)
 
 <hr />
 
 [![Build Status][build-badge]][build]
 [![Code Coverage][coverage-badge]][coverage]
-[![downloads][downloads-badge]][npmcharts]
-[![version][version-badge]][package]
-[![MIT License][license-badge]][LICENSE]
+[![downloads][downloads-badge]][npmcharts] [![version][version-badge]][package]
+[![MIT License][license-badge]][license]
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-45-orange.svg?style=flat-square)](#contributors)
-[![PRs Welcome][prs-badge]][prs]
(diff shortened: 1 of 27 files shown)


README.md
```diff
@@ -171,8 +171,8 @@ function App() {
 ```
 
 `downshift` is the only component. It doesn't render anything itself, it just
-calls the child function and renders that. Wrap everything in
-`<Downshift>{/* your function here! */}</Downshift>`.
+calls the child function and renders that. Wrap everything in `<Downshift>{/*
```
petetnt: This description should be updated too:

> `Downshift` is the only component. It doesn't render anything itself, it just calls the function passed to the `render` prop and renders that.  Wrap everything in `<Downshift render={/* your function here! /*} />`

README.md
```diff
@@ -245,147 +246,160 @@ properties:
...
-Called whenever the input value changes. Useful to use instead or in combination of `onStateChange` when `inputValue` is a controlled prop to [avoid issues with cursor positions](https://github.com/paypal/downshift/issues/217).
+Called whenever the input value changes. Useful to use instead or in combination
+of `onStateChange` when `inputValue` is a controlled prop to
+[avoid issues with cursor positions](https://github.com/paypal/downshift/issues/217).
 
-- `inputValue`: The current value of the input
-- `stateAndHelpers`: This is the same thing your `children` prop
-  function is called with (see [Child Callback Function](#child-callback-function))
+* `inputValue`: The current value of the input
+* `stateAndHelpers`: This is the same thing your `children` prop function is
```
petetnt: > your `render` prop function is called with...

kentcdodds:
Thanks for the review @petetnt! You've been added to the contributors table :tada:

PR #264, opened 2017-11-27
Add warning when user use Object for selectedItem

philipyoungg:
**What**: Add `console.warn` when user passed Object to `selectedItem`

**Why**: To improve DX—to minimize a case when user get `[object Object]` as an input value. Almost refactored my whole codebase. I thought I can't use `Object` as a selected value. Thankfully I found `itemToString` API that accept a function with current `selectedItem` passed on the argument.

**How**: Two things: 
1. Adding simple `isPlainObject` function that check if the `selectedItem` is Object literal, and
2. use that function inside `itemToString`'s defaultProps

**Checklist**:

- [N/A] Documentation
- [x] Tests
- [x] Ready to be merged 


src/downshift.js
```diff
@@ -67,7 +68,12 @@ class Downshift extends Component {
     defaultIsOpen: false,
     getA11yStatusMessage,
     id: generateId('downshift'),
-    itemToString: i => (i == null ? '' : String(i)),
+    itemToString: i => {
+      if (isPlainObject(i)) {
```
kentcdodds: I'd prefer this to only happen in dev. I think the build is wired up to handle `process.env.NODE_ENV` properly. Could you make this: `if (process.env.NODE_ENV !== 'production' && isPlainObject(i)) {`

src/downshift.js
```diff
@@ -67,7 +68,12 @@ class Downshift extends Component {
...
+        console.warn('you passed an Object as the value of `selectedItem`. Please refer to `itemToString` API documentation on Downshift\'s repository.')
```
kentcdodds: I think prettier should have reformatted this automatically when you committed. Did you `npm install`?

src/downshift.js
```diff
@@ -67,7 +68,12 @@ class Downshift extends Component {
     defaultIsOpen: false,
     getA11yStatusMessage,
     id: generateId('downshift'),
-    itemToString: i => (i == null ? '' : String(i)),
+    itemToString: i => {
```
kentcdodds: I'd like to have a test for this. Could you add one in [`downshift.misc.js`](https://github.com/paypal/downshift/blob/master/src/__tests__/downshift.misc.js)? Just do `jest.spyOn(console, 'warn')` on the first line of the test, then `console.warn.mockRestore()` on the last line to clean up. You could even use a snapshot if you want: `expect(console.warn).toHaveBeenCalledTimes(1)` and then `expect(console.warn.mock.calls).toMatchSnapshot()`

philipyoungg: Sure!

src/downshift.js
philipyoungg: Would `process.env.NODE_ENV === 'development'` more prefereable in this instance?

PR #265, opened 2017-11-27
[WIP] Add support for React Native as a target

### LINKED ISSUE #185: Implement support for React Native

eliperkins:
**What**: Adds support for using 🏎 in React Native. Closes #185.

**Why**: Prior to this, React Native could not be targeted, as there are DOM-specific codepaths. This ensures those codepaths are not followed on React Native.

**How**: Added the basic `downshift` implementation into a React Native app and started hammering out errors, stripping away DOM-related code.

**Checklist**:

None of these are done yet! This is a work-in-progress. Putting this up for review as a check to make sure I'm headed in the right direction.

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 


src/downshift.js
```diff
@@ -621,7 +630,7 @@ class Downshift extends Component {
     this.internalSetState({
       type: Downshift.stateChangeTypes.changeInput,
       isOpen: true,
-      inputValue: event.target.value,
+      inputValue: isReactNative() ? event : event.target.value,
```
eliperkins: This seems a bit odd, since the `onChangeText` prop on React Native returns just the text as a string, not an object. Is there a better way to go about this?

src/downshift.js
```diff
@@ -723,7 +734,10 @@ class Downshift extends Component {
       ...state,
     })
     this.previousResultCount = resultCount
-    setA11yStatus(status)
+    // TODO: make this work.
+    if (!isReactNative()) {
+      setA11yStatus(status)
```
eliperkins: It looks like `setA11yStatus` calls a bunch of DOM-specific APIs, which don't exist on React Native. Should we fork the logic and send this a11y status elsewhere?

src/utils.js
```diff
@@ -224,6 +224,17 @@ function isDOMElement(element) {
...
 
+/**
+ * React Native sets `navigator.product` to a constant `ReactNative`.
+ * We can use this to determine if the target is React Native or not.
+ * See https://github.com/facebook/react-native/issues/10881 for more info.
+ * See https://github.com/facebook/react-native/blob/70c359000a2df091c3939f4c19db6024af992d43/Libraries/Core/InitializeCore.js#L194-L195 for more info.
+ * @return {Boolean} whether or not the platform is React native
+ */
+function isReactNative() {
+  return navigator != null && navigator.product === 'ReactNative'
```
eliperkins: Can this be `preval`ed?

eliperkins: I think the answer is no, since this needs to exist at runtime. Maybe there's a better alternative to use here that could help us cut down on some function calls.

eliperkins:
Alright, spent some time playing around with this today, and it looks like there's some corners to be cut to get this to work on React Native.

There's a few functions that are DOM-specific, like `scrollIntoView` and `setA11yStatus`, so this mainly cuts off calls to those functions and remaps event handlers like `onClick` to React Native's `onPress` and most things Just Work™! 🎊

With regards to `scrollIntoView` being DOM-specific, there are methods like [`FlatList.scrollToItem`](http://facebook.github.io/react-native/releases/0.49/docs/flatlist.html#scrolltoindex), but this would force consumers to use a `FlatList` component, which I don't think is the greatest. Users can implement this themselves if they choose to use a `FlatList`, I think.

@gricard Had mentioned in https://github.com/paypal/downshift/issues/185#issuecomment-329017757 that there are a few accessibility props that we could pass to input, button, and item views, but I omitted those for now, since they're `accessibilityTraits` and React isn't too keen on us passing those camel-case properties to DOM elements. We could switch these out with the `aria-` properties since those have no effect on React Native. If this is something we want, let me know!

I'd love to find a cleaner way to get this in without littering `isReactNative()` calls everywhere, too! Feedback is more than welcome.

Additionally, should I add in a React Native test suite as well? I believe this would just involve adding `react-native` and related Jest test utils to the test suite, and then writing a few components for this ([see also: these Jest docs](https://facebook.github.io/jest/docs/en/tutorial-react-native.html)).

eliperkins:
Oh! Additionally, there's a sample project here: https://github.com/eliperkins/DownshiftNative

src/downshift.js
```diff
@@ -181,9 +189,12 @@ class Downshift extends Component {
   ) => {
     otherStateToSet = pickState(otherStateToSet)
     this.internalSetState({highlightedIndex, ...otherStateToSet}, () => {
-      const node = this.getItemNodeFromIndex(this.getState().highlightedIndex)
-      const rootNode = this._rootNode
-      scrollIntoView(node, rootNode)
+      // TODO: this whole method is DOM specific...
```
eliperkins: Should we just omit this logic here? This is specific to DOM and to get this to work with React Native, we'd have to force consumers to use a `ScrollView` and pass us the ref.

eliperkins:
One last question, should we omit a call to `validateGetRootPropsCalledCorrectly` on React Native or should we force consumers to create a component with an "innerRef" and pass it to us? I don't quite understand `validateGetRootPropsCalledCorrectly` and `isDOMElement` well enough to have fork logic in `validateGetRootPropsCalledCorrectly` to get it to be quiet when passing a `<View>` instead of a `<div>`.

gricard:
Nice! Thanks for picking up the torch there. I got wicked sidetracked. 👍

src/downshift.js
```diff
@@ -172,6 +173,13 @@ class Downshift extends Component {
   }
 
   getItemNodeFromIndex = index => {
+    if (isReactNative()) {
+      throw new Error(
```
kentcdodds: We don't need to worry about throwing an error because `getItemNodeFromIndex` is not publicly exposed so this should never be called :)

src/downshift.js
kentcdodds: Yeah, we'll probably want some different logic here for React Native

kentcdodds:
Oh, also, THANK YOU SO MUCH FOR WORKING ON THIS! :tada:

eliperkins:
> Specifically, that we have a separate build version of the library for the react native version

@kentcdodds sounds good! I think that approach makes sense. It looks like the Preact build pipeline is pretty embedded into `kcd-scripts` and the Rollup config. Would you like it if I worked the React Native build pipeline into `kcd-scripts` or just keep it local to this project?

src/downshift.js
eliperkins: Heh, good call. This was part of my debugging earlier to figure out what codepath was being followed to call this. I'll remove it!

kentcdodds:
Let's keep it separate. No need to abstract it away until we know what it looks like 😉

I do expect it to be pretty similar though. Some environment variable we use with preval. We could probably make a rollup.config.rn.js file for this

eliperkins:
I took a quick pass at this, it seems like `BUILD_FILENAME_PREFIX` came in pretty handy here to spit React Native-specific builds out to a different place.

I've added a new build script (that still needs to get hooked into the normal `build` script) called `build:react-native` which adds a new env var and output directory. This means we can preval our `isReactNative` logic here.

Unfortunately, it looks like our calls to `Boolean(preval`true`)` don't get minified away and we need the `Boolean` calls based on this comment: https://github.com/paypal/downshift/blob/4dc8bf2f4d88a03d2ec485ae36a2490976fe886e/src/downshift.js#L596-L597
so we don't gain all that much in terms of limiting codepaths for non-React Native targets.

I hope I understood your comment about creating a new target here, and took the right approach! Feedback is more than welcome. 😄

eliperkins:
Oh man, I didn't realize the `package.json` work that needs to get done too. Looking into that now.

kentcdodds:
> so we don't gain all that much in terms of limiting codepaths for non-React Native targets.

Yeah... I think that [the babel plugin we're using for dead code elimination](https://github.com/babel/minify/tree/master/packages/babel-plugin-minify-dead-code-elimination) doesn't handle `Boolean(false)` very well. I think the core issue is in babel's `path.evaluate` function. Shouldn't be too terribly difficult to solve, and would be really handy to fix eventually.

Judging from [this](https://skalman.github.io/UglifyJS-online/) UglifyJS doesn't actually handle this very well either. So perhaps a better solution would be to solve the issues with integrating `preval` with `babel-plugin-istanbul` then we wouldn't need the Boolean cast at all. I think that may be the best solution...

kentcdodds:
Thanks so much for all your work on this. It's awesome!

package.json
```diff
@@ -9,6 +9,7 @@
   "scripts": {
     "add-contributor": "kcd-scripts contributors add",
     "build": "kcd-scripts build --bundle --p-react",
+    "build:react-native": "BUILD_REACT_NATIVE=true BUILD_FILENAME_PREFIX=react-native kcd-scripts build --bundle",
```
eliperkins: I'm not sure the best way to combine this build step with the normal `build` without clobbering some of the niceties of `kcd-scripts build` (by using `&&` to run both commands).

src/utils.js
```diff
@@ -224,6 +226,17 @@ function isDOMElement(element) {
   }
 }
 
+/**
+ * React Native sets `navigator.product` to a constant `ReactNative`.
```
eliperkins: This comment is no longer valid and should be removed.

eliperkins:
Alright, I kinda tackled this with as minimal destruction as I could.

I ended committing a `package.json` directly to the project, since [there's a decent amount of logic baked into `kcd-scripts` to do this for Preact](https://github.com/kentcdodds/kcd-scripts/blob/e9b1d5fa1a774e9d2f0d200ff8ffa4565edc69df/src/scripts/build/rollup.js#L68-L85).

Additionally, I appeased `babel-plugin-istanbul` and `preval` by telling Istanbul to ignore the `preval`'ed statements. Not a great long-term solution, but I think it should suit our needs, meaning we get to drop the `Boolean` cast and get better minification! The downside of this means littering an handful of `/* istanbul ignore */`s in those codepaths.

I think the biggest open question I have is this comment: https://github.com/paypal/downshift/pull/265#discussion_r154395148. I'm not sure the best way to include this in the current build script without extending `kcd-scripts` to know to invoke more commands. I'd love some feedback related to this.

eliperkins:
Hey @kentcdodds! Happy New Year! 🎊 

Any chance you could take another peek at this at some point in the next few days? I'd love a little more advice regarding two things:

1. What are the best next steps I could take to get this merged in?
1. What is the best way to tackle this sucker? https://github.com/paypal/downshift/pull/265#discussion_r154395148

Cheers!

package.json
kentcdodds: I'd just have `build:web` and a `build:native` scripts and then a `build` script that just runs them both: `npm run build:web --silent && npm run build:native --silent` 👌

src/utils.js
```diff
@@ -224,6 +226,20 @@ function isDOMElement(element) {
...
+ * concepts because of the React Native target.
+ * @return {Boolean} whether or not the platform is React Native
+ */
+/* istanbul ignore next (react-native) */
+const isReactNative = preval`module.exports = process.env.BUILD_REACT_NATIVE === 'true'`
+
+// Disabling coverage here is necessary due to a weird deal with
+// babel-plugin-istanbul + preval.macro. No idea...
+/* istanbul ignore next (preact) */
+const isPreact = preval`module.exports = process.env.BUILD_PREACT === 'true'`
```
kentcdodds: I think that these will actually need to be inlined as they were before to make sure the code is removed properly. I realize it's not as clean, but it'll work better.

src/downshift.js
```diff
@@ -661,7 +674,9 @@ class Downshift extends Component {
     this.internalSetState({
       type: Downshift.stateChangeTypes.changeInput,
       isOpen: true,
-      inputValue: event.target.value,
+      inputValue: isReactNative
+        ? /* istanbul ignore next (react-native) */ event
```
kentcdodds: Interesting! In react native the `event` is considered the `inputValue`? That's a surprise... Is it really a string? The `inputValue` should be a string that's the value of the input. Other parts of the code make that assumption. Just want to clarify that in react native `event` is a string that's the value of the input. If not, then we need to get the actual string value.

src/utils.js
Andarist: This PR got me interested and I agree - better to have those branches DCE-ed.

kentcdodds:
Thanks! I prefer option 1 👍

Andarist:
It would probably be good to add support for this in your `kcd-scripts`, like some warning when you do not have `react-native` key in your root `package.json`, but you are building for react native target?

kentcdodds:
Maybe so... But let's get it working here first 😉

eliperkins:
Ah! Thanks so much for the feedback @Andarist!!!

I apologize I haven't gotten around to wrapping this up yet, it's been super busy for me with work and life. I _should_ be able to spend an hour or two on this tomorrow though! ❤️

Andarist:
> Maybe so... But let's get it working here first 😉

Sure, sure - just throwing ideas out loud so they can be picked up later 😉 

> Thanks! I prefer option 1 👍

I agree it's nice to have a clear indication, like this additional key in package.json, that a package supports `react-native` by providing other files for it.

I'd consider going extension-less for those fields anyway though, as providing a good example for others. Extension-less entries let tools to choose on their own appropriate extension, which is useful because tools might prefer one extensions over others.

This is ofc not a problem when you do not provide various extensions, but I'd say it's still to go this way as "better" practice. In example when ESM modules will get node.js support through `.mjs` extension and if you will want to support both cjs and esm for node users, you will have to go extension-less with your "main", and provide 2 files named exactly the same under that location - 1 with '.js' and one with '.mjs' extensions.

The same way Metro Bundler if pointed to extension-less file it can try to select first .ios.js (or android.ios.js, depending on the platform), then .native.js and finally .js. Which again - is not a problem if you have a single file targeting react-native. Just commenting differences between chosen solutions. 😅

kentcdodds:
That's a fair point (though I'm growing more confident that `.mjs` will not be a real thing in the future). Either way, I'm fine with removing the extension from the `package.json` :)

nhducit:
I started using downshift yesterday. I love it so much. I'm working on both React Web/Native project. It's great if I can share downshift logic between web/native. Are there anything I can help to make this PR merged?

src/downshift.js
eliperkins: Yup! This specific callback, [`onChangeText`](https://facebook.github.io/react-native/docs/textinput.html#onchangetext), takes a function that is called with the text as a string, that's it.

Other callbacks on `TextInput` that are more generic for the input itself (like `onBlur` or `onSubmitEditing`) return an event as a `Proxy` in the shape of `{ nativeEvent: { text: string } }`, however, this specific callback just gives us a `string`.

src/utils.js
kentcdodds: Thanks! I'm less concerned about the `umd` builds and more concerned about all the other ones. Could you run the builds and see if Rollup was able to eliminate those? If not, could you try to inline the preval usage and see if Rollup can eliminate it that way. I know it's not pretty, but we've gotta work with the tools we have.

eliperkins:
Alright, I addressed some of the feedback including moving to the `"react-native"` key in the `package.json` and addressing some of the feedback around the build process.

My last question seems to be around the `isPreact` and `isReactNative` bits being `preval`-ed and minified away.

I have a big launch week this week at work, so if anyone wants to pick this up and take it across the finish line, I'm open to others taking this PR over! I've checked off "Allow edits from maintainers" but if there are other tweaks, we can PR into this branch or something too!

kentcdodds:
Do you mind rebasing this branch onto `master` to resolve the conflicts?

src/utils.js
eliperkins: Ah, gotcha. Rollup does not eliminate them there. I'll add the `preval`s back in, inline!

Do you have more reading for me to learn more about why someone might not use the `umd.min` version of the build so I can understand the benefit of optimizing those builds?

Andarist:
I can finish this up. I'm not a maintainer of this package though, so you'd have to invite me to your fork I guess.

src/downshift.js
```diff
@@ -218,9 +218,12 @@ class Downshift extends Component {
   }
 
   scrollHighlightedItemIntoView = () => {
-    const node = this.getItemNodeFromIndex(this.getState().highlightedIndex)
-    const rootNode = this._rootNode
-    scrollIntoView(node, rootNode)
+    /* istanbul ignore else (react-native) */
+    if (preval`module.exports = process.env.BUILD_REACT_NATIVE !== 'true'`) {
```
Andarist: i dont think `preval` is actually needed here, rollup should handle this just fine with `rollup-plugin-replace`

kentcdodds:
@Andarist, I've added you as a collaborator to downshift 👍 Thanks for all your help so far. Here's the normal thing I show peopel when I add them:

Andarist:
Cool, it doesnt seem there is much to be done. I will wrap this up tomorrow (its midnight here already) and ping you for a review.

kentcdodds:
Super duper!

Andarist:
@kentcdodds pretty much done, I've tweaked the build setup and it was the only thing I was looking at - I didn't actually test this on RN (functionality-wise).

Things I've noticed while doing this:
- would be cool to support building for `react-native` as core feature of `kcd-scripts` - now I had to figure my way around standard setup (i.e. had to use `--no-clean` flag etc)
- would be great to support `rollup-plugin-replace`in kcd-scripts, that way those lines with preval could get more readable
- i have no idea what [`root_handleClick`](https://github.com/paypal/downshift/blob/118a87234a9331e716142acfb95eb411cc4f8015/src/downshift.js#L135-L138) is, i dont see it anywhere in the project and it caches `this.props.onClick` which can lead to stale values being used

package.json
```diff
@@ -4,14 +4,14 @@
...
-  "react-native": "native/index.js",
+  "react-native": "dist/downshift.native.cjs.js",
   "module": "dist/downshift.esm.js",
   "typings": "typings/index.d.ts",
   "scripts": {
     "add-contributor": "kcd-scripts contributors add",
     "build": "npm run build:web --silent && npm run build:native --silent",
     "build:web": "kcd-scripts build --bundle --p-react",
-    "build:native": "BUILD_REACT_NATIVE=true BUILD_FILENAME_PREFIX=react-native kcd-scripts build --out-dir native",
+    "build:native": "BUILD_REACT_NATIVE=true BUILD_FILENAME_SUFFIX=.native kcd-scripts build --bundle cjs --no-clean",
```
eliperkins: What's the rationale around including the `--no-clean` flag here?

package.json
```diff
@@ -29,8 +29,7 @@
   "files": [
     "dist",
     "typings",
-    "preact",
-    "react-native"
```
eliperkins: Thanks for tidying this up for me!

package.json
Andarist: `build:web` are and `build:native` are ran sequentially, `kcd-scripts build --bundle` removes the target directory, so its removed during `build:web` and I dont want to remove what was just built by it by running `build:native`

the goal for me was to not introduce a separate directory and Ive wanted to output react-native version into `dist`

eliperkins: What do you think of adding `--no-clean` only to the `build` command, rather than `build:native`? That way, the `build:native` command can be run safely, clearing out the `dist` directory, while the `build` command would be able to handle the sequential case like you mentioned?

Andarist: that makes sense

Andarist: @eliperkins changed

Andarist:
@eliperkins 
I've missed your one comment, because GitHub has hid it because some files changed there after you have commented there. I prefer to have a sideways conversation here, so there is bigger chance that it won't be missed later.

By DCE-ed I've meant dead code eliminated. It's actually UglifyJS which does it and not rollup, let's look on a sample code after its gets built with rollup:
```js
```

package.json
```diff
@@ -4,11 +4,14 @@
...
   "jsnext:main": "dist/downshift.esm.js",
+  "react-native": "dist/downshift.native.cjs.js",
   "module": "dist/downshift.esm.js",
   "typings": "typings/index.d.ts",
   "scripts": {
     "add-contributor": "kcd-scripts contributors add",
-    "build": "kcd-scripts build --bundle --p-react",
+    "build": "npm run build:web --silent && npm run build:native --silent -- --no-clean",
+    "build:web": "kcd-scripts build --bundle --p-react",
+    "build:native": "BUILD_REACT_NATIVE=true BUILD_FILENAME_SUFFIX=.native kcd-scripts build --bundle cjs",
```
kentcdodds: Let's add a devDependency on `cross-env` and use that here so this works on windows machines as well (considering the `setup` script will run `validate` which will run the build. I want to make sure that script always works on windows.

src/downshift.js
kentcdodds: I like that idea. Let's add a feature to `kcd-scripts` and then upgrade this to use that feature.

I think we'll want to configure that somehow though (CLI args?) because there are some `process.env.*` variables that we want to leave as-is.

kentcdodds:
> i have no idea what root_handleClick is, i dont see it anywhere in the project and it caches this.props.onClick which can lead to stale values being used

I think that's probably a left-over mistake or something. Feel free to remove it as part of this PR! Thanks!

package.json
Andarist: done

Andarist:
> I like that idea. Let's add a feature to kcd-scripts and then upgrade this to use that feature.

I can do that, but this shouldnt hold off this PR

> I think we'll want to configure that somehow though (CLI args?) because there are some process.env.* variables that we want to leave as-is.

Not sure what variables you'd like to leave as is, I'd just take all `process.env.*` and put it through `rollup-plugin-replace`. One must actually use env var in a code for it to have any effect, so danger of using some additional env var that was not supposed to get through is actually pretty low imho.

> I think that's probably a left-over mistake or something. Feel free to remove it as part of this PR! Thanks!

gonna send separate PR for this

kentcdodds:
The thing is we want to use `process.env.NODE_ENV` for some things: https://github.com/paypal/downshift/blob/531bd8ea83ae5ecc3c163357ac9be4dc721d2060/src/downshift.js#L78

I imagine we'll want to use it in the future as well. Currently the UMD bundle replaces that (`development` for umd and `production` for umd.min), but for cjs and esm we want to leave those as-is. But honestly those are probably the only ones we'd care to leave as-is so we can probably make that work :+1:

I agree that should not hold up this PR.

Is this ready to merge then?

kentcdodds:
I think when we merge this we'll merge this with a note that it's experimental and could break. Would be good to get feedback on it before it's official.

Andarist:
> I imagine we'll want to use it in the future as well. Currently the UMD bundle replaces that (development for umd and production for umd.min), but for cjs and esm we want to leave those as-is. But honestly those are probably the only ones we'd care to leave as-is so we can probably make that work 👍

damn, I totally forgot about `process.env.NODE_ENV` checks, you are right ofc - it's probably the only cause we'd like to omit for cjs & es bundles

> Is this ready to merge then?

from my POV - yes

> I think when we merge this we'll merge this with a note that it's experimental and could break. Would be good to get feedback on it before it's official.

Sure thing.

PR #270, opened 2017-11-30
feat(itemCount): add `setItemCount` and `unsetItemCount`

kentcdodds:
**What**: add `setItemCount` and `unsetItemCount` 

**Why**: This enables developers who need to both asynchronously load options within downshift and window the results. The `itemCount` needs to be set from within downshift so they cannot reasonably use the `prop`.

**How**: Adding a member property to keep track of the set value and a getter (and unsetter as convenience) for that member property.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged
      

kentcdodds:
I'm not sure why the netlify deploy is failing, but I'm confident this is harmless :)

PR #279, opened 2017-12-08
fixes scroll into view cases above viewport

### LINKED ISSUE #274: Keying down from last item to first item doesn't scroll all the way
* `downshift` version: v1.22.1
* `node` version: v8.9.0
* `npm` (or `yarn`) version: v5.5.1

Relevant code or config is a bunch of custom internal junk, but I reproduced it at the [example](https://codesandbox.io/s/9rWEmzEv8).

What you did:
1. Load up the example
2. Click on the dropdown arrow
3. Press the down arrow key
4. Press the up arrow key
5. Press the down arrow key

What happened:
Where did the highlight go? (hint the math calculating what to scroll to didn't scroll all the way to the top)
image

Reproduction repository:

Internal stuff bleh, but I used your example, hopefully that works.

Problem description:

The math calculating where to scroll to doesn't account for the parent's position properly when the node is above the page viewport.

Suggested solution:

This code seems to work for me, but I haven't fully tested it yet

```JavaScript
if (nodeRect.top < 0) {
    // the item above view
    scrollParent.scrollTop += nodeRect.top - scrollParentTop;
    return;
  }
```

Zashy:
**What**: Fixes #274

**Why**: See #274. Scrolling was weird when going from last to first item with arrow keys.

**How**: I added a change I used in my project, and then worked on figuring out the unit tests. On reviewing the unit tests I realized the intent of the previous code was meant for multiple cases when the item is above the current viewport. I worked out that there are three cases. They are hard to explain, so I made an image to better explain.

image

* The scrollable parent is above the viewport, and the item is above that parent
* The scrollable parent is above the viewport, and the item is above the viewport, but below the parent
* The scrollable parent is within the viewport, but the item is above the viewport

The first two simplified into the same equation after working them out, and was the same as the existing logic. The third case was the one I had fixed already. I then added in the conditional logic to select the cases, and applied the formulas. I also updated one unit test, and added two new unit tests to account for all three cases I found.

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged
      
### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ autocomplete/dropdown/select/combobox components</p>
 [![downloads][downloads-badge]][npmcharts] [![version][version-badge]][package]
 [![MIT License][license-badge]][license]
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-50-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-51-orange.svg?style=flat-square)](#contributors)
 [![PRs Welcome][prs-badge]][prs] [![Chat][chat-badge]][chat]
 [![Code of Conduct][coc-badge]][coc]
 
@@ -75,7 +75,7 @@ harder to contribute to.
   * [highlightedIndex](#highlightedindex)
   * [inputValue](#inputvalue)
   * [isOpen](#isopen)
-  * [`selectedItem`](#selecteditem)
+  * [selectedItem](#selecteditem)
   * [render](#render)
   * [id](#id)
   * [environment](#environment)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -279,8 +278,7 @@ This function is called anytime the internal state changes. This can be useful
 if you're using downshift as a "controlled" component, where you manage some or
 all of the state (e.g. isOpen, selectedItem, highlightedIndex, etc) and then
 pass it as props, rather than letting downshift control all its state itself.
-The parameters both take the shape of internal state (`{highlightedIndex:
-number, inputValue: string, isOpen: boolean, selectedItem: any}`) but differ
+The parameters both take the shape of internal state (`{highlightedIndex: number, inputValue: string, isOpen: boolean, selectedItem: any}`) but differ
```
kentcdodds: I'm guessing this reformatting happened because you either didn't install the project dependencies or the recent prettier release has some bugs 🤔

Zashy:
will do

README.md
Zashy: I tried to follow the instructions as best as I could. Not sure where that was introduced, but somewhere between install and doing the contribution script.

kentcdodds: No worries :smile:

kentcdodds:

PR #282, opened 2017-12-13
Consolidate scrolling of the list when highlighted index changes and never scroll from a `onMouseEnter` event (fixes #281)

### LINKED ISSUE #240: Keyboard selection is rudely interrupted and overridden by auto-complete mouse-enter actions.
- `downshift` version: 1.16.0
- `node` version: 8.5.0
- `npm` (or `yarn`) version: yarn 1.0.2

### Description
In any downshift configuration (at least thus far in my experience), the mouse position will fire (and select whatever is underneath it) immediately after an arrow-selecting and shift-arrow-paging event.

### Replication
1. Navigate to any example: http://downshift.netlify.com/?selectedKind=Examples&selectedStory=basic&full=0&down=1&left=1&panelRight=0
2. Open the menu by changing the input
3. Make sure there are multiple "pages" of options.
4. Place your cursor over the first option in the list
5. Using the down-arrow on your keyboard, navigate down the list
6. As soon as your arrow-down action selects an option beyond the visible set (either by arrowing-down on the last option, or shift-arrowing-down to change the whole page) the option underneath the cursor position will subsequently be selected.

### Expected
I would expect the cursor to not activate an option while using arrow navigation, unless of course, the cursor is moved.

### LINKED ISSUE #281: Scrolling is janky when controlling the `highlightedIndex` prop
I originally filed #280 even though it wasn't the exact issue I was hitting, but was related. I was able to reduce my issue down to a test case.

* `downshift` version: 1.22.3
* `node` version: 7.9.0
* `npm` (or `yarn`) version: 4.2.0

Reproduction repository: https://codesandbox.io/s/x3y2xzppyw

Problem description:

When you control the highlightedIndex prop, in `componentDidUpdate` it checks if the index has changed and if so scrolls that item into view:

```js
    if (
      this.isControlledProp('highlightedIndex') &&
      this.props.highlightedIndex !== prevProps.highlightedIndex
    ) {
      this.scrollHighlightedItemIntoView()
    }
```

The problem is that the index can change just simply by scrolling the list. After you scroll the list, Chrome apparently fires a `onMouseEnter` event if the cursor is sitting over an item, even if the browser has hidden the cursor because and you're moving it with the keyboard or something. This is related to #240, but not exactly the same. In that issue, that event firing causes issues when navigating with the keyboard.

This also causes problems when scrolling with the trackpad though, and is probably hit much more frequently. You have to hover the cursor over the list to scroll it, so the cursor will always be hovering over an item.

In that case, as you scroll the list with the trackpad, if you've scrolled far enough, Downshift will try to scroll the item into view and mess up your scrolling. There seems to be an issue with the order of events: after you've scrolled, if the previous highlighted item has been scrolled out of view, Downshift thinks it needs to scroll the list, even though it shouldn't because the *new* highlighted item that the cursor sits over it obviously in view.

Here's a video: http://jlongster.com/s/downshift-scroll.mov

Downshift should never touch scrolling if you're scrolling with the trackpad.

Even if we fix this bug and downshift properly sees that the new highlighted item is in view and doesn't touch scrolling, since downshift scrolls *partially* hidden items to be fully shown, there's still a chance that my cursor will sit over the item at the bottom and cause similar problems. The fix should be to detect if the user is scrolling and never call `scrollIntoView`.

Suggested solution:

Detect if the user is scrolling and never call `scrollIntoView` if they are. The difficulty is how to detect this, since we don't know anything about the scrolled element. `scrollIntoView` used `getClosestScrollParent` to find it, but we need to set a scroll event handler on it at mount time and don't have a child node to walk up from.

jlongster:
Fixes #281. 

As I dug into this I realized that it would be nicer to consolidate the code for scrolling to an item so it behaves exactly the same whether or not you are controlling the prop. This will reduce bugs between those use cases.

The way to do this is to always scroll within `componentDidUpdate`. The only problem is that this makes it harder to control scrolling; my first attempt added an option to `setHighlightedIndex` to avoid scrolling, and since that's called directly from `onMouseEvent` we can pass that information. But in `componentDidUpdate` we have no idea why it's changing. However this is always the case in the controlled prop scenario anyway, so we have to accept it.

The only solution is an ugly setTimeout which tells the system to not scroll it at all. This nicely fixes the problem entirely; it'll never scroll it if you are moving the cursor around or scrolling yourself and makes it feel a lot better. Note that this also changes the uncontrolled prop scenario: if you hover over an element at the bottom of the list, **it will no longer scroll down** to show it fully in view. This is better behavior in my opinion though as I'd never expect it to move around under me just by hovering.

This does not fix #240, as it's a different issue caused by a similar `onMouseEnter` event. A similar fix could be adopted there where it sets a timeout to tell the system to avoid any `onMouseEnter` events after key up/down.

**Checklist**:

* [ ] Documentation NA
* [x] Tests
* [x] Ready to be merged
      
### DIFF
--- a/src/__tests__/downshift.misc-with-utils-mocked.js
+++ b/src/__tests__/downshift.misc-with-utils-mocked.js
@@ -1,0 +1,52 @@
+// this is stuff that I couldn't think fit anywhere else
+// but we still want to have tested.
+
+import React from 'react'
+import {mount} from 'enzyme'
+import Downshift from '../'
+import {scrollIntoView} from '../utils'
+
+jest.useFakeTimers()
+jest.mock('../utils')
+
+test('does not scroll from an onMouseEnter event', () => {
+  class HighlightedIndexController extends React.Component {
+    state = {highlightedIndex: 10}
+    handleStateChange = changes => {
+      if (changes.hasOwnProperty('highlightedIndex')) {
+        this.setState({highlightedIndex: changes.highlightedIndex})
+      }
+    }
+    render() {
+      return (
+        <Downshift
+          onStateChange={this.handleStateChange}
+          highlightedIndex={this.state.highlightedIndex}
+          render={({getInputProps, getItemProps}) => (
+            <div>
+              <input {...getInputProps()} />
+              <div {...getItemProps({item: 'hi', 'data-test': 'item-1'})} />
+              <div {...getItemProps({item: 'hey', 'data-test': 'item-2'})} />
+            </div>
+          )}
+        />
+      )
+    }
+  }
+  const wrapper = mount(<HighlightedIndexController />)
+  const input = wrapper.find('input')
+  const item = wrapper.find(sel('item-1'))
+  item.simulate('mouseenter')
+  jest.runAllTimers()
+  expect(scrollIntoView).not.toHaveBeenCalled()
+  // now let's make sure that we can still scroll items into view
+  // ↓
+  input.simulate('keydown', {key: 'ArrowDown'})
+  expect(scrollIntoView).toHaveBeenCalled()
+})
+
+function sel(id) {
+  return `[data-test="${id}"]`
+}
+
+/* eslint no-console:0 */
--- a/src/__tests__/downshift.props.js
+++ b/src/__tests__/downshift.props.js
@@ -129,7 +129,6 @@ test('uses given environment', () => {
   wrapper.unmount()
   expect(environment.addEventListener).toHaveBeenCalledTimes(2)
   expect(environment.removeEventListener).toHaveBeenCalledTimes(2)
-  expect(environment.document.getElementById).toHaveBeenCalledTimes(1)
 })
 
 test('can override onOuterClick callback to maintain isOpen state', () => {
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -213,10 +213,7 @@ class Downshift extends Component {
     otherStateToSet = {},
   ) => {
     otherStateToSet = pickState(otherStateToSet)
-    this.internalSetState(
-      {highlightedIndex, ...otherStateToSet},
-      this.scrollHighlightedItemIntoView,
-    )
+    this.internalSetState({highlightedIndex, ...otherStateToSet})
   }
 
   scrollHighlightedItemIntoView = () => {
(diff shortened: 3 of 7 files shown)


src/__tests__/downshift.props.js
```diff
@@ -129,7 +129,6 @@ test('uses given environment', () => {
   wrapper.unmount()
   expect(environment.addEventListener).toHaveBeenCalledTimes(2)
   expect(environment.removeEventListener).toHaveBeenCalledTimes(2)
-  expect(environment.document.getElementById).toHaveBeenCalledTimes(1)
```
jlongster: This function is no longer called, as its called in the `componentDidUpdate` lifecycle.

src/downshift.js
```diff
@@ -688,15 +685,13 @@ class Downshift extends Component {
...
-      ...rest
-    } = {},
-  ) => {
+  getItemProps = ({
+    onMouseEnter,
+    onClick,
+    index,
+    item = requiredProp('getItemProps', 'item'),
+    ...rest
+  } = {}) => {
```
jlongster: I don't know why this changed. It got a different prettier version somehow or something?

kentcdodds:
Ok! I've written a test for it and I'm much happier with this now that I can see the test. Makes a lot of sense. Thank you very much for the contribution :clap:

kentcdodds:
Thanks so much for your help @jlongster! I've added you as a collaborator on the project. Please make sure that you review the `other/MAINTAINING.md` and `CONTRIBUTING.md` files (specifically the bit about the commit messages and the git hooks) and familiarize yourself with the code of conduct (we're using [the contributor covenant](http://contributor-covenant.org/)). You might also want to [watch the repo](https://help.github.com/articles/watching-repositories/) to be notified when someone files an issue/PR. Please continue to make PRs as you feel the need (you can make your branches directly on the repo rather than your fork if you want). Thanks! And welcome to the team :)

src/__tests__/downshift.misc-with-utils-mocked.js
```diff
@@ -0,0 +1,52 @@
...
+          )}
+        />
+      )
+    }
+  }
+  const wrapper = mount(<HighlightedIndexController />)
+  const input = wrapper.find('input')
+  const item = wrapper.find(sel('item-1'))
+  item.simulate('mouseenter')
+  jest.runAllTimers()
```
jlongster: I learned a lot from this test. I didn't know jest had these functions. That's pretty cool!

PR #284, opened 2017-12-17
Don't Set aria-activedescendant when the menu is closed

cycomachead:
**What**:
Checks isOpen before setting the aria-activedescendant prop.

**Why**:

When you provide a `defaultHighlightedIndex`, there will always be value for highlighted index. It's an aria violation to have the active descendant be the Id of an item that is off-screen / not rendered.

(Additionally, closing the menu doesn't unset the highlighted index.)

**How**:
Simply adds a boolean check before setting the value.

**Checklist**:

* [x] Documentation (N/A)
* [x] Tests
* [x] Ready to be merged
      
* [x] Added myself to contributors table :tada:


cycomachead:
Happy to help! :)

It's always great to contribute back to projects people use, especially when it's accessible. :)

Issue #283, opened 2017-12-16
Improved Mapping of ARIA Roles and Attributes.


cycomachead:
* `downshift` version: 1.22.3
* `node` version: n/a
* `npm` (or `yarn`) version: n/a

Relevant code or config
See any example. :)

Problem description:
Currently, the ARIA roles used by downshift don't seem to map to the suggested locations. 
The ARIA Authoring Practices lists a few examples which are basically what downshift aims to solve:
https://w3c.github.io/aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

Suggested solution:

* The `combobox` role should be on the parent tag of the `<input>`. Since downshift already provides `getRootProps()`, I propose moving the role there and  updating docs to say this is a required function.
    * Secondarily, the related aria attributes for expanded and controlled should be moved here.
* The `input` should have have the role `textbox` and it should have `aria-controls` pointing to the items container. we can just update `getInputProps`
* There is a role (`listbox`) that should be applied to the element that contains the list. It would also need an `id` (for the aria-controls attribute above), and in the example provided, it should be `aria-labelledby` the main label. 
    * All this suggests we need a function `getItemListProps()` or something? 
* Each item needs a role of `option` and should have a state for `aria-selected` (which would be when an item is highlighted).

Additionally, I think we could make some improvements to the examples to use `ul` and `li` as the Authoring Practices suggest, which is a bit more semantically accurate. 

I'd also propose adding a section which illustrates the resulting DOM and prop getters mappings:

```jsx

<div {...getRootProps()}>
  <label {...getLabelProps()}>...</label>
  <input {...getInputProps()} />
  <ul {...getItemListProps()}>
    <li {...getItemProps()}>...</li>
  </ul>
</div>
```

phew...That was a lot, sorry! I'm happy to start working on a PR for these if it sounds like changes you'd like to make. I'm most unsure about adding another props getter, and what to call it.

kentcdodds:
This is all fantastic! If you could make a pull request for these changes, we can iterate on the solution, and then I'd love to have it merged!

Instead of `getItemListProps` we could call it `getMenuProps` because I think that makes the most sense for folks using downshift.

Thanks again!

kentcdodds:
This has been fixed and merged in the `next` branch. Should be part of v2.0.0 which should hopefully be released in the next week or so.

PR #285, opened 2017-12-18
ARIA Updates to Match W3C Recommendations

cycomachead:
**What**:
Follows #283.

This makes downshift follow the recommended approaches listed here:
https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html

**Why**:

There are a couple minor violations of the ARIA spec, such as the requirement that an element with a role of `textbox` (an `input` element) be a child of the main `combobox` element. The `aria-expanded` attributes also belong there instead of on the input.

Additionally, there are some roles (`listbox` which belongs on the menu, and `option` which belongs on each item) that are added.

**How**:

Mostly, this follows existing patterns. I've added a `getMenuProps` function which is designed to be called once on the element containing the list of items.

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [ ] Ready to be merged
      
### My TODOs and Questions

- [ ] Should we update the docs to suggest that `getLabelProps` is recommended or required? We are now assigning an `id` to the label element, so it makes more sense to use is instead of providing your own `id` everywhere. (though, of course, this is still functional).
- [ ] Code: I'm not sure how much error handling should be done around mismatched `id`s. Perhaps we could/should write an `assertIdsMatch()` function? I'm happy to C&P for now, but wanted to leave this open.
- [ ] TESTS 😮 
- [ ] Add at least one example which uses `ul` and `li`

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,26 +61,31 @@ harder to contribute to.
 
 * [Installation](#installation)
 * [Usage](#usage)
-* [Props](#props)
+* [Basic Props](#basic-props)
+  * [render](#render)
+  * [itemToString](#itemtostring)
+  * [onChange](#onchange)
+  * [stateReducer](#statereducer)
+* [Advanced Props](#advanced-props)
   * [defaultSelectedItem](#defaultselecteditem)
   * [defaultHighlightedIndex](#defaulthighlightedindex)
   * [defaultInputValue](#defaultinputvalue)
   * [defaultIsOpen](#defaultisopen)
-  * [itemToString](#itemtostring)
   * [selectedItemChanged](#selecteditemchanged)
   * [getA11yStatusMessage](#geta11ystatusmessage)
-  * [onChange](#onchange)
   * [onSelect](#onselect)
   * [onStateChange](#onstatechange)
-  * [stateReducer](#statereducer)
   * [onInputValueChange](#oninputvaluechange)
   * [itemCount](#itemcount)
   * [highlightedIndex](#highlightedindex)
   * [inputValue](#inputvalue)
   * [isOpen](#isopen)
   * [selectedItem](#selecteditem)
-  * [render](#render)
   * [id](#id)
+  * [inputId](#inputid)
+  * [labelId](#labelid)
+  * [menuId](#menuid)
+  * [getItemId](#getitemid)
   * [environment](#environment)
   * [onOuterClick](#onouterclick)
 * [stateChangeTypes](#statechangetypes)
@@ -134,36 +139,38 @@ function BasicAutocomplete({items, onChange}) {
       render={({
         getInputProps,
         getItemProps,
+        getMenuProps,
         isOpen,
         inputValue,
         selectedItem,
         highlightedIndex,
       }) => (
         <div>
-          <input {...getInputProps({placeholder: 'Favorite fruit ?'})} />
-          {isOpen ? (
-            <div style={{border: '1px solid #ccc'}}>
-              {items
-                .filter(
-                  i =>
-                    !inputValue ||
-                    i.toLowerCase().includes(inputValue.toLowerCase()),
-                )
-                .map((item, index) => (
-                  <div
-                    {...getItemProps({item})}
-                    key={item}
-                    style={{
-                      backgroundColor:
-                        highlightedIndex === index ? 'gray' : 'white',
-                      fontWeight: selectedItem === item ? 'bold' : 'normal',
-                    }}
-                  >
-                    {item}
-                  </div>
-                ))}
-            </div>
-          ) : null}
+          <label {...getLabelProps()}>What's your favorite fruit?</label>
+          <input {...getInputProps({placeholder: 'Enter your favorite'})} />
+          <ul {...getMenuProps()} style={{border: '1px solid #ccc'}}>
+            {isOpen
+              ? items
+                  .filter(
+                    i =>
+                      !inputValue ||
+                      i.toLowerCase().includes(inputValue.toLowerCase()),
+                  )
+                  .map((item, index) => (
+                    <li
+                      {...getItemProps({item})}
+                      key={item}
+                      style={{
+                        backgroundColor:
+                          highlightedIndex === index ? 'gray' : 'white',
+                        fontWeight: selectedItem === item ? 'bold' : 'normal',
+                      }}
+                    >
+                      {item}
+                    </li>
+                  ))
+              : null}
+          </ul>
         </div>
       )}
     />
@@ -184,31 +191,17 @@ function App() {
 calls the render function and renders that. ["Use a render
 prop!"][use-a-render-prop]! `<Downshift render={/* your JSX here! */} />`.
 
-## Props
+## Basic Props
 
-### defaultSelectedItem
+This is the list of props that you should probably know about. There are some
+[advanced props](#advanced-props) below as well.
 
-> `any` | defaults to `null`
+### render
 
-Pass an item or an array of items that should be selected by default.
+> `function({})` | _required_
 
-### defaultHighlightedIndex
-
-> `number`/`null` | defaults to `null`
-
-This is the initial index to highlight when the menu first opens.
-
-### defaultInputValue
-
-> `string` | defaults to `''`
-
-This is the initial input value.
-
-### defaultIsOpen
-
-> `boolean` | defaults to `false`
-
-This is the initial `isOpen` value.
+This is called with an object. Read more about the properties of this object in
+the section "[Render Prop Function](#render-prop-function)".
 
 ### itemToString
 
@@ -216,42 +209,6 @@ This is the initial `isOpen` value.
 
 Used to determine the string value for the selected item (which is used to
 compute the `inputValue`).
-
-### selectedItemChanged
-
-> `function(prevItem: any, item: any)` | defaults to: `(prevItem, item) => (prevItem !== item)`
-
-Used to determine if the new `selectedItem` has changed compared to the previous
-`selectedItem` and properly update Downshift's internal state.
-
-### getA11yStatusMessage
-
-> `function({/* see below */})` | default messages provided in English
-
-This function is passed as props to a `Status` component nested within and
-allows you to create your own assertive ARIA statuses.
-
-A default `getA11yStatusMessage` function is provided that will check
-`resultCount` and return "No results." or if there are results but no item is
-highlighted, "`resultCount` results are available, use up and down arrow keys to
-navigate." If an item is highlighted it will run `itemToString(highlightedItem)`
-and display the value of the `highlightedItem`.
-
-The object you are passed to generate your status message has the following
-properties:
-
-<!-- This table was generated via http://www.tablesgenerator.com/markdown_tables -->
-
-| property              | type            | description                                                                                  |
-| --------------------- | --------------- | -------------------------------------------------------------------------------------------- |
-| `highlightedIndex`    | `number`/`null` | The currently highlighted index                                                              |
-| `highlightedItem`     | `any`           | The value of the highlighted item                                                            |
-| `inputValue`          | `string`        | The current input value                                                                      |
-| `isOpen`              | `boolean`       | The `isOpen` state                                                                           |
-| `itemToString`        | `function(any)` | The `itemToString` function (see props) for getting the string value from one of the options |
-| `previousResultCount` | `number`        | The total items showing in the dropdown the last time the status was updated                 |
-| `resultCount`         | `number`        | The total items showing in the dropdown                                                      |
-| `selectedItem`        | `any`           | The value of the currently selected item                                                     |
 
 ### onChange
 
@@ -265,41 +222,6 @@ with the item that was selected and the new state of `downshift`. (see
 * `selectedItem`: The item that was just selected
 * `stateAndHelpers`: This is the same thing your `render` prop function is
   called with (see [Render Prop Function](#render-prop-function))
-
-### onSelect
-
-> `function(selectedItem: any, stateAndHelpers: object)` | optional, no useful
-> default
-
-Called when the user selects an item, regardless of the previous selected item.
-Called with the item that was selected and the new state of `downshift`. (see
-`onStateChange` for more info on `stateAndHelpers`).
-
-* `selectedItem`: The item that was just selected
-* `stateAndHelpers`: This is the same thing your `render` prop function is
-  called with (see [Render Prop Function](#render-prop-function))
-
-### onStateChange
-
-> `function(changes: object, stateAndHelpers: object)` | optional, no useful
-> default
-
-This function is called anytime the internal state changes. This can be useful
-if you're using downshift as a "controlled" component, where you manage some or
-all of the state (e.g. isOpen, selectedItem, highlightedIndex, etc) and then
-pass it as props, rather than letting downshift control all its state itself.
-The parameters both take the shape of internal state (`{highlightedIndex: number, inputValue: string, isOpen: boolean, selectedItem: any}`) but differ
-slightly.
-
-* `changes`: These are the properties that actually have changed since the last
-  state change. This also has a `type` property which you can learn more about
-  in the [`stateChangeTypes`](#statechangetypes) section.
-* `stateAndHelpers`: This is the exact same thing your `render` prop function is
-  called with (see [Render Prop Function](#render-prop-function))
-
-> Tip: This function will be called any time _any_ state is changed. The best
-> way to determine whether any particular state was changed, you can use
-> `changes.hasOwnProperty('propName')`.
 
 ### stateReducer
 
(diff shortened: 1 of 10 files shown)


cycomachead:
This is definitely a WIP, but I think should be enough to get a feel for how things will look. Please feel free to give me comments / feedback / thoughts. 

It's the holidays, so I'm not sure how much extra time I'll have but I should have pretty good access to GitHub/etc. :)

README.md
```diff
@@ -495,6 +495,19 @@ There are no required properties for this method.
 > accessibility). However, we include this so you don't forget and it makes
 > things a little nicer for you. You're welcome 😀
 
+#### `getMenuProps`
+
+This method should be applied to the element which contains your list of items. Typically, this will be a `<div>` or a `<ul>` that surrounds a `map` expression.
+This handles the proper ARIA roles and attributes.
+
+TODO: optional arguments
```
kentcdodds: Make sure we don't forget this TODO :wink:

src/downshift.js
```diff
@@ -478,13 +480,32 @@ class Downshift extends Component {
...
     {suppressRefError = false} = {},
   ) => {
+    const {isOpen} = this.getState()
     // this is used in the render to know whether the user has called getRootProps.
     // It uses that to know whether to apply the props automatically
     this.getRootProps.called = true
     this.getRootProps.refKey = refKey
     this.getRootProps.suppressRefError = suppressRefError
+
+    // TODO: Error Handling?
```
kentcdodds: Yes, let's do that :+1: Similar to how the label's `htmlFor` and `inputId` are validated. Except instead of throwing an error, let's use `console.error` and wrap it all in `process.env.NODE_ENV === 'production'` so it can be eliminated for production. In a future breaking version of `downshift`, I want to move all the validation to this to make downshift smaller.

src/downshift.js
```diff
@@ -680,6 +722,28 @@ class Downshift extends Component {
   }
   //\\\\\\\\\\\\\\\\\\\\\\\\\\\\\ INPUT
 
+  /////////////////////////////// MENU
+  getMenuProps = (props = {}) => {
+    // TODO: Error handling, mismatched Ids?
```
kentcdodds: Yup, we'll want it here too :+1:

kentcdodds:
ping @cycomachead would you like someone else to take this over? Looks like @mperrotti would like it https://github.com/paypal/downshift/issues/287#issuecomment-361974471 and is willing to help :)

cycomachead:
Hey, I'd love some help! I'm also happy to keep discussing options or to let someone push to this branch.

I haven't forgotten this, but I've been super swamped with other projects. (A project I help maintain is migrating cloud service providers and rewritten app, so my time's been vacuumed up.)

kentcdodds:
That's fine @cycomachead! I can't blame you. It took me forever to review it for you anyway :wink:

Perhaps you could give @mperrotti commit access to your fork and he could continue the work :+1:

cycomachead:
Done! :)

mperrotti:
Thanks @cycomachead! I'll be able to get on this Sunday or Monday

kentcdodds:
Any updates here?

kentcdodds:
Alright! So I had some git trouble and I'm afraid that I had to basically start over (so now I'm the author of the commit, sorry :-( but you're still on the contributor's table!)

I'm feeling pretty good about the changes. If folks could give this a review that would be great!

src/downshift.js
```diff
@@ -543,6 +558,8 @@ class Downshift extends Component {
       'aria-label': isOpen ? 'close menu' : 'open menu',
       'aria-expanded': isOpen,
       'aria-haspopup': true,
+      // TODO: is this supposed to be here? This already has `aria-label`
+      'aria-labelledby': this.labelId,
```
kentcdodds: In particular, I'm wondering about this. cc @cycomachead

src/downshift.js
```diff
@@ -587,44 +604,15 @@ class Downshift extends Component {
...
-        }" for your label, but the id of your input is "${
-          this.inputId
-        }". You must either remove the id from your input or set the htmlFor of the label equal to the input id.`,
-      )
-    }
-    this.inputId = firstDefined(this.inputId, props.htmlFor, `${this.id}-input`)
-    return {
-      ...props,
-      htmlFor: this.inputId,
-    }
```
kentcdodds: Seriously simplified things by adding the new props. We no longer validate for you, but you can set your ids earlier on and we don't have to validate. You can still set ids directly on elements like before, but you wont get validation anymore. I think that's an ok tradeoff.

src/downshift.js
cycomachead: Hmm, I can't recall exactly why that was there. Looking that the ARAI 1.1 examples (see ex 3), it seems like the existing aria-label should be sufficient: https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/combobox/aria1.1pattern/listbox-combo.html

cycomachead:
Wow, thanks so much for helping see this through! :) Happy to help, and sorry things have been a bit crazy lately. (No worries about the commit history either!)

I gave the latest updates a short read-through and so far everything makes sense. Totally on board with the simplifications that come with removing the extra validation checks.

kentcdodds:

cycomachead:
Wow, thanks! I really appreciate that!

I love seeing the code of conduct and the time (and care) that's been put into making a project that's accessible from the contribution side as well its implementation. :)

src/__tests__/__snapshots__/downshift.aria.js.snap
```diff
@@ -0,0 +1,107 @@
...
+  </label>
+  <input
+    aria-autocomplete="list"
+    aria-controls="downshift-0-menu"
+    aria-expanded="false"
+    aria-labelledby="downshift-0-label"
+    autocomplete="off"
+    data-testid="input"
+    id="downshift-0-input"
+    role="listbox"
```
1Copenut: I'd recommend removing the `role=listbox` attribute here. It's throwing an error in the axe-core Chrome plugin, and is declared further down on your unordered list.

I'm using https://www.w3.org/TR/wai-aria-1.1/#combobox as my guide and cross-reference for any errors thrown by axe-core.

src/__tests__/__snapshots__/downshift.aria.js.snap
```diff
@@ -0,0 +1,107 @@
...
+    data-testid="label"
+    for="downshift-0-input"
+    id="downshift-0-label"
+  >
+    label
+  </label>
+  <input
+    aria-autocomplete="list"
+    aria-controls="downshift-0-menu"
+    aria-expanded="false"
```
1Copenut: You can safely remove `aria-expanded...` from the input, where you've defined it on the containing div.

src/__tests__/__snapshots__/downshift.aria.js.snap
```diff
@@ -0,0 +1,107 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`basic snapshot 1`] = `
+<div
+  aria-expanded="false"
+  aria-haspopup="listbox"
+  aria-labelledby="downshift-0-label"
+  aria-owns="downshift-0-menu"
```
1Copenut: The `aria-owns="downshift-0-menu"` attribute is throwing an error on the Chrome axe-core plugin because it's referencing the unordered list. I have experimented with this a little; if you can append aria-owns when `aria-expanded` state is true, the error will go away. It's also okay to have `aria-owns` removed when Downshift is collapsed.

There's support for this idea in the [Rich Internet Applications combobox spec](https://www.w3.org/TR/wai-aria-1.1/#combobox), version 1.1:

> When a combobox is expanded, authors MUST ensure it contains or **owns** an element that has a role of listbox, tree, grid, or dialog. This element is the combobox popup. When the combobox is expanded, authors MUST set aria-controls on the textbox element to a value that refers to the combobox popup element.

src/__tests__/__snapshots__/downshift.aria.js.snap
```diff
@@ -0,0 +1,107 @@
...
+  <label
+    data-testid="label"
+    for="downshift-0-input"
+    id="downshift-0-label"
+  >
+    label
+  </label>
+  <input
+    aria-autocomplete="list"
+    aria-controls="downshift-0-menu"
```
1Copenut: The `aria-controls="downshift-0-menu"` attribute is also throwing an error because it's referencing the unordered list of results when they're not available. If you can add/remove this attribute at the same time you add/remove the `aria-owns` attribute on the container, the error will go away.

The axe-core scanner does not report an error if Downshift is collapsed, and `aria-controls` is removed. You could tie the `aria-controls` attribute to the `aria-expanded` logic.

src/__tests__/__snapshots__/downshift.aria.js.snap
```diff
@@ -0,0 +1,107 @@
...
+    aria-expanded="false"
+    aria-haspopup="true"
+    aria-label="open menu"
+    data-testid="button"
+    data-toggle="true"
+    role="button"
+    type="button"
+  />
+  <ul
+    aria-labelledby="downshift-0-label"
```
1Copenut: If you wanted this label to be more explicit about what the results list included, it could be changed to an `aria-label` with a prop to let users enter their own string.

kentcdodds:
This is all very helpful @1Copenut! Thank you very much. I've opened a new PR that addresses your suggestions. https://github.com/paypal/downshift/pull/396

Thank you again!

src/__tests__/__snapshots__/downshift.aria.js.snap
kentcdodds: This is a great suggestion. Now if you pass `aria-label` to `getMenuProps` we'll use that instead of `aria-labelledby`. It works perfectly :+1:

src/__tests__/__snapshots__/downshift.aria.js.snap
SiTaggart: `aria-owns` is redundant here. Owns is an explicit way to say soemthing is a child of this element, without is having to be in the DOM. Your `listbox` is a direct DOM child so you can remove it.

src/__tests__/__snapshots__/downshift.aria.js.snap
SiTaggart: +1

src/__tests__/__snapshots__/downshift.aria.js.snap
```diff
@@ -0,0 +1,107 @@
...
+    for="downshift-0-input"
+    id="downshift-0-label"
+  >
+    label
+  </label>
+  <input
+    aria-autocomplete="list"
+    aria-controls="downshift-0-menu"
+    aria-expanded="false"
+    aria-labelledby="downshift-0-label"
```
SiTaggart: This is redundant if you have correctly associated your label element with the input, like you have with the `for` attribute.

src/__tests__/__snapshots__/downshift.aria.js.snap
SiTaggart: You absolutely need to remove this role. At its core, a Combobox is a composite of a `textbox` and a "popup", most commonly a `listbox`. The textbox is important, it's what takes the value once set, what indicates you can type and edit, the thing that takes focus when interacting with the control, etc. You can just remove it as an `ipnut` has an implicit role of `textbox`, or you can explicitly set it to `role="textbox`

src/__tests__/__snapshots__/downshift.aria.js.snap
```diff
@@ -0,0 +1,107 @@
...
+    autocomplete="off"
+    data-testid="input"
+    id="downshift-0-input"
+    role="listbox"
+    value="item"
+  />
+  <button
```
SiTaggart: This button is essentially redundant. The input does all the work for you here, and focus should remain in the input when a user is interacting with the Combobox. If I actioned this, you'd have to focus me back into the input so I can interact with the options correctly. Feel free to keep an icon or something that visually depicts it's dropdown-y-ness, but it's easier to just remove the button.

The `listbox` often shows on focus on the input, or pressing enter and/or down arrow when focused in the input.

src/__tests__/__snapshots__/downshift.aria.js.snap
SiTaggart: If you want to be labelling explicitly the listbox options, which maybe contextually different than that of the overall label, what you are going to want to try to replicate semantically is a `select[multiple]` with a `optgroup` e.g.

```
<select multiple>
  <optgroup label="pets">
    <option>Cat</option>
    <option>Dog</option>
    <option>Fish</option>
  </optgroup>
</select>
```

The psuedo markup structure you want to recreate that is 

```
<div role="listbox">
  <div role="group" aria-label="Pets">
    <span role="none">Pets</span> <!-- optgroup shows a visible label, you could choose to exclude this but it's nice for a visual grouping -->
    <span role="option>Dog</span>
    <span role="option>Cat</span>
    <span role="option>Fish</span>  
  </div>
</div>
```

src/__tests__/__snapshots__/downshift.aria.js.snap
```diff
@@ -0,0 +1,107 @@
...
+    type="button"
+  />
+  <ul
+    aria-labelledby="downshift-0-label"
+    data-testid="menu"
+    id="downshift-0-menu"
+    role="listbox"
+  >
+    <li
+      aria-selected="true"
```
SiTaggart: I see this is set to `true` because it is the option that matches the set value of the `input`, just a note to make sure it's set to true also when it is "in focus" or the active options I've arrowed to and is the current value of `aria-activedescendant` on the `input.

src/__tests__/__snapshots__/downshift.aria.js.snap
```diff
@@ -0,0 +1,107 @@
...
+    aria-autocomplete="list"
+    aria-controls="downshift-0-menu"
+    aria-expanded="false"
+    aria-labelledby="downshift-0-label"
+    autocomplete="off"
+    data-testid="input"
+    id="downshift-0-input"
+    role="listbox"
+    value="item"
+  />
```
SiTaggart: The snapshot doesn't cover mid-interaction, i.e. the input has my focus and I have used my arrow keys to "focus" an option. When doing so set `aria-activedescendant` on the input and the value of it should match the ID of the option currently selected or "in focus"

SiTaggart: Which I see you are doing in the code, so you should be good. Might just be missing setting `aria-selected="true"` on the highlighted item when you're interacting with it, not just when you've set the overall value. (this is for a VoiceOver quirk mainly)

src/__tests__/__snapshots__/downshift.aria.js.snap
kentcdodds: That's helpful! Is it harmful to include? Some use cases will have the listbox NOT be a direct DOM child. It'd be easier to just include it if it's not harmful?

src/__tests__/__snapshots__/downshift.aria.js.snap
kentcdodds: Interesting! Thanks. I think this isn't the level of abstraction for that (lots of use cases for this component), but perhaps we should have an example for this 🤔

src/__tests__/__snapshots__/downshift.aria.js.snap
kentcdodds: What about mouse-only users?

src/__tests__/__snapshots__/downshift.aria.js.snap
kentcdodds: It depends I think. The default use-case for downshift is that the selected item has to be explicitly selected. Then there's a concept of a highlighted item which can only be selected when clicked or on enter-keypress. This behavior can be changed to select the highlighted item, and in that case this will behave properly.

src/__tests__/__snapshots__/downshift.aria.js.snap
kentcdodds: We do other stuff for VoiceOver saying the highlighted item. If you want to try this component out with VoiceOver, give this a look: https://deploy-preview-396--downshift.netlify.com/

kentcdodds:
Thanks for your feedback @SiTaggart!

src/__tests__/__snapshots__/downshift.aria.js.snap
SiTaggart: Absolutely, no harm is done when you add it regardless of DOM tree (as long as it matches the ID correctly of course). As you point out, super useful when using a positioning lib.

src/__tests__/__snapshots__/downshift.aria.js.snap
SiTaggart: Click event on the input? The input is a much better pointer target in general because of its size. 

You can split the interaction too if you desire:

- onClick will show the listbox 
- onFocus won’t
- onKeyPress `down` and `enters will show

Or whatever suits the full desired interaction. Sometimes it’s useful to not show the dropdown as you’re tabbing through a form for example, and only show it when I desire, similar to a native select element.

src/__tests__/__snapshots__/downshift.aria.js.snap
SiTaggart: Sure, I guess i’m Aware of an old voiceover quirk. If you don’t set `aria-selected=“true”` when `aria-activedescendant` matches an option ID, it often won’t announce the option.

src/__tests__/__snapshots__/downshift.aria.js.snap
SiTaggart: Having a "clear" button is great though! 👍

src/__tests__/__snapshots__/downshift.aria.js.snap
SiTaggart: If you want my advice, I'd remove all that stuff and leave it up to the ARIA implementation to do the work. The live region will be fighting with this now better ARIA implementation, and what you might be making up for is poor VoiceOver support, but in JAWS and NVDA it works perfectly and you'll get weird UX side effects for those folks. Arguably it's far more common to use those over VoiceOver anyway.

By correctly setting `aria-activedescendant` with `aria-selected="true"` on the maatching option, you'll get VoiceOver announcing the highlighted option in Chrome easily. Safari is broken, but that's not your fault. Don't make other more common use cases suffer, because of Safari.

I often see implementations calling out the number of available option too, but there really is no need. If my first down arrow press is to open _and_ highlight the first option, the screen reader will tell me how many options there are. It'll say something like "Black, 1 of 6". 

You totally don't need any of that jank.

kentcdodds: Hmmm... I'll have to test things out myself I guess. This stuff is hard! I wish the standard were more standard. I feel like I'm coding in the IE6 days. Thanks for your feedback!

Issue #301, opened 2018-01-08
Focus not reapplied to input when an item is selected with the mouse


MarkFalconbridge:
* `downshift` version: 1.3

What you did:

Using the example [here](https://codesandbox.io/s/zx1kj58npl), open the drop down and select an item with the mouse, once selection has happened, focus should go back to the input element but it doesn't. Focus does go back to the input element if you use the keyboard (actually I suspect it never leaves the input) to select an item so mouse selection should do the same.

What happened:

Focus isn't returned to the input.

It is very easy to work around this by handling the onSelect prop but I had expected this to be taken care of by default.

kentcdodds:
Yes, I agree this should be handled by default. Would you like to [make a pull request](http://makeapullrequest.com)?

donysukardi:
But the expected behavior when using keyboard and mouse should be slightly different.

If user clicks the arrow button to open the dropdown and make a selection, I'm not entirely sure if user actually expects the input to get focus (and opens up the keyboard on mobile).

kentcdodds:
I think that if the user's using the mouse they don't care where the focus goes after they click an item. Makes just as much sense for it to go to the input as the body (which is where it would go otherwise).

PR #307, opened 2018-01-13
Adding focus back to input after mouse click

curtiswilkinson:
**What**: This implements the enhancement outlined in https://github.com/paypal/downshift/issues/301

**Why**: I agree that it's nice to have consistency between mouse/keyboard selection of items

**How**: I added a class method for focusing the input so that I could re-use existing input selection code

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged 

Hopefully y'all agree with the idea behind this change, it might be worth checking out in storybook to see how it feels :)

Have a great weekend!


src/downshift.js
```diff
@@ -266,13 +266,17 @@ class Downshift extends Component {
...
       () => {
-        const inputNode = this._rootNode.querySelector(`#${this.inputId}`)
-        inputNode && inputNode.focus && inputNode.focus()
+        this.focusInput()
         cbToCb(cb)()
       },
     )
   }
 
+  focusInput = () => {
```
kentcdodds: This can actually be a regular method because we don't pass it as a callback to anything: `focusInput() {`

src/downshift.js
```diff
@@ -287,7 +291,10 @@ class Downshift extends Component {
             : this.props.itemToString(item),
         ...otherStateToSet,
       },
-      cbToCb(cb),
+      () => {
+        this.focusInput()
+        cbToCb(cb)()
```
kentcdodds: Let's forward the arguments and we don't need the cbToCb utility in this situation:

```js
(...args) => {
  this.focusInput()
  typeof cb === 'function' && cb(...args)
}
```

curtiswilkinson:
Okay, all done! Thanks for getting feedback to me so quickly

I had to quickly add another test that exercised the `typeof` guard for `call(..args)` because it was dropping coverage, hopefully that's okay :)

PR #308, opened 2018-01-15
[bug][hotfix] Fix focusInput() by using getElementById

brikou:
**What**:

Updated querSelector to use escaped selector

**Why**:

Because if my id is "fields[0].name" then focusInput throw an exception

**How**:

easily, I just used my keyboard (and my fingers)

**Checklist**:

* [ ] Documentation
* [x] Tests
* [x] Ready to be merged
      

kentcdodds:
Thanks a bunch!

Issue #319, opened 2018-01-30
"closeOnSelection" property(Multiple selection out of box)


notruth:
I want to propose new Prop named `closeOnSelection`.

What happened, Problem description:
[This codesandbox-example](https://codesandbox.io/s/y0oo25o591)([Your Multi-select example](https://codesandbox.io/s/W6gyJ30kn) with latest Downshift version) contains hardcoded solution: Menu doesn't close(as it were), it open after close. Because of this, the `highlightedIndex` is reset; as well as existence of a lot of handlers that just control the state of `isOpen`.

Suggested solution:

```javascript
  defaultProps = {
    // ...
    closeOnSelection: true,
    // ...
  }
  
  selectItem = (item, otherStateToSet, cb) => {
    otherStateToSet = pickState(otherStateToSet)
    this.internalSetState(
      {
        ...this.props.closeOnSelection
          ? {
              isOpen: false,
              highlightedIndex: this.props.defaultHighlightedIndex, 
            }
          : {},
        selectedItem: item,
        inputValue:
          this.isControlledProp('selectedItem') &&
          this.props.breakingChanges.resetInputOnSelection
            ? this.props.defaultInputValue
            : this.props.itemToString(item),
        ...otherStateToSet,
      },
      (...args) => {
        this.focusInput()
        typeof cb === 'function' && cb(...args)
      },
    )
  }
```

**!NOTE:** At now this will work fine only for the case with `Input` element, because it's [focusing after select]( https://github.com/paypal/downshift/blob/master/src/downshift.js#L296).

kentcdodds:
Seems like a good idea!

What if we made the solution more powerful. What if we added a prop called `modifyStateChange` and we put it [here](https://github.com/paypal/downshift/blob/e4846d86544a9a23ed05d8573ee9d6f5eb0ca054/src/downshift.js#L347):

```js
stateToSet = this.props.modifyStateChange(stateToSet, state)
```

Where the default would simply return `stateToSet` but you could provide your own function that could modify the state that will be set. You could check the `type` property of `stateToSet` against [this list](https://github.com/paypal/downshift/blob/e4846d86544a9a23ed05d8573ee9d6f5eb0ca054/src/downshift.js#L109-L124) to know whether you want to modify how the state is changed.

This would actually allow people to not have to control state quite as much I think which would be pretty handy!

What do you think?

notruth:
@kentcdodds, Yeah! Looks great! i like this solution 👍, thanks.
[...]

PR #320, opened 2018-01-30
Add new prop 'stateReducer'

notruth:
**What**:

At now, you could provide your own function that could modify the state that will be set. You could check the type property of stateToSet against this list to know whether you want to modify how the state is changed. [@kentcdodds](https://github.com/kentcdodds)

**Why**:
Chain of ideas: [#319](https://github.com/paypal/downshift/issues/319).

**How**:

```javascript
  internalSetState(stateToSet, cb) {
    // ...
    return this.setState(
      state => {
        state = this.getState(state)
        stateToSet = isStateToSetFunction ? stateToSet(state) : stateToSet

        // Your own function that could modify the state that will be set.
        stateToSet = this.props.modifyStateChange(stateToSet, state)
        
        // ...
      },
    // ...
  }
```

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [ ] Ready to be merged
      

Tests and documentation I can add tomorrow(if nobody else wants).
(modify my wording, if I stumbled somewhere 😁 )

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ harder to contribute to.
   * [id](#id)
   * [environment](#environment)
   * [onOuterClick](#onouterclick)
+  * [modifyStateChange](#modifystatechange)
 * [Control Props](#control-props)
 * [Render Prop Function](#render-prop-function)
   * [prop getters](#prop-getters)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -398,6 +399,49 @@ const ui = (
...
+modify the state change that will take place which can give you fine grain
+control over how the component interacts with user updates without having to
+use [Control Props](#control-props).
+
+* `state`: The full current state of downshift.
+* `stateToBeSet`: The state that is about to be set (including the `type` or
+  origin of the change which has references in `Downshift.stateChangeTypes`
+  as in [`onStateChange`](#onstatechange).
+
+This is a pure function. You _might_ think of it as a simple reducer!
```
kentcdodds: This is where I decided to switch the function signature to accept state first. I think some folks will get giddy about this :wink:

README.md
```diff
@@ -398,6 +399,49 @@ const ui = (
...
+    {/* your callback */}
+  </Downshift>
+)
+
+function modifyStateChange(state, stateToBeSet) {
+  // this prevents the menu from being closed when the user
+  // selects an item with the keyboard
+  switch (stateToSet.type) {
+    case Downshift.stateChangeTypes.keyDownEnter:
+    case Downshift.stateChangeTypes.clickItem:
```
kentcdodds: As I was copying your test over for the example I realized that if the user selected an item with the mouse it would still close and then realized we didn't have a `type` for that situation, so I added one in this PR :+1:

kentcdodds:
@notruth, could you give this a look over and make sure you like it? I'll merge it when you give me the 👍

README.md
notruth: I know, I planned to add `{ type:  Downshift.stateChangeTypes.itemClick }` [here](https://github.com/paypal/downshift/blob/master/src/downshift.js#L738) after that PR 😅.

kentcdodds:
Sweet. I'll merge this myself in an hour or two... I just tweeted about it and I'm sure we'll get some interesting feedback. I think maybe the prop name could be improved...

README.md
notruth: Right. State first its better; like simple reducer.

notruth:
That's a good idea thank you!

kentcdodds:
[Someone suggested the name `stateMiddleware`](https://twitter.com/_jpstone/status/958736750423236608). What do you think? Which is better `modifyStateChange` or `stateMiddleware`? I'm thinking that this is a pattern I want to encourage in the community. So the more general the better so there can be consensus and discoverability. That's why I'm leaning to `stateMiddleware`...

Though I actually kind of like `stateReducer` too 🤔

notruth:
`stateMiddleware` looks good.
Also can still like `stateChangeMiddleware` or `stateToSetMiddleware`...

kentcdodds:
What about `stateReducer`?

notruth:
Yes, probably better `reducer`, because arguments looks like reducer arguments)

kentcdodds:
Cool! I think I want to change it to `stateReducer`. Would you like to make that update? I can if not :)

notruth:
I can forget somewhere to replace wording) сould You do it?

kentcdodds:
Sure :+1:

geoffdavis92:
This looks awesome. Great work @notruth @kentcdodds

notruth:
Awesome, thanks!

kentcdodds:
Thank you @notruth!

PR #323, opened 2018-02-02
Focus restoration

notruth:
**What**:
Better(another) Focus restoration.

**Why**:
Previously, focus restored only on Input field(element with `getInputProps()`) and after [`.selectItem()`]() or [`.clearSelectrion()`]() called. This methods can be called manually, and the user hardly wants to focus a `input` after that. Secondly, the focus will no be restore in cases where the logic is not related to the input field.
In Downshift lifecycle only one event can deFocus current focused element, it's `clickItem` event(i mistaken?). Will not be better to solve the problem with the loss of focus upon this event cause?
I think focusing main trigger(`input` earlier) after selecting or cleaning should be left to the user because now the focus will not be lost after `item` clicks :).

**How**:
So, I just added `event.preventDefault()` into `item->onMouseDown` handler.

```javascript
  // ...
  getItemProps = ({
    // ...
  } = {}) => {
    // ...
    return {
      // ...
      onMouseDown: composeEventHandlers(onMouseDown, event => {
        // Avoid of change active element.
        event.preventDefault()
      }),
      onClick: composeEventHandlers(onClick, () => {
        this.selectItemAtIndex(index, {
          type: Downshift.stateChangeTypes.clickItem,
        })
      }),
      ...rest,
    }
  }
```
And removed code related to `.focusInput()`.

**Checklist**:

* [ ] Documentation
* [x] Tests
* [ ] Ready to be merged
      

All ok, if you don't agree this PR(solution, idea). Otherwise, help me pls with `Tests` and `Documentation`. Thanks!


notruth:
Wow! Thanks.

PR #341, opened 2018-02-23
test(cypress): add initial cypress tests!

kentcdodds:
**What**: Just for fun, I wanted to see what it would be like to add some cypress tests for the storybook. Turns out it was pretty easy and I want to do more!

**Why**: Increase our confidence that downshift works in a real browser

**How**:
- Install cypress, eslint-plugin-cypress, and serve
- Add a few scripts
- Upgrade some deps (mostly just because I was getting some annoying warnings due to outdated packages)
- Configure cypress (and eslint for the cypress folder)
- Update the basic example to be more testable
- Write tests for the basic example!

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged
      

kentcdodds:
cc @brian-mann :wink: This was a delight!

kentcdodds:
I added a few reviewers, but don't feel pressured. Also if I didn't add you and you would like to review, please feel free to do so :)

kentcdodds:
It's hard to tell from this gif, but cypress is pretty rad. Here's a recording it created when I ran it locally:

ezgif com-video-to-gif

@brian-mann, maybe you could help me understand why I get the initial load -> refresh -> run tests experience... I seem to get this in every project I use cypress.

kentcdodds:
Thanks!

jennifer-shehane:
@kentcdodds Adding a [`baseUrl` to your `cypress.json`](https://on.cypress.io/configuration#Global) should fix the 'flashing' of initial load -> refresh -> run tests. Cypress essentially does not know before opening where you are planning to visit (unless specified in `cypress.json`), so it initially opens on localhost + a random port, then switches to your localhost + port you specified in your visit.

kentcdodds:
Thanks @jennifer-shehane! I think you've told me that before actually. I still don't understand why that happens...

> Cypress essentially does not know before opening where you are planning to visit (unless specified in cypress.json)

That makes sense

> so it initially opens on localhost + a random port

This doesn't make sense to me. Could you help me understand why cypress opens to any port at all? Why doesn't it just wait until my `cy.visit` call before sending the browser anywhere? Also, it seems like the entire cypress app is refreshed, not just the browser. I think I'm missing something here 🤔

jennifer-shehane:
@kentcdodds So, this isn't evident in the video, but the url I'm talking about is the main Cypress browser url. 

**With baseUrl set -> loads main window in localhost set in baseUrl**
screen shot 2018-02-23 at 1 33 52 pm

**Without baseUrl set -> loads main window in localhost + random port**
screen shot 2018-02-23 at 1 34 26 pm

Cypress internally switches to the domain that you use in your cy.visit() so that your application’s code will run under the current domain at all times.

We had a lot of CORS related issues before when we did not set the main window to the same url as the application under test.  Previously we implemented an endless amount of hacks and internal translations to figure out the domain you were supposed to be on without actually being on the domain. This caused code to behave different and caused subtle issues. 

The entire proxy layer was written to handle https certificates, continue to inject (even on https pages), and know when to automatically bypass injection so you can open other tabs while testing in Cypress. This also allows us to unlock the ability to do things like whitelisting or blacklisting specific 3rd party domains, or even be able to stub not just XHR’s but any kind of HTTP request.

**TLDR:** iframes are complicated.

kentcdodds:
This makes total sense. Thank you for taking the time to explain! I'm sure it's somewhere, but somehow I missed this in the docs. If it's not there, then it would be a good idea to mention that 99% of the time people will want to set the `baseUrl` configuration option (whether that's via `cypress.json` or otherwise).

Thanks again!

Issue #340, opened 2018-02-22
Does resetIdCounter() work?


jaredpalmer:
Currently upgrading an app to React 16 and it seems that downshift is breaking hydration. To confirm, I was able to reproduce on a brand new Razzle app. 

See https://github.com/jaredpalmer/downshift-razzle-bug

Thoughts?

notruth:
@jaredpalmer, hm... 🤨
It seems that `webpack`(under the hood of the `razzle`) uses same [`idCounter`](https://github.com/paypal/downshift/blob/master/src/utils.js#L1) on both(client and server) sides. And this happens only in DEV mode...

I don't know why this feature was added in Downshift, but the solution is to shove [this variable](https://github.com/paypal/downshift/blob/master/src/utils.js#L1) and relevant logics into the `Downshift` constructor scope(i think).

I recommend You to pass `ID`s manually, like:
javascript```
<input {...getInputProps({
  id: 'my-input-id',
  placeholder: 'Favorite fruit ?'
})} />
```
this solution is cleaner;)

kentcdodds:
The feature is intended for server side rendering and there's actually an existing test for it. Not sure what is going on, but @notruth's workaround will work. You can also pass an ID prop to downshift directly.

But we should probably figure out what's going on 🤔

I won't have time to work on this myself though.

notruth:
@kentcdodds, The problem is that `webpack` uses the same [`idCounter`](https://github.com/paypal/downshift/blob/master/src/utils.js#L1) on both(client and server) sides in DEV mode. In `razzle` DEV mode client and server sides works parallel, and `webpack`(core of `razzle`) uses the same module(`downshift/utils.js`) on both sides.

I can add a PR tomorrow with the relevant decision.

notruth:
Simply put, the [`idCounter`](https://github.com/paypal/downshift/blob/master/src/utils.js#L1) variable shouldn't stick out. Because `webpack` uses the same variable address for both sides(If u increment it on server, it increments on client).

notruth:
I found.

This(root) [`generateId()`](https://github.com/paypal/downshift/blob/master/src/downshift.js#L72) calls in Init time and `idCounter` right away rises to `2`.

On **server-side** after that we reset the `idCounter` to `0` every render time.
It means that others `generateId()`'s begin counting from `0`.
But on **client-side** others `generateId()`'s begin counting from `2`, because no reset occurs.

[...]

PR #345, opened 2018-02-25
fix(generateId): fix initial idCounter and use single unique id for a Downshift instance

donysukardi:
**What**:

This fixes mismatch in client side and SSR generated Id and ensure that Downshift instance is tagged with a single unique id.

**Why**:

Refer to https://github.com/paypal/downshift/issues/340

**How**:

* Set the initial `idCounter` to 0 instead of 1 to match it with `resetIdCounter` call.
* Use `id` to construct all generated ids

**Checklist**:

* [ ] Documentation
* [x] Tests
* [x] Ready to be merged
      
### DIFF
--- a/src/__tests__/utils.reset-id-counter.js
+++ b/src/__tests__/utils.reset-id-counter.js
@@ -3,10 +3,6 @@ import {mount} from 'enzyme'
 import Downshift from '../'
 import {resetIdCounter} from '../utils'
 
-test('does not throw', () => {
-  expect(() => resetIdCounter()).not.toThrow()
-})
-
 test('renders with correct and predictable auto generated id upon resetIdCounter call', () => {
   resetIdCounter()
 
--- a/src/__tests__/utils.reset-id-counter.js
+++ b/src/__tests__/utils.reset-id-counter.js
@@ -10,35 +10,51 @@ test('does not throw', () => {
 test('renders with correct and predictable auto generated id upon resetIdCounter call', () => {
   resetIdCounter()
 
-  const render = ({getInputProps, getLabelProps}) => (
+  const render = ({getInputProps, getLabelProps, getItemProps}) => (
     <div>
       <label {...getLabelProps()} />
       <input {...getInputProps()} />
+      <div>
+        <span
+          {...getItemProps({
+            item: 0,
+            'data-test': 'item-0',
+          })}
+        >
+          0
+        </span>
+      </div>
     </div>
   )
 
   const {id, wrapper} = setup({render})
   const label = wrapper.find('label').first()
   const input = wrapper.find('input').first()
+  const item = wrapper.find('[data-test="item-0"]').first()
   expect(id).toBe('downshift-0')
-  expect(label.instance().getAttribute('for')).toBe('downshift-input-0')
-  expect(input.instance().getAttribute('id')).toBe('downshift-input-0')
+  expect(label.instance().getAttribute('for')).toBe('downshift-0-input')
+  expect(input.instance().getAttribute('id')).toBe('downshift-0-input')
+  expect(item.instance().getAttribute('id')).toBe('downshift-0-item-0')
 
   const {id: id2, wrapper: wrapper2} = setup({render})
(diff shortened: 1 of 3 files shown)


src/downshift.js
```diff
@@ -129,6 +129,8 @@ class Downshift extends Component {
       state.inputValue = this.props.itemToString(state.selectedItem)
     }
     this.state = state
+    this.uniqueId = generateId()
+    this.id = this.props.id || getPrefixedId('downshift', this.uniqueId)
```
kentcdodds: Any reason we couldn't just use a single `id` rather than have both `id` and `uniqueId`?

donysukardi: No particular reason actually. I see that there's currently no relationship between `id` in props and `id` passed in `getInputProps` or `htmlFor` in `getLabelProps`. Hence I decided to keep them separate.

donysukardi: Alternatively, we could derive the label's `htmlFor` and input's `id` from component's `id`,
```js
this.inputId = firstDefined(
  this.inputId,
  props.htmlFor,
  `${this.id}-input` // derive fr component id, yields 'downshift-0-input', etc
)
```

donysukardi: Added test for the auto generated ids 💥

src/__tests__/utils.reset-id-counter.js
```diff
@@ -1,5 +1,52 @@
+import React from 'react'
+import {mount} from 'enzyme'
+import Downshift from '../'
 import {resetIdCounter} from '../utils'
 
 test('does not throw', () => {
```
kentcdodds: We can remove this test. It's pretty useless now :+1:

src/downshift.js
kentcdodds: I was thinking it'd be:

```javascript
this.inputId = firstDefined(
  this.inputId,
  props.htmlFor,
  getPrefixedId('downshift-input-', this.id)
)
```

What do you think of that?

Or.... Actually that `getPrefixedId` util is kinda uesless. Maybe we should just do what you're talking about except as: `downshift-input-${this.id}`

What do you think? I don't really care all that much, but I'm starting to think the `getPrefixedId` should be removed...

donysukardi: Yeah.. We should just make it explicit using template literal (after all that's what it does), and also consistent with how we construct the id for items, i.e.

```
getItemId(index) {
  return `${this.id}-item-${index}`
}
```

With that, we'll have
```
this.id = this.props.id || `downshift-${this.uniqueId}` // e.g. downshift-0
```

and

```
this.inputId = firstDefined(
  this.inputId,
  props.htmlFor,
  `${this.id}-input` // downshift-0-input
)
```

kentcdodds: Let's do that!

donysukardi: Done! :boom:

src/downshift.js
```diff
@@ -129,6 +128,8 @@ class Downshift extends Component {
       state.inputValue = this.props.itemToString(state.selectedItem)
     }
     this.state = state
+    this.uniqueId = generateId()
+    this.id = this.props.id || `downshift-${this.uniqueId}`
```
kentcdodds: Could we just change this to:

```javascript
this.id = this.props.id || `downshift-${generateId()}`
```

src/utils.js
```diff
@@ -146,12 +146,11 @@ function composeEventHandlers(...fns) {
...
 /**
- * This generates a unique ID for all autocomplete inputs
- * @param {String} prefix the prefix for the id
+ * This generates a unique ID for an instance of Downshift
  * @return {String} the unique ID
  */
-function generateId(prefix) {
-  return `${prefix}-${idCounter++}`
+function generateId() {
+  return `${idCounter++}`
```
kentcdodds: nitpick: Since you'll be making a slight change anyway, I'd prefer this to be: `return String(idCounter++)` Mostly just because I like the explicit casting.

Thanks!

donysukardi:
Done and done 🎉

Issue #336, opened 2018-02-21
Browsing through the menu with shift key


tdeschryver:
* `downshift` version: 1.28.0
* `node` version: /
* `npm` (or `yarn`) version: /

Problem description:
I don't know if this is an issue or if it is intended, this is just an observation and I couldn't find a test for this use case.
When using the shift key together with arrow down while having one of the the last 5 items in the menu selected, the first item in the menu gets selected. 
The same happens with arrow up, while being at the first 5 items in the menu then the last item gets selected.

To reproduce it, you use any of the codesandbox examples or one of the storybook examples.

Suggested solution:
My suggestion would be to stay at the bottom/top of the menu, or to select the 'right' item.
But again, I might be wrong :sweat_smile:

austintackaberry:
It looks like this functionality is intentional based on this line [here](https://github.com/paypal/downshift/blob/83cb60beae498f48c454f9407e7a6bcf256879cb/src/downshift.js#L490). Based on the code and in testing, it seems like holding shift skips 5 items on each down arrow. I think the idea is that this allows you to scroll through the list more quickly.

I checked other autocomplete implementations like Google and Amazon. Holding the shift key down in Google search freezes the down arrow functionality. In Amazon, it has no effect. Personally, I really like this feature, but I do think it should be added to the docs, so that users are aware of it. Side note...does this feature have anything to do with the name? Down-shift...

tdeschryver:
@austintackaberry yea, I know it is a feature but it behaves a bit weird while being at the top or bottom of the menu.  
![downshift-example](https://user-images.githubusercontent.com/28659384/36477783-c642f814-1702-11e8-84b0-f35cfd2adeeb.gif)

austintackaberry:
Sorry, I misunderstood. If I am understanding correctly, your suggested solution is:

If you're 3 items from the bottom and you press shift + down, then instead of highlighting the top item (as it does now), it highlights the bottom item.

The only issue I see with that suggestion is if you have a very long list, and you're at the bottom, but you would like to be at the top. In that case, the current usage makes more sense I believe.

What do you think about this:

If you're 3 items from the bottom and you press shift + down, then instead of highlighting the top item (as it does now), it highlights the bottom item, but if you press shift + down again (or just press down) then it takes you to the top.

kentcdodds:
Yes, this is by design. It's a "power-user" feature

No, this feature has nothing to do with the name "downshift" 😅 

Yes, we should probably document it (any takers?)

[...]

Issue #330, opened 2018-02-13
Enter event on form wrapped input should not have e.preventDefault()


hzhu:
First of all thanks @kentcdodds for authoring this library! Also thanks to everyone else who's been helping Kent maintain this project ❤️.

* `downshift` version: 1.28.0
* `node` version: v8.1.3
* `yarn` version: 1.2.1

### What you did:
I type the string "black" (lowercase) into the form example in Downshift's storybook - then hit `ENTER`.

### What happened:
The form submission handler does not get called. However, my expectation is that the form submit handler gets called even though lowercase "black" isn't a part of the list. This is important when building autocomplete for search services. When a user hits `ENTER`, I want to still be able to execute a search on lowercase "black", even though it is not part of the list in the combobox:

<img width="572" alt="screen shot 2018-02-13 at 11 16 09 am" src="https://user-images.githubusercontent.com/1811365/36174480-4e1ed584-10c1-11e8-85a8-b99b629b610a.png">

### Reproduction repository:
Reproducible in Downshift Storybook in the [form example](https://github.com/paypal/downshift/blob/master/stories/examples/form.js).

### Problem description:
I think that a `<form>` submission handler should always be called on `ENTER`, even if the combo box list is open. I believe in the past, a `<form>` submission was not possible entirely, see #161. 

Pull request #161 made it possible to submit the `<form>`, but not when the combobox list `isOpen`.

### Suggested solution:
I am not sure I have an elegant solution. I am seeking the help of the community on this one. Please let me know if I got this all wrong and I'm simply not understanding the correct usage of Downshift.

A solution could be for Downshift to accept a prop `isForm` to Downshift. Then we will not [call `e.preventDefault`](https://github.com/paypal/downshift/blob/master/src/downshift.js#L506) when a user presses `ENTER` at all if the consumer of the component tells Downshift that a `<form>` wraps its `<input>`.

kentcdodds:
Hey @hzhu,

Thanks for the issue! I'm not sure of the best solution for this to be honest. I can think of several things you could do, but I'm not super fond of them.

I don't think I want to add a prop especially for this case. I think we can come up with something else. I'd love people to give their input to solving this problem. Anyone have ideas?

newyork-anthonyng:
@hzhu I was working on this [commit](https://github.com/newyork-anthonyng/downshift/commit/75377bf34f6981f018406756cfbffeab26751799) that seems to address the issue.

[...]

PR #346, opened 2018-02-26
Add keyDownHandlers prop

hzhu:
**What**: Allow downshift to accept a new prop called `keyDownHandlers`. 

**Why**: This can be useful when one wants to customize key down handler behavior. It allows users to address issues such as #330 and #336 by passing an handler object to the `keyDownHandlers` prop. For example, in #330 when downshift is wrapped in a `<form>`, pressing `ENTER` key should call the `<form>'`s onSubmit handler. This doesn't happen because internally, downshift calls `event.preventDefault()` - preventing the `<form>` from submitting.

By [passing a custom handler](https://github.com/paypal/downshift/issues/330#issuecomment-367159522) for `ENTER` key, we can use a handler which does not call `event.preventDefault()`. 

**How**: If `keyDownHandlers` is passed, the handlers in the object will be used when downshift handles a keyDown event. Specifically, the handler will be called in `input_handleKeyDown`.

**Checklist**:

* [x] Documentation
* [x] Tests
* [ ] Ready to be merged

P.S. Please let me know if something in this pull request isn't clear. I would love suggestions from the community as this is my first PR for Downshift! 

Thank you for reviewing this PR. Thanks @kentcdodds, @austintackaberry, @indiesquidge, and @newyork-anthonyng for having a conversation on issue #330. Thanks to @kentcdodds for coming up with a flexible solution ❤️.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ harder to contribute to.
   * [onSelect](#onselect)
   * [onStateChange](#onstatechange)
   * [stateReducer](#statereducer)
+  * [keyDownHandlers](#keydownhandlers)
   * [onInputValueChange](#oninputvaluechange)
   * [itemCount](#itemcount)
   * [highlightedIndex](#highlightedindex)
@@ -337,6 +338,31 @@ function stateReducer(state, changes) {
   }
 }
 ```
+
+### keyDownHandlers
+
+> `{ ...handlers }` | optional
+
+NOTE: Normally if you want to add keydown handlers to the button or input, you can simply pass `onKeyDown` to `getInputProps` and `getButtonProps` (respectively). However, if you want to override built-in handling of specific keys, you can use this prop.
+
+Also if you want to override the built-in behavior of _all_ the keys, you can add `event.preventDefault()` in your own handler and that will signal to downshift to skip it's built-in event handler.
+
+This prop allows you to use your own keydown handlers which can be handy if you want custom behavior. If a key and handler you pass is already defined in downshift, the custom handler you pass will be used instead.
+
+For example, if you create an Enter handler, this will not change the default behavior of the other keyDown events: ArrowUp, ArrowDown, and Escape.
+
+```jsx
+const handlers = {
+  Enter(event) {
+    // Custom behavior for ENTER keydown
+  },
+}
+
+const ui = <Downshift keyDownHandlers={handlers} />
+```
+
+⚠️ **This is an advanced feature and you should avoid using it** ⚠️
+This feature exposes the downshift instance in a way that could lead to your code breaking if we decide to change a method you are relying on.
 
 ### onInputValueChange
 
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -31,6 +31,7 @@ class Downshift extends Component {
     defaultIsOpen: PropTypes.bool,
     getA11yStatusMessage: PropTypes.func,
     itemToString: PropTypes.func,
+    keyDownHandlers: PropTypes.object,
     onChange: PropTypes.func,
     onSelect: PropTypes.func,
     onStateChange: PropTypes.func,
@@ -84,6 +85,7 @@ class Downshift extends Component {
       }
       return String(i)
     },
+    keyDownHandlers: {},
     onStateChange: () => {},
     onInputValueChange: () => {},
     onUserAction: () => {},
@@ -423,6 +425,7 @@ class Downshift extends Component {
       selectItemAtIndex,
       selectHighlightedItem,
       setHighlightedIndex,
+      moveHighlightedIndex,
       clearSelection,
       clearItems,
       reset,
@@ -446,6 +449,7 @@ class Downshift extends Component {
       selectItemAtIndex,
       selectHighlightedItem,
       setHighlightedIndex,
+      moveHighlightedIndex,
       clearSelection,
       clearItems,
       setItemCount,
@@ -502,7 +506,7 @@ class Downshift extends Component {
     },
 
     Enter(event) {
-      if (this.getState().isOpen) {
+      if (this.isOpen) {
         event.preventDefault()
         this.selectHighlightedItem({
           type: Downshift.stateChangeTypes.keyDownEnter,
@@ -668,8 +672,10 @@ class Downshift extends Component {
   }
 
   input_handleKeyDown = event => {
-    if (event.key && this.keyDownHandlers[event.key]) {
-      this.keyDownHandlers[event.key].call(this, event)
+    const handler =
+      this.props.keyDownHandlers[event.key] || this.keyDownHandlers[event.key]
+    if (handler) {
+      handler.call(this.getStateAndHelpers(), event)
     }
   }
 
--- a/src/__tests__/downshift.props.js
+++ b/src/__tests__/downshift.props.js
@@ -218,6 +218,23 @@ test('stateReducer customizes the final state after keyDownEnter handled', () =>
   )
 })
 
+test('calls key handlers when keyDownHandlers is passed', () => {
+  const keyDownHandlers = {
+    Enter: jest.fn(),
+    Escape: jest.fn(),
+    ArrowUp: jest.fn(),
+    ArrowDown: jest.fn(),
+  }
+  const keyPresses = Object.keys(keyDownHandlers)
+  const render = ({getInputProps}) => <input {...getInputProps()} />
+  const {wrapper} = setup({render, keyDownHandlers})
+
+  keyPresses.forEach(key => {
+    wrapper.find('input').simulate('keyDown', {key})
+    expect(keyDownHandlers[key]).toHaveBeenCalled()
+  })
+})
+
 function mouseDownAndUp(node) {
   node.dispatchEvent(new window.MouseEvent('mousedown', {bubbles: true}))
   node.dispatchEvent(new window.MouseEvent('mouseup', {bubbles: true}))


src/downshift.js
```diff
@@ -502,7 +506,7 @@ class Downshift extends Component {
     },
 
     Enter(event) {
-      if (this.getState().isOpen) {
+      if (this.isOpen) {
```
indiesquidge: Was this an intentional change? `isOpen` is not an instance property, it is pull from the internal state, right?

README.md
```diff
@@ -338,6 +339,34 @@ function stateReducer(state, changes) {
...
+const handlers = {
+  Enter(event) {
+    // Add customized behavior for enter keypress below
+    if (this.isOpen) {
+      this.selectHighlightedItem({
+        type: Downshift.stateChangeTypes.keyDownEnter,
+      })
+    }
+  }
+}
```
indiesquidge: Does this example imply that developers must also include the existing handler logic if they write their own? Perhaps I'm misunderstanding something, but I thought that this solution allowed for the user to write their own handler without having to worry about what Downshift was doing with that handler previously.

For example, I would have thought I could just write

```js
const handlers = {
  Enter (event) {
    // my custom behavior for ENTER event
  }
}
```

And Downshift would have "merged" my custom handler with whatever it needed to do internally on its own.

Correct me if I'm wrong, but the way I'm reading this example it seems like the developer is having to do the "merging" themselves by including Downshift's existing logic at the bottom of the handler themselves (which I don't think would even work because `this` will be different depending on the context).

kentcdodds: No, that's not the intention of the feature. The intention is to allow you to completely override the behavior for a specific key. If you want downshift to merge behavior, then you'll use an `onKeyDown` event handler yourself. This is why I wanted to add [this note](https://github.com/paypal/downshift/pull/346#discussion_r170666325).

hzhu: Including Downshift's existing logic is optional (technically it is the existing logic minus the `e.preventDefault()`). From my understanding, `this` would never differ here - the object returned from `getStateAndHelpers()` is always bound.

Would updating the example in the doc from this:
```jsx
const handlers = {
  Enter(event) {
    // Add customized behavior for enter keypress below
    if (this.isOpen) {
      this.selectHighlightedItem({
        type: Downshift.stateChangeTypes.keyDownEnter,
      })
    }
  }
}

const ui = <Downshift keyDownHandlers={handlers} />
```

to this:
```jsx
const handlers = {
  Enter (event) {
    // my custom behavior for ENTER event
  }
}

const ui = <Downshift keyDownHandlers={handlers} />
```

be more clear?

src/downshift.js
hzhu: By reverting it to how it originally was, `getState` needs to be exposed from the object returned by `getStateAndHelpers()`. This is necessary because `getStateAndHelpers` object is always bound:

```
  input_handleKeyDown = event => {
    const handler =
      this.props.keyDownHandlers[event.key] || this.keyDownHandlers[event.key]
    if (handler) {
      handler.call(this.getStateAndHelpers(), event) // always bound
    }
  }
```

I am running into a wall after this though. `getState()` throws an error because at some point, `this.state` is `undefined`:

screen shot 2018-02-26 at 11 34 13 am

I will keep investigating this issue in the mean time ⌛️ .

README.md
austintackaberry: Maybe it would be worth adding to the docs something explicit such as...

"For example, if you create an Enter handler, this will not change the default behavior of the other keyDown events: ArrowUp, ArrowDown, and Escape."

README.md
```diff
@@ -338,6 +339,34 @@ function stateReducer(state, changes) {
 }
 ```
 
+### keyDownHandlers
+
+> `{ ...handlers }` | optional
+
+Passing a `keyDownHandlers` object to this prop allows you to use your own key down handlers which can be handy if you want custom behavior.
```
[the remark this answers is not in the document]
hzhu: Thanks for the awesome suggestion Kent! Please let me know if I can better word anything else 😄 .

README.md
indiesquidge: > From my understanding, this would never differ here - the object returned from getStateAndHelpers() is always bound.

Ah, my apologies. I see now you are explicitly binding the context on the handler invocation.

src/downshift.js
```diff
@@ -668,8 +672,10 @@ class Downshift extends Component {
   }
 
   input_handleKeyDown = event => {
-    if (event.key && this.keyDownHandlers[event.key]) {
-      this.keyDownHandlers[event.key].call(this, event)
+    const handler =
+      this.props.keyDownHandlers[event.key] || this.keyDownHandlers[event.key]
+    if (handler) {
+      handler.call(this.getStateAndHelpers(), event)
```
indiesquidge: > [From my understanding, this would never differ here - the object returned from getStateAndHelpers() is always bound.](https://github.com/paypal/downshift/pull/346#discussion_r170706095)

Ah, my apologies. I see now you are binding the context on the handler invocation here.

Just trying to think from an outside perspective, would this not be confusing for someone attempting to implement custom behavior?

For example, if I want to call an instance method in my custom ENTER handler, I may write something like this

```jsx
const handlers = {
  Enter (event) {
    this.handleEnter()
  }
}

<Downshift keyDownHandlers={handlers} />
```

Where `handleEnter` is an instance method on my component. This code will break however, because from the internally bound context, `this` will not have any such `handleEnter` method.

README.md
```diff
@@ -338,6 +339,31 @@ function stateReducer(state, changes) {
...
+
+This prop allows you to use your own keydown handlers which can be handy if you want custom behavior. If a key and handler you pass is already defined in downshift, the custom handler you pass will be used instead.
+
+For example, if you create an Enter handler, this will not change the default behavior of the other keyDown events: ArrowUp, ArrowDown, and Escape.
+
+```jsx
+const handlers = {
+  Enter(event) {
+    // Custom behavior for ENTER keydown
+  },
```
indiesquidge: > [The intention is to allow you to completely override the behavior for a specific key. If you want downshift to merge behavior, then you'll use an onKeyDown event handler yourself.](https://github.com/paypal/downshift/pull/346#discussion_r170705698)

Does this not have any unforeseen consequences though? I see that you've mentioned using `onKeyDown` within `getInputProps` as a solution for both both #330 and #336, so I'm wondering when it would be a good idea to completely override the handler for a specific key.

I imagine that the code run within the default handlers is important to how Downshift works. By completely overriding it, the user (perhaps unintentionally) is forgoing any and all functionality that Downshift provides. That may sound obvious, but if you look at what things like #330 and #336 are attempting, it is not to remove all of Downshift's functionality; #330 is looking to get around the `event.preventDefault()` that stops the form submission when Downshift is open, but it does not wish to forgo Downshift's functionality to select the highlighted item if there is one.

I guess at this point I'm curious when it benefits the developer to use `keyDownHandlers` over a solution like adding `onKeyDown` to `getInputProps`.

src/downshift.js
kentcdodds: Maybe instead of binding things to `this`, we should just pass it as a second argument? I think using `this` could really be misleading...

README.md
kentcdodds: So, to prevent downshift from running its internal behavior, you must call `event.preventDefault()`. However, as you mentioned, in #330, the behavior that we don't like is the `event.preventDefault()` call. In #330, we don't the `event.preventDefault()` to be called at all. So we can neither leave things as they are nor prevent the default behavior.

> I imagine that the code run within the default handlers is important to how Downshift works.

Yes, it is, but when you're trying to customize how downshift works internally, you're expected to understand what the current behavior is. Luckily what downshift is doing for these is fairly simple (and is using functions that you can already use yourself anyway).

> By completely overriding it, the user (perhaps unintentionally) is forgoing any and all functionality that Downshift provides.

People who don't like how downshift works already have an understanding of the functionality downshift provides (that's a natural pre-requisite to not liking it :wink:).

When using `keyDownHandlers` you _will_ quite possibly duplicate logic in your own use case, but I don't see that as a problem. I see the `keyDownHandlers` use cases as very limited in scope, but ultimately quite powerful at the same time.

The fact is that short of creating a special prop for this use case specifically, there's no other way to accomplish this. And I'd much rather solve this problem using a more generally useful prop than a one-off prop and adding a bunch of conditional logic in the code. This attitude is also what lead to the `stateReducer` prop which provides a great deal more value than solving only the original problem it was intended to solve.

> I guess at this point I'm curious when it benefits the developer to use keyDownHandlers over a solution like adding onKeyDown to getInputProps.

To be brief, you'd use `keyDownHandlers` when you want to prevent the default behavior of the built-in keyDown handlers for a specific key, but keep the existing behavior for all other keys. You would also use it if you want to override the existing behavior of the keyDown handlers without calling `event.preventDefault()`. The use cases for these is pretty narrow, but it's the only way that I like. :)

src/downshift.js
kentcdodds: I totally forgot that we were doing `.call(this.getStateAndHelpers(), event)` and thought we were doing `.call(this, event)`. So your changes make total sense.

I'm thinking that I'd like to change the implementation slightly though. Could we do this instead?

```javascript
handler(event, this.getStateAndHelpers())
```

It doesn't resemble a true event handler anymore (because those only receive a single argument), but I think it makes more sense than using `this`. People could give things a much better name:

```javascript
const keyDownHandlers = {
  Enter(event, downshift) {
    if (downshift.isOpen) {
      // do something
    }
  }
}
```

What do you think?

README.md
indiesquidge: Just to understand where you're coming from a bit better, you're saying the use-case for this `keyDownHandlers` prop is as follows:

> you'd use `keyDownHandlers` when you want to prevent the default behavior of the built-in keyDown handlers for a specific key, but keep the existing behavior for all other keys. You would also use it if you want to override the existing behavior of the keyDown handlers without calling `event.preventDefault()`.

To break that up, let me look at the cases one at a time.

> to prevent the default behavior of the built-in keyDown handlers for a specific key, but keep the existing behavior for all other keys.

To consider an alternative, this can also be done already with `onKeyDown`.

```jsx
<input 
  {...getInputProps({
    onKeyDown: e => {
      if (e.key === 'Enter') {
        e.preventDefault()
        // prevent custom behavior for ENTER;
        // keep existing behavior for all other keys
      }
    }
  })}
/>
```

> to override the existing behavior of the keyDown handlers without calling `event.preventDefault()`

I would say this makes a stronger point than the first, but I would posit the question _why would you want to do this?_ Up until this point, #330 and #336 were the poster children for why a feature like `keyDownHandlers` might be desirable, but as we established, #330 wouldn't actually benefit from this feature, and #336 was actually just a question on why the default behavior is the way it is for this library (which I too am confused by; why is the "default" for holding the shift key to jump 5 indices?).

Even if we assume there is some narrow use-case for wanting to override the default behavior for all key down handlers, this can still be accomplished with `onKeyDown` by checking the event key.

Based on this, it seems like **anything you can do with `keyDownHandlers` you can already do—without any changes to the lib—using `onKeyDown` in `getInputProps`**.

On top of the functionality already being here, `onKeyDown` is more idiomatic with React (people already know they can provide an input element an `onKeyDown` handler); it requires no changes to the current code in Downshift (thus avoiding more docs to maintain and more custom props for people to understand); and it won't completely override what Downshift does under the hood unless you explicitly tell it to with `event.preventDefault()`.

indiesquidge: > People who don't like how downshift works already have an understanding of the functionality downshift provides

Why are we assuming that people who want to do something custom don't like what Downshift does already? What if they merely want a say in what happens on a type of event, but still want Downshift to do its own thing with the event as well? #330 is a perfect example of this.

It's just that the more I look at this feature, the less I'm convinced by any problem it is solving for people that they didn't have the ability to solve using `onKeyDown`. In fact, it _adds_ a layer of complexity because the developer must decide if what they really want is to completely override existing functionality or just add custom behavior on top of what exists.

If I'm fighting a losing battle and this feature is desirable no matter what, that's fine. At that point I would just strongly urge the docs to make a point that this is probably not what you want to use if you simply want to do something "extra" with a key down event. The only time you'd want this feature is if you want to _completely override_ Downshift's functionality to handle key down events.

kentcdodds: I think you're still misunderstanding things a bit...

> To consider an alternative, this can also be done already with onKeyDown.

Your example works, however doing that calls `event.preventDefault()` which will prevent the enter key from submitting the form, which is why #330 was brought up in the first place.

> as we established, #330 wouldn't actually benefit from this feature

That is not correct. I had attempted to establish that #330 would actually benefit from this feature.

The workaround for #330 without this feature is to re-implement the entire `onKeyDown` prop and override it entirely:

```javascript
<input
  {...getInputProps()}
  onKeyDown={customKeyDownThatImplementsNintyPercentOfDownshiftLogic}
/>
```

As the name implies, that custom key down handler will have to reimplement a bunch of stuff built into downshift just so it doesn't have to use `event.preventDefault()`. Specifically, what it'd have to duplicate is [this code](https://github.com/paypal/downshift/blob/5278c2c3c437e4861e7180d8a98160caff8872f7/src/downshift.js#L487-L517):

https://github.com/paypal/downshift/blob/5278c2c3c437e4861e7180d8a98160caff8872f7/src/downshift.js#L487-L517

They'd also have to reimplement `moveHighlightedIndex` (which I think we should probably expose in `getStateAndHelpers` anyway, so maybe not).

All that said, I'm starting to feel you have some good points. If we expose `moveHighlightedIndex`, then it's actually not a ton of code to duplicate. So I think I'd be fine with either. I'm not the one who has the use case though so I'd like to hear from folks who actually have this use case.

I think that we can boil down the decision to this:

1. Add `keyDownHandlers` as another way to do the same thing that does it in a slightly different way (which adds complexity to both downshift and its API)
2. Keep things as they are and if someone doesn't like that they have to use `event.preventDefault()` to prevent the downshift behavior, then they simply override things as I have shown above and have to reimplement any of the logic they _do_ want for their implementation. This makes it less ergonomic for people going against the grain.

Whatever we go with, I still think we should expose `moveHighlightedIndex`, so I didn't factor that into the options.

I'm leaning to option 2 now personally...

indiesquidge: > Your example works, however doing that calls `event.preventDefault()` which will prevent the enter key from submitting the form, which is why #330 was brought up in the first place.

I'm not sure I understand what you're saying here. Even without `event.preventDefault()` here in `onKeyDown`, Downshift will stop the form submission on ENTER regardless. AFAICT, _that_ was why #330 was brought up in the first place.

My example was to show that you could use `onKeyDown` "to prevent the default behavior of the built-in keyDown handlers for a specific key, but keep the existing behavior for all other keys", which was one of the points you brought up for `keyDownHandlers` as a prop.

If you wanted to submit the form, you could do that here as well.

```jsx
<input
  {...getInputProps({
    onKeyDown: e => {
      if (e.key === 'Enter') {
        this.props.onSubmit(e)
      }
    }
  })}
/>
```

> The workaround for #330 without this feature is to re-implement the entire `onKeyDown` prop and override it entirely

That is not necessarily true. Looking at the above snippet, one is able to submit a form _and_ retain the event handling provided by Downshift; the event will propagate from `onKeyDown` first and then travel on to Downshift's internal `keyDownHandlers`.

The whole benefit of `onKeyDown` (that I think I'm failing to explain properly) is that it lets you intercept the event, do what you want based on the event, and then let it continue on to `keyDownHandlers` if you so choose (otherwise you can stop it in `onKeyDown` by calling `event.preventDefault()`.

kentcdodds: I'm not 100% certain, but I feel like calling the `onSubmit` callback and submitting the form are two different things entirely.

> Also, is there a reason you have defined onKeyDown outside of getInputProps in your example?

The point of my example was to demonstrate how you could override the behavior without calling `event.preventDefault()` which is what you have to do today to make the form's submit event fire when the Enter key is pressed on the input regardless of the open state. _That_ is the reason #330 was opened in the first place. _That_ is the motivation behind this PR.

indiesquidge: > I'm not 100% certain, but I feel like calling the onSubmit callback and submitting the form are two different things entirely.

I'm not certain either, and I would be more inclined to agree in the scope of vanilla JS, but in React, the only way to get a form to submit is to pass it a submission handler, no?

```jsx
```

ericedem: This might be a naive question, but shouldn't the event only be restricted from propagating if the menu is open and there is an item highlighted? Otherwise the focus is still on the input itself, and I would expect an event to propagate like normal:

```
Enter(event) {
    if (this.getState().isOpen
      && this.getState().highlightedIndex) {
      event.preventDefault()
      this.selectHighlightedItem({
        type: Downshift.stateChangeTypes.keyDownEnter,
      })
    }
},
```

indiesquidge: > I'm not 100% certain, but I feel like calling the onSubmit callback and submitting the form are two different things entirely.

Yeah, I'm not certain either, but unless I'm mistaken, in React the only way to get a form to submit is to pass a submission handler, either to form itself or to a submit button within the form.

If there is a difference between invoking a callback on form submission and just "submitting the form", I'd be curious to know what it is.

indiesquidge: > The point of my example was to demonstrate how you could override the behavior without calling `event.preventDefault()` which is what you have to do today to make the form's submit event fire when the Enter key is pressed on the input regardless of the open state.

Hmm, yeah this makes sense to me know. I hadn't considered if a user was only providing an `onSubmit` handler to the form element itself. I guess it just felt strange to me to do this and not also invoke the same `onSubmit` handler in the `onKeyDown` event for ENTER.

This is how we're currently doing it

```jsx
<form onSubmit={this.handleSubmit}> // invoked if the user clicks the "Search" button
  <Downshift
    render={({ getInputProps, highlightedIndex }) => (
      <input
        {...getInputProps({
          onKeyDown: e => {
            if (e.key === 'Enter' && highlightedIndex) {
              this.handleSubmit(e) // invoked on ENTER key down
            }
          }
        })}
      />
    )}
  />
  <button type='submit'>Search</button>
</form>
```

ericedem: So in our application we have two different types of Downshift autosuggest inputs:

1. Autosuggest values, and force the user to make a selection before doing any kind of submit action.
2. Autosuggest values, but allow the user to submit a "custom" value that isn't selected from the list.

We also have deeply nested components that have abstractions for handling aria compliance, error handling etc., so we have been relying on normal event propagation to handle for submission. The eventual `onSubmit()` handler gathers up all the values from this and potentially other inputs and submits everything.

The #1 style form is easily supported by downshift, but the #2 style is hindered by the `event.preventDefault()`.

As another option, would it make sense to support a prop like `allowCustomValues` that would enable the behavior I mentioned above? This would allow both of these styles of forms.

I could also be wildly oversimplifying things, I am not familiar with all use cases for Downshift :)

donysukardi: @ericedem you can refer to the Type Ahead example https://codesandbox.io/s/82m2px40q9 or https://codesandbox.io/s/1k3m67koj for your second use case.

hzhu: Hi Kent, sorry for the late response! I've been caught up with some work so I'm writing this at 12am 😴 . Thanks to everyone for contributing to a constructive conversation. I really appreciate it. I think everyone made some solid points in this thread. I would like to add my two cents after having read these conversations. 

My teammate and I have implemented a "workaround" solution addressing #330 in our autocomplete. Our workaround was the one @indiesquidge posted. He's actually my teammate and we'll both be at Kent's March testing workshops 😸 we are **excited**! For the sake of reference for all readers, here is our "workaround":

```jsx
<form onSubmit={this.handleSubmit}> {/* invoked on "Search" button click */}
  <Downshift
    render={({ getInputProps, highlightedIndex }) => (
      <input
        {...getInputProps({
          onKeyDown: e => {
            if (e.key === 'Enter' && highlightedIndex === null) {
              this.handleSubmit(e) // invoked on ENTER key down
            }
          }
        })}
      />
    )}
  />
  <button>Search</button>
</form>
```

I am now leaning towards **not using the keyDownEvents prop to solve issue #330**. The following statement made me reconsider:

>⚠ *This is an advanced feature and you should avoid using it* ⚠
>This feature exposes the downshift instance in a way that could lead to your code breaking if we decide to change a method you are relying on.

I would prefer the "workaround" that I have. I would have to do some logical checks to get a `<form>` to call the `onSubmit` handler on `ENTER`, but at least it is guaranteed to be safe. As stated above, I should probably avoid relying an exposed instance property. The current workaround is less code and more safe (e.g. it won't break when someone decides to rename `selectHighlightedItem` in downshift). 

In my use case, I'm only trying to invoke a the form's `onSubmit` handler. Like Kent said, I think this is completely different than a form submission. 

If we want to support narrow use cases to *completely* override Downshift's keydown handlers, then we could continue with this pull request. Otherwise, my personal vote is for option 2. Kent, if you're leaning towards option 2, then I can close this pull request.

> 2) Keep things as they are and if someone doesn't like that they have to use event.preventDefault() to prevent the downshift behavior, then they simply override things as I have shown above and have to reimplement any of the logic they do want for their implementation. This makes it less ergonomic for people going against the grain.

austintackaberry: Not sure if anyone still wants opinions on this, but I like @hzhu's implementation. Per the original issue #330, the main problem is it should be possible to have the form submission handler be called when Enter event is fired:

"I think that a <form> submission handler should always be called on ENTER, even if the combo box list is open."

Notice that the issue DIDN'T say "I think you should be able to use the default functionality of Enter within a form"

I don't think @hzhu actually wanted to use the default functionality when the form is submitted. I think he just wanted to invoke onSubmit which would look something like this:

```jsx

submitHandler(event) {
   // Do something with form contents
}

const handlers = {
   Enter(event) {
    // NO event.preventDefault()
  },
}

const ui = 
   <form onSubmit={submitHandler}>
      <Downshift keyDownHandlers={handlers} />
      <input /> // non-downshift input
      <input /> // non-downshift input
   </form>
```
If that is the case, then you certainly could just handle the "submit" in the custom Enter handler like this:

```jsx
const handlers = {
   Enter(event) {
    event.preventDefault()
    // Do something with form contents
  },
}

const ui = 
   <form>
      <Downshift keyDownHandlers={handlers} />
      <input /> // non-downshift input
      <input /> // non-downshift input
   </form>
```
But then why do this in comparison to @indiesquidge solution above? In my opinion, the answer to this question is because then you can grab the form data from the event if that is what you want (though from what I have read this is not recommended with react). You wouldn't have to have onChange handlers to grab the state of the non-downshift inputs nor would you need to use refs. You could just grab the values in your Enter event handler.

ericedem: @donysukardi Sorry, that doesn't address the issue, since not do an [implicit submission](https://www.w3.org/TR/html52/sec-forms.html#implicit-submission) on enter is the core problem here.

I think the workaround passing a submit handler into a an input works ok for simple use cases, but the nice thing about using the standard eventing structure is when you have multiple abstraction layers.

If I had a form using input components with two levels of abstraction it would need to look like this:

```
<form onSubmit={submitHandler}>
  <CustomInput onSubmit={submitHandler}/>
</form>

function CustomInput ({ submitHandler }) {
  // do some other things
  return <AutocompleteInput onSubmit={submitHandler} />
}

function AutocompleteInput ({ submitHandler }) {
  return (
    <Downshift>
      {({ getInputProps, highlightedIndex }) => (
      <input
        {...getInputProps({
          onKeyDown: e => {
            if (e.key === 'Enter' && highlightedIndex === null) {
              submitHandler(e) // invoked on ENTER key down
            }
          }
        })}
    </Downshift>
  )}
}
```

This, has the result of sprinkling knowledge of the form submission through several places. With events, the input doesn't need to know anything about what it is used for.

kentcdodds: @ericedem is correct. The implicit submission is an important thing we want to allow.

What I'm starting to think is that we need a downshift 2.0.0 that changes the way you prevent the built-in behavior of downshift. It should be a fairly trivial upgrade for people. I'm going to file a new issue for this so we can discuss it, but I think that would resolve the issue in #330.

kentcdodds:
I'm going to go ahead and close this in favor of: https://github.com/paypal/downshift/issues/353

I don't think this feature will be necessary when that's implemented :+1: (any takers?)

Issue #350, opened 2018-02-28
Selecting text in the input can lead to reset()


jduthon:
* `downshift` version: 1.27.1
* `node` version: 8
* `npm` version: 5.5.1
(But I believe all versions are concerned)

When clicking to select the text from the input that `downshift` controls, if you end the selection outside of the input it can end up resetting the input to the `selectedItem` value, while the user is actually trying to interact with the input and select the full text to replace it.

It is reproducible on the simplest version of downshift from the examples in this repo : https://codesandbox.io/s/6z67jvklw3

Steps to reproduce are the following : 
1. Type and select `banana`
2. Change `banana`to `bana`
3. Click on the input, slightly to the right of `bana` (so not directly on the text)
4. Select the full text, and end the text selection outside of input
The input is reset to `banana`, whereas you were trying to edit it.

(If this description is not perfectly clear, which I assume could be the case, I may do a video to demonstrate this, feel free to ask)

The problem clearly comes from https://github.com/paypal/downshift/blob/master/src/downshift.js#L832 

Suggested solution:
I believe a check could be added to know if the `mouseDown` event happened on the input, if it's the case downshift should not reset.
I tend to believe though that this reset on outerClick is not necessarily useful, as `downshift` is already resetting `onBlur` and I cannot see any use case right now where `outerClick` will happen, but not `onBlur`, except the one from this bug.

austintackaberry:
When you say 

> 3. Click on the input, slightly to the right of banana (so not directly on the text)

At this point, shouldn't the input value be "bana" not "banana"? Also, when I click on the input slightly to the right of "bana", nothing happens. I only see the behavior you describe onBlur.

jduthon:
@austintackaberry yes absolutely correct, sorry. I will edit

austintackaberry:
I was able to reproduce this issue with your version and the latest version of downshift. As you mentioned, this is clearly a bug.

I like your suggested solutions; however, it looks to me like the onBlur is a button prop not an input prop:
https://github.com/paypal/downshift/blob/1dec961ad5e787ad29556c85afe8ddc8bb05968f/src/downshift.js#L543

jduthon:
Good (kind of haha).
Well let's wait a bit to have an opinion on how to fix it better.
I wouldn't mind doing it as soon as we know which way to go.

austintackaberry:
[...]

PR #354, opened 2018-03-02
Fix incorrect input reset

jduthon:
**What**:
Bug with selecting the input text which might lead to reseting the state

**How**:
I added a `onFocus` on the input, which is responsible with the `onBlur` to keep a `isInputFocused` boolean updated.
If the input is focused, we do not want to reset the state and fire the `onOuterClick` prop.
It seemed to me that was a clearer way to go than what was discussed in [the issue](https://github.com/paypal/downshift/issues/350).
Though if you consider it not the right way, feel free to ask for edit.

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged


austintackaberry:
I like this solution better as well! :+1:

src/downshift.js
```diff
@@ -827,7 +840,8 @@ class Downshift extends Component {
         if (
           (event.target === this._rootNode ||
             !this._rootNode.contains(event.target)) &&
-          this.getState().isOpen
+          this.getState().isOpen &&
+          !this.isInputFocused
```
kentcdodds: What if instead we just change this to:

```javascript
// above:
const {document} = this.props.environment
// this line
(this.inputId && document.getElementById(this.inputId) === document.activeElement)
```

Then I think we could remove all the focus logic.

Want to give that a try? The tests shouldn't need to change if it works.

jduthon: I like it, and looks like it's working.
It is breaking the test though as it seems like enzyme doesn't really support the `document.activeElement` https://github.com/airbnb/enzyme/issues/1128
I believe I can just go ahead and set the `document.activeElement` in my test.
And what if instead of looking for the `getElementById` we would do as simple text check : 
`(!this.inputId || this.inputId !== document.activeElement.id)` ?

jduthon: Well seems like the testing might get a bit hairier with the `document.activeElement` solution.
To use `document.activeElement` from what I read we might need to use `JSDOM` to simulate this.

kentcdodds: > And what if instead of looking for the getElementById we would do as simple text check:
 `(!this.inputId || this.inputId !== document.activeElement.id)`

That's fantastic.

Yeah, go ahead and set `document.activeElement` manually in your tests. Kinda weird, but should be ok. If you want you could try to add a browser test via cypress:

https://github.com/paypal/downshift/blob/1dec961ad5e787ad29556c85afe8ddc8bb05968f/cypress/integration/basic.js

Learn more about cypress [here](https://cypress.io).

I think we should have both, one for the quick feedback loop and the other to ensure it works without jsdom

kentcdodds: If you can't get it to work with the jest tests then that's fine. Add an [`/* istanbul ignore next */` ](https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md) comment on the line above this condition and it should prevent coverage being recorded for that line and we'll cover it using cypress :+1:

jduthon: Well I went with the latest and removed the jest test.
Code coverage is still 100% as this line is in a condition that I believe is considered executed.
I have only heard about cypress but will try to go ahead and make the test

kentcdodds: Awesome :+1:

Let me know if you need a hand :)

kentcdodds: Oh, and to start up cypress, run `npm run test:cypress:dev`. I should probably add that to the contributing docs... Would you like to?

jduthon: Hmmm not so sure I have the time atm, but I might do it later if you want.
I believe the `cypress` test is up, it's pretty cool and quite straightforward.
I tested with `npm run test:cypress` though, hope it's equivalent.

jduthon:
You are welcome.
Thanks a lot for `downshift`, it's pretty sweet!

kentcdodds:
Actually, now I'm starting to wonder if we should change this to be more generic... How about this:

```javascript
this._rootNode.contains(document.activeElement)
```

This way if the active element is the button or really anything else within the component it wont reset. I'm not 100% certain this is what we're looking for, but it's something to think about and maybe try... Would you like to give it a try?

kentcdodds:

jduthon:
@kentcdodds what would be the advantage to make this more generic.
I can only really think of the case where the input is focused where our scenario can happen (a mouse up outside of the element should not reset). Not so sure what would happen in the case of the button, I am not really familiar with the button globally though.
Also we might introduce unwanted behaviour if we were to do this?
And performance wise the current implementation is faster than the `contains` would be, but well, this is a very shallow point and if there is any reason for the more generic version it would not make any sense to hold this.

Let me know if you want to go for the `contains` version, I believe it should work for sure.
And thanks a lot for adding me to the contributors! :)

austintackaberry:
Using [this](https://codesandbox.io/s/9rWEmzEv8) as an example...

Let's say the user types out 'Luke Skywalker' and does a `mouseDown` on the menu toggle button. Then they change their mind and does a `mouseUp` outside of downshift.

Do we want that to reset? I would say no which would suggest that the `contains` version is the way to go.

kentcdodds:
The thing is that downshift has absolutely 0 opinions about what you render within it. So some implementation could add a second input that serves to further filter the list of items. We wouldn't want to have this same bug on that second input. That's just one example. The use cases are wide and varied. Also, I think that this change would make the solution more robust :)

kentcdodds:
> Separately though, what if the user clicks on the label? I would think the ideal behavior would be the same as onOuterClick.

That's a fair point.... Hmmm.... I'm not certain what to do here. @austintackaberry, would you mind opening a new issue that outlines the problem and arguments for and against for us to discuss this further? That way we can get more people to weigh in.

austintackaberry:
Not at all!

src/downshift.js
yfchina143: hi guys! how did u set up the document.activeelement in enzyme? im using activeElement in my component code. However, in my enzyme testing, it won't recognize the activeelement. i kept getting undefined.

kentcdodds: If you read the thread you commented on you'll see that we didn't do it either....

Issue #353, opened 2018-03-02
Change how to prevent default events


kentcdodds:
Right now if you add a custom event handler (like `onKeyDown` to the `getInputProps` for example), you can prevent the default behavior with `event.preventDefault()`. This has worked pretty well, but honestly it's overloading the feature of preventing the default from events so we should come up with another way to signal preventing the default behavior.

What if instead you simply add a property to the event like: `event.preventDownshiftDefault = true` and then downshift references that rather than `event.defaultPrevented`?

I think that'd be pretty simple and straightforward and the upgrade for this breaking change would be pretty smooth. In fact, we could support both for a little while without much trouble.

Thoughts?

ericedem:
Ok so just to clarify the particular case mentioned in #330. The new workaround will be:

```
getInputProps({
  onKeyDown(event) {
    if (event.key === 'Enter') {
      // If nothing is highlighted, let the normal input event through.
      if (!isOpen || highlightedIndex === null) {
        event.preventDownshiftDefault = true;
      }
    }
  }
})
```

Which overrides the default behavior of calling `event.preventDefault()` when there is no item highlighted, but otherwise let Downshift behave normally.

kentcdodds:
Thanks for the example @ericedem. That's exactly what I would do :+1:

ericedem:
If there is a consensus, happy to take this one on.

kentcdodds:
So I want this to be implemented. I'd be happy to implement it myself, but I'm happy to accept a PR from anyone who gets to it before me.

For this first iteration we should implement it as a non-breaking change and support both this and the current mechanism. It shouldn't be too much work. Probably just documentation, tests, and an update in the utils file for `composeEventHandlers`:

https://github.com/paypal/downshift/blob/10d2cf6128bc008bd7984d537edd812dd2e4306e/src/utils.js#L132-L146

Should probably just be:

```js
// TODO: remove everything after the || in the next breaking change
return event.preventDownshiftDefault || event.defaultPrevented
```

kentcdodds:
Perfect @ericedem :+1: Thanks!

ericedem:
@kentcdodds I have an implementation working for this, where do you recommend putting docs. Under [getInputProps](https://github.com/paypal/downshift#getinputprops) perhaps? Or maybe a new section related to handlers?

kentcdodds:
Let's add a new section for it. It applies to several prop getters

kimgronqvist:
[...]

PR #358, opened 2018-03-05
 feat(handlers): prevent default behavior with preventDownshiftDefault

ericedem:
**What**: allow event handlers to prevent default downshift behavior

This is the non-breaking change for #353.

**Why**: Previously, you could only prevent default behavior by calling event.preventDefault(). This worked, but it would also prevent events from triggering other default behaviors like form submission.

**How**: composeEventHandlers() now checks for `event.preventDownshiftDefault` in addition to `event.defaultPrevented`. The check for `event.defaultPrevented` will be removed in the next breaking change.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged


README.md
```diff
@@ -660,6 +661,25 @@ These are values that represent the current state of the downshift component.
...
 
+## EventHandlers
+
+You can provide your own input handlers to Downshift which will be called before the default handlers.
+
+If you would like to prevent the default behavior, you can set the event's `preventDownshiftDefault` property to `false`.
+
+For example:
+
+```
```
kentcdodds: Let's add `javascript` here so we get syntax highlighting.

Also, could we add an example that shows that you can completely override the built-in prop by adding the prop after spreading the prop getter?

src/__tests__/utils.compose-event-handlers.js
```diff
@@ -0,0 +1,45 @@
+import {composeEventHandlers} from '../utils'
+
+test('prevent default handlers when defaultPrevented is true', () => {
+  const handler = composeEventHandlers(
+    e => {
+      e.defaultPrevented = true
+    },
+    () => {
```
kentcdodds: Rather than this, could we use Jest mock functions (via `jest.fn()`) and assertions? Could make the `defaultCalled` cleaner as well.

README.md
```diff
@@ -660,6 +661,36 @@ These are values that represent the current state of the downshift component.
...
+
+```javascript
+let inputProps = {
+  ...getInputProps(customProps),
+  onKeyDown(event) {
+    // Implement my own behavior...
+  },
+}
+```
+
```
kentcdodds: Cool, thanks for the update. Could we make these examples more practical so folks who may be more beginner will understand. So like:

```javascript
<input {...getInputProps()} onKeyDown={event => { /* my own stuff */} />
```

And the same for the example above?

src/utils.js
```diff
@@ -130,18 +130,20 @@ function debounce(fn, time) {
...
 /**
- * This is intended to be used to compose event handlers
- * They are executed in order until one of them calls
- * `event.preventDefault()`. Not sure this is the best
- * way to do this, but it seems legit...
+ * This is intended to be used to compose event handlers.
+ * They are executed in order until one of them sets
+ * `event.preventDownshiftDefault = true`. This allows the
+ * normal event to maintain its intended `event.defaultPrevented`
+ * state.
```
kentcdodds: The last sentence of this comment is unnecessary.

README.md
```diff
@@ -660,6 +661,38 @@ These are values that represent the current state of the downshift component.
 As a convenience, the `id` and `itemToString` props which you pass to
 `<Downshift />` are available here as well.
 
+## EventHandlers
+
+You can provide your own event handlers to Downshift which will be called before the default handlers.
+
```
kentcdodds: Maybe we should add a simple example here:

```javascript
<input {...getInputProps({onKeyDown: event => /* your handler code */ })} />
```

README.md
```diff
@@ -660,6 +661,38 @@ These are values that represent the current state of the downshift component.
...
+If you would like to prevent the default handler behavior in some cases, you can set the event's `preventDownshiftDefault` property to `false`:
+
+```javascript
+<input
+  {...getInputProps({
+    onKeyDown(event) {
```
kentcdodds: Let's change this to an arrow function just in case someone new copy/pastes this and tries to use `this` and doesn't understand why it's not working.

src/__tests__/utils.compose-event-handlers.js
```diff
@@ -0,0 +1,38 @@
...
+  const composedHandler = composeEventHandlers(customHandler, defaultHandler)
+
+  composedHandler({})
+  expect(customHandler).toHaveBeenCalledTimes(1)
+  expect(defaultHandler).toHaveBeenCalledTimes(0)
+})
+
+test('call default handler when defaultDownshiftPrevented and defaultPrevented are false', () => {
+  const customHandler = jest.fn(() => {})
+  const defaultHandler = jest.fn(() => {})
```
kentcdodds: If you don't care about the implementation you can just do: `jest.fn()` and that's the same as what you have here.

Other than that, these tests are awesome :+1:

ericedem:
> Thanks so much for iterating with me on this so well.

Thanks for the feedback! Better to get this done right. :)

kentcdodds:

PR #359, opened 2018-03-05
[Typescript] Fix return type for itemToString under Actions interface

franklixuefei:
**What**:
[Typescript] Fixed the return type for itemToString under Actions interface

**Why**:
Types for itemToString() between DownshiftProps and Actions are not consistent.

**How**:
Changed the return type from `void` to `string`

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [ ] Ready to be merged
      
### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -146,7 +146,7 @@ export interface Actions {
   clearSelection: (cb?: Function) => void
   clearItems: () => void
   reset: (otherStateToSet?: object, cb?: Function) => void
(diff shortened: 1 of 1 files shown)


kentcdodds:
Thanks a bunch for the help!

Antontelesh:
An interesting thing here: we have `itemToString` under `Actions`, although it’s not an action. It just happened to be a function.

I believe, next time we do something related to typings, we can refactor it. It’s not something causing a bug, but it’d be better to fix it in the future.

Issue #334, opened 2018-02-19
[react-native] Button props - Cannot read property activeElement of undefined


eyalcohen4:
* `downshift` version: 1.28.0
* `node` version: 8.5.0
* `npm` (or `yarn`) version: 5.6.0
* `react-native` version: 0.52.0
* `react` version: 16.2.0

Relevant code or config

```javascript
```

What you did:
Hey! tried to use the new and experimental react-native version (Thanks for that!), With Button instead of TextInput as the trigger.

What happened:
`Cannot read property activeElement of undefined` error has thrown.
![screen shot 2018-02-19 at 15 59 24](https://user-images.githubusercontent.com/16276358/36381382-0ce7cf36-158e-11e8-97df-cde6608666fa.png)

Reproduction repository:
https://github.com/eyalcohen4/downshift-react-native-bug

Problem description:
I think its related to [the access attempt for `document` property of `environment` object, while it's missing on react-native when I print `environment`](https://github.com/paypal/downshift/blob/master/src/downshift.js#L566). 

Suggested solution:
Check if we are in react-native env using `isReactNative` before [this](https://github.com/paypal/downshift/blob/master/src/downshift.js#L566) condition - 
`
    if (
      !isReactNative() && 
      this.props.environment.document.activeElement ===
      this.props.environment.document.body
    ) {
      event.target.focus()
    }
`

Will be happy to make a PR if it looks like the right solution. 
cc @kentcdodds @eliperkins

kentcdodds:
That solution sounds good to me! What do you think @eliperkins?

eliperkins:
Hm, I’m not too familiar with what the use-case would be here, but my knee jerk reaction would be that we should still attempt to call `focus` on the input field, which we should be able to do if there’s a ref for it. But putting the check there for now seems like a nice way to degrade gracefully, rather than throw an error!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

eyalcohen4:
In my particular example there is only a button -
It's the use case of a drop-down without Auto complete.

maybe we should refocus the button (which is what I think line 569 trying
to achive) in native environment - I'll reaserch for a way to implement
this.

Thanks!

kentcdodds:
Hmmm.... The intention of the existing code is to refocus on the target of the event (which is normally the `input` but could also be a button in the case of a drop-down. I'm not sure how focus works in native mobile apps, but if you do have to manage focus, then we should probably do it in the same way.

[...]

PR #368, opened 2018-03-10
fix(native): focus input on button clicked (#334)

eyalcohen4:
**What**:
Prevent error in button_handleClick, and refocus `TextInput` if exists - both only in react-native environment.

**Why**:
#334 
In the context of react-native, when we use a button instead of text input, `Cannot read property activeElement of undefined error` has thrown. The cause was a check for document object in native environment. Also, We wanted to re-focus the `TextInput` element if one exists. 

**How**:

 in order to refocus the `TextInput` after a button click, `refKey` added to `getInputProps`, and  assigned only in react-native using `inputNode` method, as in `rootNode` and `getRootProps`. 
For disabling focusing the event target in react-native, I simply added `!preval`module.exports = process.env.BUILD_REACT_NATIVE === 'true'``.

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [x] Ready to be merged
      
The `refKey` props still need documentation and tests for react-native.

Thanks!

### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -564,6 +564,7 @@ class Downshift extends Component {
     event.preventDefault()
 
     if (
+      /* istanbul ignore next (react-native) */
       preval`module.exports = process.env.BUILD_REACT_NATIVE === 'true'` &&
       this._inputNode
     ) {
@@ -574,6 +575,7 @@ class Downshift extends Component {
     // don't give the button the focus properly.
     /* istanbul ignore if (can't reasonably test this) */
     if (
+      /* istanbul ignore next (react-native) */
       !preval`module.exports = process.env.BUILD_REACT_NATIVE === 'true'` &&
       this.props.environment.document.activeElement ===
         this.props.environment.document.body
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -562,15 +562,25 @@ class Downshift extends Component {
 
   button_handleClick = event => {
     event.preventDefault()
+
+    if (
+      preval`module.exports = process.env.BUILD_REACT_NATIVE === 'true'` &&
+      this._inputNode
+    ) {
+      this._inputNode.focus()
+    }
+
     // handle odd case for Safari and Firefox which
     // don't give the button the focus properly.
     /* istanbul ignore if (can't reasonably test this) */
     if (
+      !preval`module.exports = process.env.BUILD_REACT_NATIVE === 'true'` &&
       this.props.environment.document.activeElement ===
-      this.props.environment.document.body
+        this.props.environment.document.body
     ) {
       event.target.focus()
     }
+
     this.toggleMenu({type: Downshift.stateChangeTypes.clickButton})
   }
 
@@ -610,7 +620,16 @@ class Downshift extends Component {
 
   /////////////////////////////// INPUT
 
-  getInputProps = ({onKeyDown, onBlur, onChange, onInput, ...rest} = {}) => {
+  inputRef = node => (this._inputNode = node)
+
+  getInputProps = ({
+    onKeyDown,
+    onBlur,
+    onChange,
+    onInput,
+    refKey = 'ref',
+    ...rest
+  } = {}) => {
     this.getInputProps.called = true
     if (this.getLabelProps.called && rest.id && rest.id !== this.inputId) {
       throw new Error(
@@ -659,6 +678,9 @@ class Downshift extends Component {
       ...eventHandlers,
       ...rest,
       id: this.inputId,
+      [refKey]: preval`module.exports = process.env.BUILD_REACT_NATIVE === 'true'`
+        ? this.inputRef
+        : null,
     }
   }
 


src/downshift.js
```diff
@@ -562,15 +562,27 @@ class Downshift extends Component {
 
   button_handleClick = event => {
     event.preventDefault()
+
+    if (
+      /* istanbul ignore next (react-native) */
```
eliperkins: I think you may need to lift this `instanbul` declarative one line higher to ensure it's ignored for now.

src/downshift.js
```diff
@@ -610,7 +622,16 @@ class Downshift extends Component {
 
   /////////////////////////////// INPUT
 
-  getInputProps = ({onKeyDown, onBlur, onChange, onInput, ...rest} = {}) => {
+  inputRef = node => (this._inputNode = node)
```
eliperkins: From the Travis CI failure, it looks like this should be ignored for now too.

eliperkins:
@kentcdodds can you give this a quick look over? I'm starting to get hesitant about sprinkling all these `istanbul ignore`'s all over.

src/downshift.js
```diff
@@ -659,6 +680,9 @@ class Downshift extends Component {
       ...eventHandlers,
       ...rest,
       id: this.inputId,
+      [refKey]: preval`module.exports = process.env.BUILD_REACT_NATIVE === 'true'`
+        ? this.inputRef
+        : null,
```
kentcdodds: I'd say let's remove the condition here and just always apply the `ref` here.

But we need to make sure we call both the given `ref` as well as our own, so could you do this?

1. Add a new function to `utils` called `callAll`:

```javascript
/**
 * Function that returns a function which calls all given functions
 * with the given arguments arguments
 */
function callAll(...fns) {
  return (...args) =>
    fns.filter(Boolean).forEach(fn => {
      if (fn) {
        fn(...args)
      }
    })
}
```

2. Add that function to the imports at the top, then use it here:

```javascript
[refKey]: callAll(this.inputRef, rest[refKey])
```

3. Bonus points: do the same thing to the `[refKey]: ...` above in `getRootProps` :+1:

eyalcohen4:
Thanks for the feedback @eliperkins @kentcdodds!! Will add a test and replace the conditional ref both on the input and root. hope to do so soon

Great weekend

eyalcohen4:
I'm having a hard time with the test, after applying the ref's as Kent suggests and testing manually - hope one of you can guide me through the right way or give some tips:

I want to validate that the TextInput is focused on the button press. 

I mocked the `event` object which passed by react-native to the handle method of onPress, 
But when trying to access the inputRef from the Input props he's returning `undefined` instead of the ref, although when running in react-native environment (both on the simulator and real-device) he's holding it.

Maybe you know how one can simulate the environment in the tests to have inputRef? I wasn't able to find such way yet but will research more.  Maybe it's not even the right direction?
 
Let me know what you think.
```js
  test('foucs input node in case of button press', () => {
  const RootView = ({innerRef, ...rest}) => <View ref={innerRef} {...rest} />
  const ButtonMock = jest.fn(props => <Button {...props} />)
  const InputMock = jest.fn(props => <TextInput {...props} />)

  const renderSpy = jest.fn(({getRootProps, getInputProps, getItemProps, getButtonProps}) => (
    <RootView {...getRootProps({refKey: 'innerRef'})}>
      <InputMock {...getInputProps({ refKey: 'inputRef'})} />
      <ButtonMock {...getButtonProps()} title="Downshift"/>
      <View>
        <Text {...getItemProps({item: 'foo', index: 0})}>foo</Text>
        <Text {...getItemProps({item: 'bar', index: 1})}>bar</Text>
      </View>
    </RootView>
  ));
  const element = <Downshift>{renderSpy}</Downshift>
  TestRenderer.create(element)
  const [[buttonProps]] = ButtonMock.mock.calls;
  const [[inputProps]] = InputMock.mock.calls;

  buttonProps.onPress({ preventDefault: () => {}});

  // here inputRef is a function which returns undefined
  expect(inputProps.inputRef().isFocused()).toBeTruthy();
})
```

andrroy:
@eyalcohen4 Have you had any progress on this PR? Is there anything I can do to help get it ready for merge? 🙂

eyalcohen4:
Thanks @andrroy, I'm having a hard time with the test's here - after applying the ref's as Kent suggests and testing manually.

In this test, I wanted to validate that the TextInput is focused on the button press.

the problem is when trying to access the ref on node (jest) environment his value is `undefined`, although when running in react native environment (both on the simulator and real-device) he's holding it.

Maybe one of you know how we can simulate the environment in the tests to have `inputRef`? I wasn't able to find such way yet, or maybe it's not even the right direction?
 
```js
  test('foucs input node in case of button press', () => {
  const RootView = ({innerRef, ...rest}) => <View ref={innerRef} {...rest} />
  const ButtonMock = jest.fn(props => <Button {...props} />)
  const InputMock = jest.fn(props => <TextInput {...props} />)

  const renderSpy = jest.fn(({getRootProps, getInputProps, getItemProps, getButtonProps}) => (
    <RootView {...getRootProps({refKey: 'innerRef'})}>
      <InputMock {...getInputProps({ refKey: 'inputRef'})} />
      <ButtonMock {...getButtonProps()} title="Downshift"/>
      <View>
        <Text {...getItemProps({item: 'foo', index: 0})}>foo</Text>
        <Text {...getItemProps({item: 'bar', index: 1})}>bar</Text>
      </View>
    </RootView>
  ));
  const element = <Downshift>{renderSpy}</Downshift>
  TestRenderer.create(element)
  const [[buttonProps]] = ButtonMock.mock.calls;
  const [[inputProps]] = InputMock.mock.calls;

  buttonProps.onPress({ preventDefault: () => {}});

  // here inputRef is a function which returns undefined
  expect(inputProps.inputRef().isFocused()).toBeTruthy();
})
```

PR #370, opened 2018-03-12
Add docs for default keydown handlers

indiesquidge:
**What**:

Add documentation for Downshift's internal keydown handlers.

**Why**:

Due to a number of issues being raised and comments being written on the topic of keydown events when using Downshift, it seems like a good idea to document Downshift's implicit keydown handlers. This compliments nicely with the section already written about how to customize or circumvent Downshift's internal event handlers (re: #358)

**How**:

Added a section in the README outlining what each keydown handler does by default.

**Checklist**:

* [x] Documentation
* [ ] ~~Tests~~ N/A
* [x] Ready to be merged

You'll note that I have not yet checked "Ready to be merged". This is due to an outstanding question on the expected behavior of the 'Escape' keydown event (see https://github.com/paypal/downshift/issues/336#issuecomment-372207149)


README.md
```diff
@@ -782,12 +811,12 @@ Examples exist on [codesandbox.io][examples]:
...
-3. Update the title and description
-4. Update the code for your example (add some form of documentation to explain
-   what it is)
-5. Add the tag: `downshift:example`
+1.  Fork [this codesandbox](http://kcd.im/ds-example)
+2.  Make sure your version (under dependencies) is the latest available version.
+3.  Update the title and description
+4.  Update the code for your example (add some form of documentation to explain
+    what it is)
+5.  Add the tag: `downshift:example`
```
indiesquidge: I'm not sure I understand what's causing some of these changes.

Everything other than the chunk added about the default event handlers is being added automatically after I run and complete the `add-contributor` script. I can split the changes into two commits to show this if need be.

Some of these spacing changes are innocuous, but it seems like the automated changes being made to the contributors table are swapping out absolute paths with local paths (e.g. `https://github.com/paypal/downshift/commits?author=kentcdodds "Code"` is replaced with `/paypal/downshift/commits?author=kentcdodds "Code"`, and so on).

If someone has a clue as to why this is happening, I'd like to remove these undesired changes from this pull request.

austintackaberry: You didn't add the spaces manually? That's interesting...

Also, I'm not sure what caused the paths to change from absolute to local, but I have seen this behavior in previous add-contributor commits.

kentcdodds: Yeah, I'm pretty sure that prettier was updated since it was last run on this file or something. I'm not really concerned, but I am a bit confused by that.

I'm pretty sure what caused this [this pr](https://github.com/jfmengels/all-contributors-cli/pull/84). I've released a fix to all-contributors-cli. If you reinstall dependencies then run `./node_modules/.bin/kcd-scripts contributors generate` it should get fixed up.

README.md
```diff
@@ -676,6 +678,33 @@ As a convenience, the `id` and `itemToString` props which you pass to
 
 ## Event Handlers
 
+Downshift has a few events for which it provides implicit handlers. All of these
+handlers prevent the default event action, and a few of them have additional
+functionality.
```
kentcdodds: Could we rephrase this to be something more like:

```
Downshift has a few events for which it provides implicit handlers. Several of these handlers call `event.preventDefault()`. Their additional functionality is described below.
```

README.md
```diff
@@ -676,6 +678,33 @@ As a convenience, the `id` and `itemToString` props which you pass to
...
+
+* `ArrowUp`: moves the highlighted index up by 1. If this shift key is held when
+  this event fires, the highlighted index will jump up 5 indices instead of 1. NOTE:
+  if the current highlighed index is within the top 4 indices, the bottom-most index
+  will be highlighted.)
+
+* `Enter`: if the menu is open, select the currently highlighted item. NOTE: as
+  stated above, the usual 'Enter' event is prevented by Downshift's default implicit
+  enter handler, so, for example, a form submission event will not work as one
+  might expect. See below for customizing the handlers.
```
kentcdodds: This is only true for when the menu is open. If the menu is closed then `event.preventDefault()` will not be called and form submission should work.

README.md
```diff
@@ -676,6 +678,33 @@ As a convenience, the `id` and `itemToString` props which you pass to
...
+
+* `Escape`: will close the menu, set the highlighted index to the
+  `defaultHighlightedIndex`, and "reset" the input value to the previous selected
+  item.
```
kentcdodds: Could we call this more generally "reset" so we can simply say:

```javascript
`Escape`: will reset downshift's state. This means that `highlightedIndex` will be set to the `defaultHighlightedIndex`, the `inputValue` will be set to the `itemToString` value of the `selectedItem`, and the `isOpen` state will be set to `false`.
```

README.md
indiesquidge: Nice! Thank you, @austintackaberry and @kentcdodds. The change made to the executable module worked. Should be good for another review now :)

README.md
```diff
@@ -676,6 +678,33 @@ As a convenience, the `id` and `itemToString` props which you pass to
...
 
+Downshift has a few events for which it provides implicit handlers. Several of
+these handlers call `event.preventDefault()`. Their additional functionality is
+described below.
+
+### default handlers
+
+* `ArrowDown`: moves the highlighted index down by 1. If this shift key is held
+  when this event fires, the highlighted index will jump down 5 indices instead of 1.
+  NOTE: if the current highlighed index is within the bottom 4 indices, the top-most
```
kentcdodds: This isn't quite accurate. For example, if you're on the second to last item and arrow down with the shift key you will be placed at the forth item. It's always a movement of 5 regardless of where you are in the last. You might say it's the equivalent of doing the same action 5 times.

README.md
```diff
@@ -676,6 +678,33 @@ As a convenience, the `id` and `itemToString` props which you pass to
...
+  index will be highlighted.)
+
+* `ArrowUp`: moves the highlighted index up by 1. If this shift key is held when
+  this event fires, the highlighted index will jump up 5 indices instead of 1. NOTE:
+  if the current highlighed index is within the top 4 indices, the bottom-most index
```
kentcdodds: Same here

README.md
```diff
@@ -676,6 +678,33 @@ As a convenience, the `id` and `itemToString` props which you pass to
...
+  will be highlighted.)
+
+* `Enter`: if the menu is open, select the currently highlighted item. If the menu
+  is open, the usual 'Enter' event is prevented by Downshift's default implicit enter
+  handler; so, for example, a form submission event will not work as one might expect.
```
kentcdodds: Please add at the end of this sentence `(though if the menu is closed the form submission will work normally)`

indiesquidge: Oh, I figured I had this caveat covered by changing the start of the sentence to "If the menu is open…". Would you prefer I remove that and add your suggestion to the end?

README.md
indiesquidge: > For example, if you're on the second to last item and arrow down with the shift key you will be placed at the forth item. It's always a movement of 5 regardless of where you are in the last.

Hmm, this isn't the behavior that I observe. I get the same behavior that was the point of confusion in #336, namely, if **any** of the final 5 indices (which I actually _did_ mistype here when I wrote 4) are highlighted and the shift key is held when an ArrowDown event is fired, the top-most index is highlighted.

README.md
kentcdodds: Yeah, let's just make the addition I suggested. I feel like it could be ambiguous otherwise.

README.md
kentcdodds: That's odd... I could have sworn the logic what do as I suggested. Maybe I changed it and forgot? 🤔 Or maybe there's something weird with that example.

kentcdodds: Just checked. Looks like that's the logic alright. I'm ok with that. This is a rarely used feature anyway and I think both behaviors are just as reasonable as the other. This is fine! Thanks!

Issue #23, opened 2017-07-26
Flow typed


souporserious:
How do you feel about adding Flow to this project? Could even add inline comments while doing this so docs can be generated later.

We could even take it a step further and remove PropTypes in favor of just Flow with [babel-plugin-react-flow-props-to-prop-types](https://github.com/thejameskyle/babel-plugin-react-flow-props-to-prop-types).

nitin42:
Can I work on this ? (PR) I've used Flow earlier!

nitin42:
having a type system makes it easy to understand the codebase rather than increasing verbosity @kentcdodds

kentcdodds:
I'm not worries about increased verbosity. I'm worried about familiarity for beginners. But maybe that concern is overblown and it won't be bad. Feel free to submit a pull request. Thanks!

nitin42:
Awesome!!

kentcdodds:
I'll tell you what. If we can add a short intro and links to resources in the CONTRIBUTING.md, then I say let's go for adding flow to the project :+1:

nitin42:
Intro and links about setting up Flow or ?

kentcdodds:
Intro and links about the basics of what Flow is and the syntax. Everything someone would need to know to contribute to a flow codebase for the first time.

kentcdodds:
Could just be a link to a blog post.

nitin42:
Sure! I'll add a section for Flow in the `CONTRIBUTING.md` 👍

vejersele:
Hi All,

I see that the PR for this added a .flow.js file which is awesome, but what about actually adding types to the code itself? It would make code changes a lot safer since you would have static type checking and there would be one type definition less to maintain, since it can be automatically generated?

kentcdodds:
Hey @vejersele! I've definitely considered adding flow to my libraries. My main concern is that it will hinder contributions because folks who are unfamiliar with flow will have a hard time contributing....

ttamminen:
If the source code itself is going to be typed then TypeScript should be also evaluated. I am biased towards TypeScript, so I will stay out of the discussion. Just reminding about the options. 😊

The feature sets on both type systems are more than enough.

What should be evaluated then?

- activity on 3rd party library definition creation [TS](https://www.npmjs.com/~types#packages) & [Flow](https://github.com/flowtype/flow-typed/tree/master/definitions/npm)
- tooling
- which libraries have taken the TypeScript/Flow to the internals

[...]

PR #372, opened 2018-03-13
[#23] Flow definitions

stereobooster:
**What**: added script to generate flow definitions
**Why**: because people use downshift with Flow
**How**: flowgen generates type definitions from TypeScript definitions
**Checklist**:

* [ ] Documentation
* [ ] Tests
* [ ] Ready to be merged


kentcdodds:
Hey thanks! This seems great! Though honestly it doesn't seem like it would be all that hard to maintain them separately considering how small the API surface is 🤔 If these bugs give us a lot of trouble I'd rather do that. Luckily for us your PR has given us a great place to start! Would by be opposed to just doing that instead?

stereobooster:
> Though honestly it doesn't seem like it would be all that hard to maintain them separately

You mean inside https://github.com/flowtype/flow-typed?

Current state of flowtype tooling is sad, it is not optimised for developers workflow. I would prefer to keep definitions here (if this is possible, I also would understand if you do not want it)

> If these bugs give us a lot of trouble

The only issue is this with `kcd-scripts`. Other one I want to fix

kentcdodds:
The `kcd-scripts` error says:

```
[error] typings/index.flow.js: SyntaxError: `declare export interface` is not supported. Use `export interface` instead (9:16)
[error]    7 |  */
[error]    8 |
[error] >  9 | declare export interface DownshiftState {
[error]      |                ^
[error]   10 |     highlightedIndex: number | null,
[error]   11 |         inputValue: string | null,
[error]   12 |         isOpen: boolean,
```

It's not a kcd-scripts error at all, it looks to be a prettier error. We could probably add that file to [the `.prettierignore` file](https://github.com/paypal/downshift/blob/f6e838b924ca8c9c207ca04154d07c11d76dfad1/.prettierignore) and avoid that error (though that means it wont be formatted by prettier, but as it is that's the case anyway).

I wasn't suggesting flow-typed. I was just suggesting that rather than generating the types we maintain them by hand. Considering they're so small I don't know whether it's worth making the tooling work.

If we _do_ want to go with the generated approach, then I'll want the generation of the flow types to happen as part of the precommit hook. I can help make that happen because it takes a little bit of understanding of how kcd-scripts works. But I still think that maintaining them completely separately is preferrable...

stereobooster:
> I wasn't suggesting flow-typed. I was just suggesting that rather than generating the types we maintain them by hand. Considering they're so small I don't know whether it's worth making the tooling work.

Ah I see. I have no preference. As soon as we get through with TypeScript update (https://github.com/paypal/downshift/pull/373) I will update this PR

> It's not a kcd-scripts error at all, it looks to be a prettier error. 

This is strange, I was sure Prettier can eat any flow file. I will double check this

kentcdodds:
Thanks for all your help @stereobooster! You're a rock star!

giphy

stereobooster:
This doesn't work so far. I asked for advice in flow channel, but no luck so far

stereobooster:
closed in favour https://github.com/paypal/downshift/pull/399

PR #373, opened 2018-03-13
Change typings

stereobooster:
**What**: More strict type definitions. Use parametric type `Item` instead of `any`
**Why**: To have more type safety
**How**: Introduced parametric type
**Checklist**:

* [ ] Documentation
* [x] Tests
* [x] Ready to be merged

---

This is not final version, rather a code to be discussed. Current implementation not ideal, it would be nice if typescript could extract type from given data instead of declaring it explicitly

```ts
declare const Downshift: DownshiftInterface<any>
```

but I haven't figured out how to do it properly. As of now I'm using this hack:

```ts
const TypedDownShift: DownshiftInterface<Item> = Downshift
```

UPD: we can use something like https://github.com/piotrwitek/utility-types to extract `Item` type from other information

Also would be nice to type `object` and `Function` in a more proper manner.


Antontelesh:
Is it backwards compatible? Will downshift work if used without explicit type casting?

kentcdodds:
I definitely would like to avoid pushing a major version bump on downshift because typing changes... But if it really improves the experience for people then I'd be willing to do it (in combination with a few other changes I've been wanting to make... so it may take some time.)

stereobooster:
> Is it backwards compatible?

Not quite

Old:

```ts
      <DownShift onChange={this.onChange}>
        {({
          getButtonProps,
          getInputProps,
          getItemProps,
          isOpen,
          inputValue,
          selectedItem,
          highlightedIndex,
        }: ControllerStateAndHelpers) => (
```

New (`<Item>`):

```ts
      <DownShift onChange={this.onChange}>
        {({
          getButtonProps,
          getInputProps,
          getItemProps,
          isOpen,
          inputValue,
          selectedItem,
          highlightedIndex,
        }: ControllerStateAndHelpers<Item>) => (
```

But if you do something like this, you are ok:

```ts
      <DownShift onChange={this.onChange}>
        {({
          getButtonProps,
          getInputProps,
          getItemProps,
          isOpen,
          inputValue,
          selectedItem,
          highlightedIndex,
        }) => (
```

kentcdodds:
Is there any way to make it backward compatible? Even if it's sub-optimal? That'll allow us to release it and then push the breaking changes later.

stereobooster:
> Is there any way to make it backward compatible? Even if it's sub-optimal?

Duplicate everything like this:

```ts
export type ControllerStateAndHelpersParameterized<Item> = DownshiftState<Item> &
  PropGetters<Item> &
  Actions<Item>

export type ControllerStateAndHelpers = ControllerStateAndHelpersParameterized<any>
```

Antontelesh:
It would be pointless to explicitly refer to `ControllerStateAndHelpers` without type argument like you has shown in “Old” example. You already have that type inferred. I believe, nobody does this.

That makes me think these changes are safe enough to be released under minor version.

What do you think about it?

stereobooster:
Just in case it wasn't clear, my example is from dowshift codebase itself (see "Files changed" tab)

Antontelesh:
In that case we shouldn’t encourage that type of usage of TypeScript.

Types are inferred and it’s completely optional to name them everywhere.

But as there might be someone already using that syntax, for them these changes are breaking.

So we have two options: to create duplicates like @stereobooster mentioned. Or to schedule this PR onto the next major version, when some other significant change happens.

I personally prefer the latter.
@kentcdodds, you decide :)

stereobooster:
Any advice on how to fix this hack?

```ts
const TypedDownShift: DownshiftInterface<Item> = Downshift
```

kentcdodds:
I do want to wrap up #285 and a few other things, so maybe I'll take some time next week to put together a major version release. Go ahead and get this PR ready and when I start work on the next major version I'll switch the base of this PR to a `next` branch I'll create and then I'll use that to start making beta releases for people to try out :+1: Thanks!

tsconfig.json
```diff
@@ -1,7 +1,14 @@
 {
   "compilerOptions": {
     "jsx": "react",
-    "noUnusedLocals": true
+    "noUnusedLocals": true,
```
stereobooster: the most strict config possible

typings/index.d.ts
```diff
@@ -114,47 +111,53 @@ interface OptionalExtraGetItemPropsOptions {
...
+  extends OptionalExtraGetItemPropsOptions {
   index?: number
-  item: any
+  item: Item
 }
 
-export interface PropGetters {
+export interface PropGetters<Item> {
   getRootProps: (options: GetRootPropsOptions) => any
   getButtonProps: (options?: GetButtonPropsOptions) => any
```
stereobooster: Tried to add even more types, but it is hard to encode dynamic nature of this function

```ts
type GetItemPropsResult<Element> = {
  id: string,
} & (({
  onMouseMove: (e: React.MouseEvent<Element>) => void,
  onMouseDown: (e: React.MouseEvent<Element>) => void,
} & ({
  onPress: (e: React.MouseEvent<Element>) => void, // should be react native type
} | {
  onClick: (e: React.MouseEvent<Element>) => void,
})) | {})

getItemProps: <Element>({options, ...rest}: {options: GetItemPropsOptions<Item>}) => GetItemPropsResult<Element> & typeof rest
```

typings/index.d.ts
```diff
@@ -114,47 +111,53 @@ interface OptionalExtraGetItemPropsOptions {
...
-    cb?: Function,
+    otherStateToSet?: {},
+    cb?: Callback,
   ) => void
-  clearSelection: (cb?: Function) => void
+  clearSelection: (cb?: Callback) => void
   clearItems: () => void
-  reset: (otherStateToSet?: object, cb?: Function) => void
-  itemToString: (item: any) => string
+  reset: (otherStateToSet?: {}, cb?: Callback) => void
```
stereobooster: Replaced `object` with `{}` and `Function` with `Callback`

typings/index.d.ts
```diff
@@ -24,77 +26,72 @@ export enum StateChangeTypes {
...
 }
 
-export interface Document {
-  getElementById: (id: string) => HTMLElement
-}
-
-export interface A11yStatusMessageOptions {
+export interface A11yStatusMessageOptions<Item> {
   highlightedIndex: number | null
-  highlightedValue: any
```
stereobooster: Removed this, because I haven't found any other mentions of this property

test/basic.test.tsx
```diff
@@ -1,82 +1,75 @@
-import * as React from 'react';
-import Downshift, { ControllerStateAndHelpers, StateChangeOptions } from '../';
+import * as React from 'react'
+import Downshift, { StateChangeOptions, DownshiftInterface } from '../'
+
+type Item = string
+const TypedDownShift: DownshiftInterface<Item> = Downshift
```
stereobooster: Is there a better way to do this

Antontelesh: I tried to check how people solve this usually, and this seems to be the best possible solution as of now.

Antontelesh: I believe it will be great to have an example of simple usage of Downshift, like it was used before, without supplying type arguments. Just to ensure that the old way still works.

typings/index.d.ts
Antontelesh: Does `{}` allow to put a non-empty object to the argument?

stereobooster: Yes https://www.typescriptlang.org/play/#src=const%20i%20%3D%20(x%3A%20%7B%7D)%20%3D%3E%20x%0D%0Ai(%7B%20a%3A%201%20%7D)%0D%0A

test/basic.test.tsx
stereobooster: Add example of simple usage of Downshift in `Custom.tsx`. Or you want to add one more file?

Antontelesh: Great! Thank you! Now we have examples of both typed and untyped versions.

typings/index.d.ts
tansongyang: As long as @kentcdodds is cool with this, we should remember to add this to the changelog.

kentcdodds: Yeah, I think that must have been a mistake. That isn't used anywhere in the codebase.

kentcdodds:
> I did want to mention that there were a lot of whitespace changes that sometimes made it hard to understand the diff

Here's a tip: Add ?w=1 to the URL of a PR and it'll not show whitespace only changes. Like [so](https://github.com/paypal/downshift/pull/373/files?w=1) :+1:

stereobooster:
Isn't whitespaces handled by Prettier? This is my IDE changed whitespaces, but I expected that Prettier will take care about it anyway

kentcdodds:
Yes, prettier formatted these files, so no worries @stereobooster :+1:

kentcdodds:
Though actually I just realized that [`kcd-scripts`](https://github.com/kentcdodds/kcd-scripts) will not have prettier autoformat `tsx` files which is why this actually was _not_ formatted by prettier (and never has been).

If anyone wants to submit a PR to kcd-scripts to have it support both you'd just need to change these here:

- https://github.com/kentcdodds/kcd-scripts/blob/240fcd4dbb04a85de24149c0b304b048772f2e64/src/scripts/format.js#L36
- https://github.com/kentcdodds/kcd-scripts/blob/240fcd4dbb04a85de24149c0b304b048772f2e64/src/config/lintstagedrc.js#L14

Should be pretty straightforward :+1: Then we could upgrade kcd-scripts in this project and run `npx kcd-scripts format` :+1:

stereobooster:
I fixed formatting. If there are no better options for this hack `const TypedDownShift: DownshiftInterface<Item> = Downshift` then it is mergable. Do we need documentation for this?

kentcdodds:
Alrighty! So I'm getting ready to merge this into a new `next` branch. Could you please rebase this PR? Just update the `.all-contributors` file so it's valid json, then run `npx kcd-scripts contributors generate` to update the `README.md`.

stereobooster:
Updated. Do I need to squeeze it?

kentcdodds:
No worries about squashing it, I can do that when I merge it :+1:

stereobooster:
typescript tests pass

kentcdodds:
I'm really sorry. I'm not sure what I did, but this PR now has 2 big merge conflicts. Could you give those a look. I'm about ready to merge this into a new `next` branch based on the current master.

test/custom.test.tsx
```diff
@@ -1,60 +1,64 @@
...
-    };
-
-    render() {
-        const items = this.state.items;
-        const defaultSelectedItem = this.state.items[0];
-
-        return (
-            <Downshift defaultSelectedItem={defaultSelectedItem}>
-                {({
-                    getToggleButtonProps,
```
stereobooster: this was changed in master. Should I change it to master version?

kentcdodds: Yes, sorry, I didn't consider how that would impact your PR. I renamed (and aliased) `getButtonProps` to `getToggleButtonProps`. Sorry about that.

kentcdodds:
Super! I'll merge this into `next` now :+1:

kentcdodds:
If you could keep an eye on the `next` branch that would be great @stereobooster!

Let's make sure the typings are good to go when we're ready to release `next`.

kentcdodds:

Issue #364, opened 2018-03-07
Space keyDown to toggle menu can't be used without resetting input


austintackaberry:
* `downshift` version: v1.29.1

Relevant code or config

https://github.com/paypal/downshift/blob/68499b42eff38474947e55e9ea636a340984608e/src/downshift.js#L681-L685

**What you did and what happened:**

https://codesandbox.io/s/vv3km9z9m5

![undefined](https://user-images.githubusercontent.com/29493001/37124186-0ff2fd1e-221c-11e8-8b84-cf16aed6b5da.gif)

**Problem description:**

Using default behavior, the user cannot toggle the menu using only the keyboard without resetting the input

**Suggested solution:**

Don't reset the input onBlur by default if the activeElement is a downshift button?

austintackaberry:
Thoughts anyone??

kentcdodds:
Thanks for bringing this up @austintackaberry. I like your suggested solution :+1:

Anyone can feel free to open a PR to fix this behavior. Please include tests! (Perhaps a Cypress test would be best for this one).

austintackaberry:
Thanks, I'll give it a try!

kentcdodds:
This was resolved :+1: thanks @austintackaberry

PR #374, opened 2018-03-14
WIP: test coverage issues with onBlur

austintackaberry:
**What**: Allow user to tab from input to button without resetting, and allow user to shift+tab from button to input without resetting. See https://github.com/paypal/downshift/issues/364 for more details.

**Why**: If user wanted to only use keyboard to navigate between input and toggle button, the current usage would be a nuisance

**How**: added checks in `input_handleBlur` and `button_handleBlur` to make sure that the new `activeElement` is not a downshift button or downshift input, respectively. Added cypress tests which pass, but I have some work to do regarding keeping and obtaining 100% test coverage

**Checklist**:

* [ ] Documentation - N/A
* [ ] Tests - kinda
* [ ] Ready to be merged
      
Had to add setTimeout in order to prevent the activeElement from being <body>. I looked around and couldn't find any other way to do it. See this link for more info: https://stackoverflow.com/questions/121499/when-a-blur-event-occurs-how-can-i-find-out-which-element-focus-went-to

It all works, and the cypress tests pass.

Having some issues with the jest tests:

```
● on input blur resets the state

expect(jest.fn()).toHaveBeenLastCalledWith(expected)

Expected mock function to have been last called with:
  [ObjectContaining {"inputValue": "animal", "isOpen": false, "selectedItem": "animal"}]
But it was not called.

      170 |   const {input, renderSpy, items} = setupDownshiftWithState()
      171 |   input.simulate('blur')
    > 172 |   expect(renderSpy).toHaveBeenLastCalledWith(
      173 |     expect.objectContaining({
      174 |       isOpen: false,
      175 |       inputValue: items[0],
```
Also, I need to add jest tests for the `setTimeout` fns.

I'll look into my testing issues later, but if anyone has suggestions, let me know!

### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -689,8 +689,8 @@ class Downshift extends Component {
     // Need setTimeout, so that when the user presses Tab, the activeElement is the next focused element, not the body element
     setTimeout(() => {
       const downshiftButtonIsActive =
-        this.props.environment.document.activeElement.tagName === 'BUTTON' &&
-        this.props.environment.document.activeElement.dataset.toggle
+        this.props.environment.document.activeElement.dataset.toggle &&
+        this._rootNode.contains(this.props.environment.document.activeElement)
       if (!this.isMouseDown && !downshiftButtonIsActive) {
         this.reset({type: Downshift.stateChangeTypes.blurInput})
       }
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -1,6 +1,6 @@
 /* eslint camelcase:0 */
 
-import React, {Component} from 'react'
+import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import preval from 'preval.macro'
 import setA11yStatus from './set-a11y-status'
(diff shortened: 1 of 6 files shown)


kentcdodds:
Awesome! The Cypress test looks super!

For the Jest test, you should be able to add `jest.runAllTimers()` after the blur event and it'll work.

Learn more: https://facebook.github.io/jest/docs/en/timer-mocks.html

Also, if you have to skip the precommit hook (because tests are busted) but still want things formatted, you can run: `npx kcd-scripts format ./path/to/file` and it'll format that file for you :)

Thanks so much!

austintackaberry:
Thanks for the help! I managed to get all of the existing Jest tests to pass. Now just working on the test coverage. Also, thanks for the formatting tip.

austintackaberry:
I think I understand your suggestion. I added another cypress test to catch the issue you pointed out. Thanks for the review!

src/downshift.js
```diff
@@ -679,9 +686,15 @@ class Downshift extends Component {
...
 
   input_handleBlur = () => {
-    if (!this.isMouseDown) {
-      this.reset({type: Downshift.stateChangeTypes.blurInput})
-    }
+    // Need setTimeout, so that when the user presses Tab, the activeElement is the next focused element, not the body element
+    setTimeout(() => {
+      const downshiftButtonIsActive =
+        this.props.environment.document.activeElement.tagName === 'BUTTON' &&
+        this.props.environment.document.activeElement.dataset.toggle
```
kentcdodds: I actually preferred when this was set to `contains`, I'd rather change this to be:

```javascript
const downshiftButtonIsActive =
  this.props.environment.document.activeElement.dataset.toggle &&
  this._rootNode.contains(this.props.environment.document.activeElement)
```

austintackaberry: Oh, yeah that makes a lot more sense

PR #376, opened 2018-03-15
Add style to TS types

jaredpalmer:
Saw this was passed directly to `<Downshift>` in the react-popper example.

Reference: https://github.com/paypal/downshift/blob/master/stories/examples/react-popper.js#L49

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -60,6 +60,7 @@ export interface DownshiftProps {
   render?: ChildrenFunction
   children?: ChildrenFunction
   id?: string
+  style?: React.CSSProperties
   environment?: Environment
   onOuterClick?: () => void
   onUserAction?: (


donysukardi:
Downshift doesn't render anything by itself, hence the style prop is not necessary as it's not used anywhere.

Perhaps we should update the example instead.

kentcdodds:
Yeah, I'm thinking that maybe it works by accident [due to this](https://github.com/paypal/downshift/blob/04635b2d5af910c9c137f06e4d894f1c50f731f2/src/downshift.js#L935-L938) but that would be surprising I think... I'm guessing it actually doesn't work `<scratching-head-emoji>`.

donysukardi:
That doesn’t pass down the style from this.props though. I checked the sample, verified that style doesn’t get rendered anywhere :D

kentcdodds:
Nice! Let's just update the example then.

Thanks for your diligence anyway @jaredpalmer!

PR #383, opened 2018-03-20
fix(mouseup): do not reset when active element is in downshift

### LINKED ISSUE #356: Default behavior for mouseUp event
**v1.28.1 issue:**

```javascript
const onMouseUp = event => {
        this.isMouseDown = false
        if (
          (event.target === this._rootNode ||
            !this._rootNode.contains(event.target)) &&
          this.getState().isOpen
        ) {
          this.reset({type: Downshift.stateChangeTypes.mouseUp}, () =>
            this.props.onOuterClick(this.getStateAndHelpers()),
          )
        }
      }
```

**v1.28.3 solution:**

```javascript
const onMouseUp = event => {
        const {document} = this.props.environment
        this.isMouseDown = false
        if (
          (event.target === this._rootNode ||
            !this._rootNode.contains(event.target)) &&
          this.getState().isOpen &&
          (!this.inputId || document.activeElement.id !== this.inputId)
        ) {
          this.reset({type: Downshift.stateChangeTypes.mouseUp}, () =>
            this.props.onOuterClick(this.getStateAndHelpers()),
          )
        }
      }
```

**What you did and what happened:**

v1.28.1 Issue
v1.28.1 Issue

**Problem description:**

Previously, downshift would reset if your mouseUp event.target was outside of downshift even though mouseDown happened on the input element. Thanks to @jduthon, in v1.28.3, it only resets if event.target is outside of downshift AND the active element is not the input element. This solution, however, does not support multiple inputs.

**Suggested solution:**

An alternative solution could be to only reset if event.target is NOT within the downshift root node AND the active element is in the root downshift node. This option would support multiple input elements. However, it would also mean that clicking on the label of a downshift input would NOT cause a reset.

**I opened this issue hoping to get additional opinions on this as a result of https://github.com/paypal/downshift/pull/354#issuecomment-370231341. Thoughts??**

kentcdodds:
**What**: do not reset when active element is in downshift

**Why**: Closes #356

**How**:
- Add util for checking whether a node is or contains something
- Use that util for both the event.target and the document.activeElement
- Use that instead of the input

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged
      
### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -14,6 +14,7 @@ import {
   getA11yStatusMessage,
   unwrapArray,
   isDOMElement,
+  isOrContainsNode,
   getElementProps,
   noop,
   requiredProp,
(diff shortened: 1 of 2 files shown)


kentcdodds:
This is the last PR before I create the `next` branch. All future work will start there until it's merged. Hopefully we'll get that merged before the week is out.

PR #390, opened 2018-03-21
fix(exports): removed statics assignments to default export

### LINKED ISSUE #380: Remove .default & .resetIdCounter from index.js
Problem description:
Such static property assignments prevent tree-shaking (which is still relevant in case of `downshift` in case it gets imported by some some unused part of the code, i.e. some ui-kit library which index.js reexports bunch of stuff):

https://github.com/paypal/downshift/blob/d2df6ec05198609ec5abbbad92bdc91dcac3b8c9/src/index.js#L9-L10

Not mentioning that adding `.default` is not proper way of supporting a default import declaration. We've discussed this once here - https://github.com/kentcdodds/kcd-scripts/pull/28

Suggested solution:

1. Remove `.default` and use `exports: 'named'` in rollup.config.js or create a separate file for `cjs` that could wrap the original entry to provide `.default`
2. make `.resetIdCounter` a separate export and not a static property of `Downshift` class

Andarist:
fixes #380

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged
      
### DIFF
--- a/README.md
+++ b/README.md
@@ -774,9 +774,8 @@ Allows reseting the internal id counter which is used to generate unique ids for
 You should never need to use this in the browser. Only if you are running an universal React app that is rendered on the server you should call [resetIdCounter](#resetidcounter) before every render so that the ids that get generated on the server match the ids generated in the browser.
 
 ```javascript
-import Downshift from 'downshift';
+import {resetIdCounter} from 'downshift';
 
-Downshift.resetIdCounter();
 ReactDOMServer.renderToString(...);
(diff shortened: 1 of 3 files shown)


kentcdodds:
Thanks for this start! Do you mind updating the tests? Also, make sure to rebase your branch on `next` :+1:

Andarist:
Done & done

kentcdodds:
Excellent! Thanks!

Issue #389, opened 2018-03-21
[Typescript] Missing setItemCount & unsetItemCount action typings


ghost:
* `downshift` version:   All since v1.22.0
* `node` version:   v8.9.4
* `npm` (or `yarn`) version:   npm v5.6.0 / yarn v1.5.1

**What you did:**
Tried using the `setItemCount` and/or `unsetItemCount` actions within a Typescript project.

**What happened:**
"Type 'ControllerStateAndHelpers' has no property 'setItemCount' and no string index signature".

**Problem description / Suggested solution:**
The typings file needs `setItemCount` and `unsetItemCount` added to the `Actions` interface signature:

```javascript
export interface Actions {
  ...
  setItemCount: (count: number) => void
  unsetItemCount: () => void
  ...
}
```

stereobooster:
Hi @Sharakai. Do you want to do PR for this?

Sharakai:
@stereobooster Happy to when I get home from work, but won't be until late night (~9pm UTC).

stereobooster:
No need to rush. Team is across the globe. Somebody will handle it as soon as you do your PR.

PR #391, opened 2018-03-21
fix(typings): added missing actions typings

Sharakai:
**What**:
Added Typescript typings for the `setItemCount`, `unsetItemCount`, and `setState` actions.

**Why**:
This resolves https://github.com/paypal/downshift/issues/389 and allows consuming Typescript projects to use these actions.

**How**:
Added the 3 functions to the `Actions` interface to match the implementations:
* `setItemCount`: [downshift.js#L196](https://github.com/paypal/downshift/blob/master/src/downshift.js#L196)
* `unsetItemCount`: [downshift.js#L197](https://github.com/paypal/downshift/blob/master/src/downshift.js#L197)
* `setState`: [downshift.js#L300](https://github.com/paypal/downshift/blob/master/src/downshift.js#L300)

(Although `setItemCount` and `unsetItemCount` do actually return the value/null respectively, I've regarded them as irrelevant implementation details)

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged
      
**Additional comments**

@Antontelesh As you commented in https://github.com/paypal/downshift/pull/359#issuecomment-370655006, I agree it's nicer to remove `itemToString` from actions. Happy to do this as part of this PR. Can't use `Pick<DownshiftProps, 'itemToString'>` as it's an optional, so will have to make a new `interface`. Any suggestions on names? `ControllerPassThroughProps`? `PassThroughProps`, ...?


typings/index.d.ts
```diff
@@ -146,6 +146,9 @@ export interface Actions {
   ) => void
   clearSelection: (cb?: Function) => void
   clearItems: () => void
+  setItemCount: (count: number) => void
+  unsetItemCount: () => void
+  setState: (stateToSet: object, cb?: Function) => void
```
stereobooster: isn't setState defined by React.Component?

kentcdodds: Could probably make this more strict. I don't know TypeScript, but the possible values are:

```javascript
{
  type: string,
  isOpen: boolean,
  inputValue: string,
  selectedItem: any,
  highlightedIndex: number | null,
}
```

kentcdodds: In a recent change we exposed our `internalSetState` method as an action (called it `setState`).

Sharakai: Agreed. I've also missed the fact that `stateToSet` can be a function.
Looking at the implementations for the actions with a `otherStateToSet` argument, and I think all the `otherStateToSet?: object` types could be stricter, too, perhaps using `Partial<StateChangeOptions>`.

kentcdodds: Talk to @stereobooster. They've done some work in the `next` branch. We're preparing for downshift v2.0.0. :)

stereobooster: This is the PR Kent talking about https://github.com/paypal/downshift/pull/373/files.

If you want to take into account those changes you need to create branch based on `next`.

kentcdodds:
The contributor of this PR is now a "ghost." I think that means their account was suspended or something. So I'll go ahead and close this PR and we can reopen it or another one if changes are necessary.

PR #394, opened 2018-03-22
391 fix type signatures

stereobooster:
**What**: fixed issues with type signatures mentioned in #391

**Why**:

**How**:

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [x] Ready to be merged
      
Also moved some signatures to be in the same order as in JS file

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -94,6 +94,10 @@ export interface StateChangeOptions<Item> {
   selectedItem: Item
 }
 
+type StateChangeFunction<Item> = (
+  state: DownshiftState<Item>,
+) => StateChangeOptions<Item>
+
 export interface GetRootPropsOptions {
   refKey: string
 }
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -119,14 +123,16 @@ export interface GetItemPropsOptions<Item>
...
   getButtonProps: (options?: getToggleButtonPropsOptions) => any
+  getToggleButtonProps: (options?: getToggleButtonPropsOptions) => any
   getLabelProps: (options?: GetLabelPropsOptions) => any
+  getMenuProps: (options?: {}) => any
   getInputProps: (options?: GetInputPropsOptions) => any
   getItemProps: (options: GetItemPropsOptions<Item>) => any
 }
 
 export interface Actions<Item> {
+  reset: (otherStateToSet?: {}, cb?: Callback) => void
```
stereobooster: Maybe `reset: (otherStateToSet?: DownshiftState<Item>, cb?: Callback) => void`?

PR #399, opened 2018-03-24
[#23] Flow definitions

stereobooster:
**What**: Added flow types

**Why**:

**How**: Used tool to convert TS definitions to Flow and edited file by hand afterwards. Added flow-covertage with 95% as a test

**Checklist**:

* N/A Documentation
* [x] Tests
* [x] Ready to be merged
      
* N/A Added myself to contributors table
      

This is working, but I want to get some feedback before merge.

It is not possible to achieve 100% flow coverage, because some functions return `any`

screen shot 2018-03-24 at 18 48 18


src/index.js
```diff
@@ -1,2 +1,5 @@
-export {default} from './downshift'
+import Downshift from './downshift'
+
+export default Downshift
+export const downshiftFactory = () => Downshift
```
stereobooster: have no better ideas how to do this

flow-typed/npm/downshift_vx.x.x.js.flow
```diff
@@ -0,0 +1,180 @@
+/**
```
kentcdodds: I don't know very much about how this file is generated, but I would expect the file to have a more specific version: `flow-typed/npm/downshift_v1.x.x.js.flow`

package.json
```diff
@@ -110,5 +114,14 @@
     "url": "https://github.com/paypal/downshift/issues"
   },
   "homepage": "https://github.com/paypal/downshift#readme",
-  "dependencies": {}
+  "flow-coverage-report": {
+    "includeGlob": [
+      "test/**/*.js",
+      "test/**/*.jsx"
+    ],
+    "threshold": 95,
```
kentcdodds: This is fine with me.

package.json
```diff
@@ -110,5 +114,14 @@
     "url": "https://github.com/paypal/downshift/issues"
   },
   "homepage": "https://github.com/paypal/downshift#readme",
-  "dependencies": {}
+  "flow-coverage-report": {
+    "includeGlob": [
+      "test/**/*.js",
+      "test/**/*.jsx"
```
kentcdodds: I prefer to not use `.jsx` as an extension and just use `.js` for everything. Could you make that change? I realize that the TS files have the `.tsx` extension, but that wasn't my choice and I try to stay out of TS decisions. Thanks :)

flow-typed/npm/downshift_vx.x.x.js.flow
```diff
@@ -0,0 +1,180 @@
...
+ * Flowtype definitions for index
+ * Generated by Flowgen from a Typescript Definition
```
kentcdodds: Should we have a script in the package.json for regenerating this file?

Potentially it'd be great if we do this as a precommit hook so we don't have to run that script manually:

```js
// lint-staged.config.js
const baseConfig = require('kcd-scripts/config').lintStaged
module.exports = Object.assign(baseConfig, {
  linters: Object.assign(baseConfig.linters, {
    '**/typings/**/*.d.ts': [
      'npm run generate-flow',
      'git add flow-typed'
    ]
  })
})
```

What do you think?

stereobooster: The problem is that this generator doesn't work. I needed to edit generated file manually. I will fix all other your notes

flow-typed/npm/downshift_vx.x.x.js.flow
stereobooster: this PR is against next branch, so it would be `downshift_v2.x.x.js.flow`

package.json
stereobooster: This is what I get when I switch from JSX to JS

```
downshift/test/custom.test.js
 6:8  error  Unable to resolve path to module 'downshift'  import/no-unresolved
39:5  error  Unexpected console statement                  no-console
```

stereobooster:
Thank you

Issue #406, opened 2018-04-03
[Typescript] Fixes to the type definitions in index.d.ts


franklixuefei:
* `downshift` version: latest
* `node` version: 8
* `npm` (or `yarn`) version: npm

Relevant code or config
#407 

What you did:
Fixed existing types and added missing types in `index.d.ts`
What happened:
Some type definitions in `index.d.ts` were either not accurate or missing.

Reproduction repository:

Problem description:

Suggested solution:

PR #407, opened 2018-04-03
Fixed several type definitions

franklixuefei:
**What**:
Fixed several type definitions as well as added the missing ones.

**Why**:
For proper typescript support.

**How**:
Please take a look at the commits.

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [x] Ready to be merged
      

franklixuefei:
Fixes #406

typings/index.d.ts
```diff
@@ -132,20 +133,20 @@ export interface PropGetters<Item> {
 }
 
 export interface Actions<Item> {
-  reset: (otherStateToSet?: {}, cb?: Callback) => void
+  reset: (otherStateToSet?: StateChangeOptions<Item>, cb?: Callback) => void
```
stereobooster: Should it be `StateChangeOptions<Item>` or `Partial<StateChangeOptions<Item>>`?

stereobooster:
Hi @franklixuefei thanks for contribution. I have one question otherwise looks ok

typings/index.d.ts
franklixuefei: Good catch! Yes indeed it should be `Partial`

franklixuefei:
@stereobooster Fixed the comment. Good catch.

kentcdodds:
Really hoping we can have a beta 2.0.0 release in the next week or two. I'm just super busy at the moment :-/

franklixuefei:
@kentcdodds That'll be awesome!

PR #410, opened 2018-04-04
Bugfix/403

janhoogeveen:
**What**:

Adds the react-lifecycles-compat polyfill

**Why**:

Otherwise you can't use Downshift with React 16.3.x and StrictMode

**How**:

The polyfill (official one from the React team) ensures libraries can support older React versions while adding support for newer lifecycle methods. It does this by polyfilling the Downshift component with new lifecycle methods.

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [ ] Ready to be merged
* [X] Added myself to contributors table

I'm not familiar with Preact, but it looks like tests break when you use the polyfill in a Preact environment. I wrote an if statement to check if a component is a class component before applying the polyfill which ensures the Preact test suite passes, but the code coverage is not less than 100% and I have trouble testing that if statement. Some help with creating a test for this conditional is welcome.


src/downshift.js
```diff
@@ -952,6 +954,11 @@ class Downshift extends Component {
   }
 }
 
+// Polyfill your component so the new lifecycles will work with older versions of React:
+if (Component.prototype && Component.prototype.isReactComponent) {
```
kentcdodds: Add `/* istanbul ignore else (preact) */` above this line and that'll take care of coverage :+1:

src/downshift.js
```diff
@@ -952,6 +954,11 @@ class Downshift extends Component {
...
+  polyfill(Downshift)
```
kentcdodds: I'm pretty sure that we'll need to reassign Downshift to this for this to work properly:

```js
Downshift = polyfill(Downshift)
```

Perhaps we should add a test to make sure this polyfill is applied (not certain how to do that though... maybe it's not worth it).

kentcdodds:
By the way, thank you so much for doing this!

src/downshift.js
kentcdodds: Just kidding. I just checked and you've got this right. Thanks! We probably still need to actually use the new lifecycle methods though...

janhoogeveen:
From what I can see Downshift is not using any deprecated lifecycle methods! That’s the great news. 

Strict Mode calls the new lifecycle methods in order to test if they work and if there are unforeseen side effects. My feeling is you don’t nees to specify them at all (since there’s no deprecated lifecycle methods at all) and that they return null or undefined by default?

kentcdodds:
Oh, gotcha. Hmmm... I'm not sure I want to add the extra dependency just for this. Seems like dead weight :-/

janhoogeveen:
Testing if the polyfill is applied should be doable by calling a new lifecycle method and checking if it returns an expected value. The polyfill has tests as well and since Strict Mode calls these methods it should be straightforward to copy their logic into Downshift.

janhoogeveen:
Agree. It would be much better if you could exclude (sub)components in Strict Mode.

src/downshift.js
janhoogeveen: Done. Code coverage is 100% now.

janhoogeveen:
I tried out this branch with npm link today in my own project and I still got errors. I must be doing something wrong. 

I'm starting to think downshift shouldn't even give any issues with React 16.3, as it depends on React as a peer dependency right? Downshift is not built against a specific version of React as it stands?

kentcdodds:
> I'm starting to think downshift shouldn't even give any issues with React 16.3, as it depends on React as a peer dependency right? Downshift is not built against a specific version of React as it stands?

That is correct. It should just work out of the box with React 16.3. If you could dig a little further to figure out why you're seeing errors that would be great. From what I can tell you shouldn't be seeing any errors at all.

drobannx:
I noticed a similar issue with Downshift when using React.StrictMode. I debugged some and found what I think is going on.

When turning on strict mode, react says the following:
```
Strict mode can’t automatically detect side effects for you, but it can help you spot them 
by making them a little more deterministic. This is done by intentionally double-invoking 
the following methods:

- Class component constructor method
- The render method
- setState updater functions (the first argument)
- The static getDerivedStateFromProps lifecycle
```

What is happening is the `internalSetState` function is doing the following:
`var isStateToSetFunction = typeof stateToSet === 'function';` outside the closure that is returned. 

This is problematic because later this is invoked:
`stateToSet = isStateToSetFunction ? stateToSet(state) : stateToSet;`

The `stateToSet` was originally a function and works great the first time but when invoked a subsequent time due to the use of React.StrictMode, stateToSet is now an object but `isStateToSetFunction` remains true due to the closure. 

It looks like the stateToSet reference gets updated with the newState instead of a new reference being created.

Hope this helps! Let me know if you need me to provide any more info.

kentcdodds:
That's very interesting @janhoogeveen! Thank you. I think that issue is unrelated to this PR in particular. Do you suppose you could file a new issue with everything you just said so we can come up with a solution? Suggestions welcome!

drobannx:
@kentcdodds - I'm not @janhoogeveen but I can create another issue with a suggestion :)

kentcdodds:
Whoops, my bad 😅

PR #412, opened 2018-04-09
Fix keyboard events for IE and Edge in preact

patrickkettner:
**What**: IE and certain versions of Edge followed an earlier draft of the DOM spec, and do not include 'Arrow' in their `event.key` events when arrow keys are used (i.e. 'Down' instead of 'ArrowDown', etc). This has been [patched in Edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8860571/), but won't be backported to IE. 

**Why**: Preact does not use syntetic events like react does, and as a result arrow keys do not currently work with preact in those browsers.

**How**: check if a keycode is within 37-40 (the range of keyboard arrow keys), and if it does not start with `Arrow`, prepend it to the value.

**Checklist**:

* [ ] Documentation NA
* [ ] Tests NA
* [ x] Ready to be merged
      
* [ ] Added myself to contributors table NA

### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -668,7 +668,9 @@ class Downshift extends Component {
   }
 
   input_handleKeyDown = event => {
-    var key = event.key
+    let key = event.key
+    // handle issue with IE/Edge not including "Arrow" as a prefix for the key
+    /* istanbul ignore next (ie) */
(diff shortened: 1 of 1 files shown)


kentcdodds:

patrickkettner:
No worries - thanks for the fast merge! Do you have any idea when your next
release will be?

kentcdodds:
It's already released

Issue #403, opened 2018-04-01
React.StrictMode issue


oliviertassinari:
* `downshift` version: 1.31.2

Relevant code or config: https://codesandbox.io/s/9lrw4n774w

What you did: I have added `React.StrictMode`. 

What happened: a `TypeError stateToSet is not a function` error is thrown.

Notice that it works fine without `React.StrictMode`. It's potentially a React bug.

stereobooster:
```
Warning: Unsafe lifecycle methods were found within a strict-mode tree:
    in IntegrationDownshift (created by WithStyles(IntegrationDownshift))
    in WithStyles(IntegrationDownshift)

componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: Input, WithStyles(FormControl), WithStyles(Input)

componentWillUpdate: Please update the following components to use componentDidUpdate instead: Input
```

oliviertassinari:
@stereobooster I can't reproduce on a simple Downshift example: https://codesandbox.io/s/24r1j952op. I keep digging;

oliviertassinari:
Ok, I have a new reproduction example without Material-UI as a dependency: https://codesandbox.io/s/9lrw4n774w.
- focus the input
- type `a`
- blur the input

kentcdodds:
Thanks! If someone would like to use the polyfill provided by the React team and update this component to not have this error in strict mode that would be wonderful!

janhoogeveen:
@kentcdodds I'll take a stab at it!

drobannx:
I noticed a similar issue with Downshift when using React.StrictMode. I debugged some and found what I think is going on.

When turning on strict mode, react says the following:

Strict mode can’t automatically detect side effects for you, but it can help you spot them 
by making them a little more deterministic. This is done by intentionally double-invoking 
the following methods:

- Class component constructor method
- The render method
- setState updater functions (the first argument)
- The static getDerivedStateFromProps lifecycle
What is happening is the internalSetState function is doing the following:
var isStateToSetFunction = typeof stateToSet === 'function'; outside the closure that is returned.

This is problematic because later this is invoked:
stateToSet = isStateToSetFunction ? stateToSet(state) : stateToSet;

The stateToSet was originally a function and works great the first time but when invoked a subsequent time due to the use of React.StrictMode, stateToSet is now an object but isStateToSetFunction remains true due to the closure.

It looks like the stateToSet reference gets updated with the newState instead of a new reference being created.

[...]

PR #423, opened 2018-04-17
fix(internalSetState): dont change references inside setState updater

drobannx:
**What**:
This fixes an issue when using the new React 16.3 React.StrictMode component which invokes setState updater functions twice. This resolves issue #403 

**Why**:
This change is necessary in order to resolve a runtime error when using React.StrictMode. (Note: StrictMode does not affect production code, only development).

**How**:
This change changes the behavior in the setState updater function by creating new references to state and the stateToSet that is in the closure.

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [x] Ready to be merged
      
I wasn't sure if there were any additional tests that needed to be added. Let me know if there are anything and I can try and add those.


kentcdodds:
Thanks for the PR!

> This change changes the behavior in the setState updater function by creating new references to state and the stateToSet that is in the closure.

This just changes the references to objects. As far as I know, `state` is actually never mutated. Unless I'm missing something this change doesn't actually affect anything.

drobannx:
Hmm, when I was running things locally in StrictMode, these two lines seemed suspect to me:

```js
stateToSet = isStateToSetFunction ? stateToSet(state) : stateToSet
stateToSet = this.props.stateReducer(state, stateToSet)	
```

The stateToSet reference gets changed to an object with the new state after the first line. Since stateToSet is in the closure and outside this scope, it remains changed on subsequent calls to the updaterFunction.

The first time this is invoked, things work fine. But the second time, before my change, stateToSet was an object and not a function.

Does that make sense?

drobannx:
@kentcdodds - I added a failing test case. This test passes with my changes and does not without. The behavior I was seeing in our Application occurred during a call of reset() so I mimicked that in the test. 

Hope this helps, let me know if theres anything else I can provide.

src/__tests__/downshift.misc.js
```diff
@@ -116,6 +116,22 @@ test('can use children instead of render prop', () => {
   expect(childrenSpy).toHaveBeenCalledTimes(1)
 })
 
+test('should not throw error during strict mode during reset', () => {
```
kentcdodds: Thanks for this! Do you mind showing me this test failing without your changes? I'm really sorry, I'm just struggling to see how your changes really impact anything here :-/

drobannx: How would you like me to do that? Would you like me to revert my change in downshift and push? I can do that first thing tomorrow when I get to work.

kentcdodds: A screenshot of the failing test would suffice.

Or if you could point me to the part of your code that's making the real impact or explain how this is making a difference. I'm sure I'm missing something, but it looks like this is just renaming variables...

drobannx: image

@kentcdodds - Here is a screenshot of the failing test.

Let me try to explain a little more of the issue.

The `internalSetState` function is not invoked twice by react when in Strict mode. The part that is invoked twice is the updater function that is passed to `this.setState`. The first time the updater function is invoked, `stateToSet` is a function (which was passed into internalSetState). Inside the updater function, the `stateToSet` variable gets reassigned to the result of that function call which is an object (the new state). React then invokes the updater function again (as part of StrictMode). This time, `stateToSet` is an object since that was changed during the previous invocation.

My change ensures that `stateToSet` variable never gets reassigned so that it remains the same on subsequent calls of the updater function.

drobannx: Here is a CodeSandbox of the issue if this helps:
[Edit 8446ynnn7j](https://codesandbox.io/s/8446ynnn7j)

kentcdodds:

PR #424, opened 2018-04-17
Change StateChangeOptions to only extend DownshiftState as a Partial

### LINKED ISSUE #417: Most properties in the StateChangeOptions ts interface should be optional
* `downshift` version: 1.30.1
* `node` version: 8.5.0
* `npm` (or `yarn`) version:  Yarn 1.1.0

## Problem description:

The only required property on the [StateChangeOptions](https://github.com/paypal/downshift/blob/18b91795101930747819fa7274f0855183671bf9/typings/index.d.ts#L94) interface that should be required is `type`. All other properties should be optional.

This is a problem when using [`stateReducer`](https://github.com/paypal/downshift#statereducer), and you want to restructure your own version of the state you're returning. The compiler complains that it can't be assigned to `StateChangeOptions`.

### Code:

```ts
// given this function the ts compiler will throw an error saying the signature doesn't match
const stateReducer = (stateAndProps, changes) => {
    const { type } = changes;
    if( type === clickItem ) {
        // If the item to be selected has `allowSelection = true` return `changes`
        // ... otherwise return a new obj with just the type
        return changes.selectedItem.allowSelection? changes : {type};
    }
}
```
Simply making the properties optional would be easy, but would be a breaking change.  The same block of code would fail `"selectedItem" might not be defined`. (This is easily fixed by adding `!` after `.selectedItem`, but it is still a breaking change. )

## Suggested solution:

We create a permanent (breaking) fix for the next major version: Make those properties optional.

We also create a temporary change on top of it, that that can be entered in to a minor release: 

```ts
// New `RequiredStateChangeOptions` is the equivilant of the old `StateChangeOptions`
type RequiredStateChangeOptions = Required<StateChangeOptions>;

// Replace all instances of `StateChangeOptions` with the required version besides for
// the signature for `stateReducer` which will now look like this:

(state: DownshiftState, changes: RequiredStateChangeOptions ) => StateChangeOptions
```

dovidweisz:
**What**: Change the type `StateChangeOptions` to only extend `DownshiftState` as a Partial, making all of those properties optional.

It fixes #417

**Why**: 

It's the right thing to do.

**How**:

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests -- tests are broken without my changes
* [x] Ready to be merged


typings/index.d.ts
```diff
@@ -133,17 +134,30 @@ export interface PropGetters<Item> {
 }
 
 export interface Actions<Item> {
-  reset: (otherStateToSet?: Partial<StateChangeOptions<Item>>, cb?: Callback) => void
+  reset: (
+    otherStateToSet?: Partial<StateChangeOptions<Item>>,
```
stereobooster: I suppose you do not need `Partial` anymore

dovidweisz: ... only if we make `type` optional.

I'm going to do it now.

dovidweisz: actually i have my misgivings.

There are times when we know that the library is going to set the `type` property for us. It would be nice not to force people to use `!` (`changes.type!.toUpperCase()`)

stereobooster: Not sure what you want to achieve. At the moment you added partial to the interface that is already used with `Partial` wrapper most of the time. It would make sense to remove `Partial` wrapper otherwise this change does not make much sense. Or is it?

dovidweisz: As of the last commit, `type` is required.

That is not correct most of the time. Most of the time you can pass `StateChangeOptions` without `type`. However I still like having the option of telling the compiler in certain instances, that `type` exists and is not null.

An example would be the first parameter passed to `onStateChange`, or the parameter passed to `stateReducer`. We can be confident that `type` will always be defined on those objects.

It's in these instances that having `type` required is helpful.

typings/index.d.ts
```diff
@@ -57,7 +57,7 @@ export interface DownshiftProps<Item> {
   stateReducer?: (
     state: DownshiftState<Item>,
     changes: StateChangeOptions<Item>,
```
stereobooster: add `Partial` here

Issue #421, opened 2018-04-16
Dropdown in iOS Safari cannot be re-opened while focused


dalehurwitz:
* `downshift` version: 1.31.6

**Relevant code or config**

```javascript
button_handleClick = event => {
  event.preventDefault()
  // handle odd case for Safari and Firefox which
  // don't give the button the focus properly.
  /* istanbul ignore if (can't reasonably test this) */
  if (
    this.props.environment.document.activeElement ===
    this.props.environment.document.body
  ) {
    event.target.focus()
  }
  this.toggleMenu({type: Downshift.stateChangeTypes.clickButton})
}
```

**What you did:** 

In iOS Safari (iPhone 7, iPhone X) the dropdown button was clicked multiple times in succession in order to toggle the menu

**What happened:**

The menu opened once then closed, then could not be re-opened while the button remained in focus.

**Reproduction repository:**

This can be reproduced in the official dropdown example - http://downshift.netlify.com/?selectedKind=Examples&selectedStory=dropdown&full=0&addons=1&stories=1&panelRight=0

**Problem description:**

Focus needs to be removed from the button in order to re-open the menu. This line in the above code appears to cause the issue: 

`event.target.focus()`

This leads to a call to `button_handleBlur` which ultimately leaves the menu closed.

Suggested solution:

kentcdodds:
Hi @dalehurwitz! Thanks for the issue!

Hmmm... I don't have an iOS device, so I wont be able to work on this myself. Would you be willing to debug this, suggest a solution and potentially open up a PR for a fix?

dalehurwitz:
This is a tough one to dev for me as I'm running windows so the kcd-scripts don't work :| Having to debug from compiled code

kentcdodds:
Sorry about that! kcd-scripts actually runs on Windows (as evidenced by [the windows CI with appveyor](https://ci.appveyor.com/project/kentcdodds/kcd-scripts/build/1.0.210)). If it's not working on your machine then there's probably something else wrong (make sure you're on the latest version of node and npm).

aMollusk:
I've had a bit of a mess around with this in browserstack, and I think @dalehurwitz is correct. It's also due to the `setTimeout` inside the `button_handleBlur`.

I've had success by wrapping the `toggleMenu` call inside a `setTimeout`, but it feels like a band-aid for a band-aid. This does however preserve the existing behaviour focus behaviour for safari/firefox/chrome, while resolving the bug in iOS.

PR #425, opened 2018-04-19
fix: resolve ios focus/blur issue

aMollusk:
**What**:  Resolving #421 

**Why**: To fix the iOS issue where the the blur event is triggered after the on click. 

**How**: Adding a `setTimeout` to the `toggleMenu` call inside the click handler.

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged
      

arahansen:
Before I open a ticket to discuss an issue I'm seeing, I wanted to maybe ask about it here:

with this change, my tests for components consuming Downshift are broken as I now need to apply fake timers any time I interact with my component. This makes writing tests for components that consume Downshift less intuitive to write. Is there possibly another way around this that would keep Downshift more synchronous?

Also possible: perhaps I should be updating my tests to make downshift components easier to test?

kentcdodds:
Sorry that this broke your tests @arahansen. If you can think of a way to make this synchronous without also having the problems this PR was intended to solve that would be super.

Alternatively, have you looked into [`react-testing-library`](https://github.com/kentcdodds/react-testing-library)? It exposes `wait` and `waitForElement` which could make things a little more straightforward and less tied to implementation details for your downshift tests.

arahansen:
The `waitForElement` sounds like a terrific idea! My team and I have looked into react-testing-library. I think our biggest concern is around the rewrite that would be required to port our tests from enzyme to react-testing-library (if you have any thoughts on that, I'd love to hear them!) 

But in the meantime, I might try pulling in an implementation of waitForElement that can work alongside enzyme

src/__tests__/downshift.get-button-props.js
```diff
@@ -19,10 +19,12 @@ test('space on button opens and closes the menu', () => {
 test('clicking on the button opens and closes the menu', () => {
   const {button, renderSpy} = setup()
   Simulate.click(button)
+  jest.runAllTimers()
```
kentcdodds: Turns out this is required by pretty much anyone testing a component that uses downshift and it's a bit annoying. @aMollusk, could you jump into https://github.com/paypal/downshift/issues/437 and chime in to discuss an alternative/workaround for tests?

Issue #433, opened 2018-04-24
Remove support for "render" prop and only support "children"


kentcdodds:
Right now we support both `<Downshift render={() => <div />} />` and `<Downshift>{() => <div />}</Downshift>`. It's kinda nice, but now that React has adopted the children version with the new context API it'd be good to just go with `children` for cohesion. In v2.0.0 we'll drop support for the prop called render.

Abdul-Moiz-Ansari:
I would love to do it. Also suggest that we should add a nice informational message in console in case if there is a 'render' prop. Just like React does in various cases.

kentcdodds:
This has been merged into the `next` branch which should be merged into `master` soon.

smelliot:
It might help to update the Readme as it still has a link that says "Use a render prop!" and links to this article: https://cdb.reacttraining.com/use-a-render-prop-50de598f11ce

blocka:
@smelliot a render prop doesn't literally mean "a prop called render". It's the idea of passing in a function to a component, whether through an explicit prop (which most people would call "render" for lack of anything better) or through the implicit "children" prop (otherwise known as "function as a child").

smelliot:
@blocka thanks, but I was merely trying to point out that there could be some confusion caused by this due to how broadly "prop" is used to describe things and that not everyone would know that it could refer to "function as a child" (this is only mentioned in one small paragraph in the article and there is no downshift documentation to say the "prop called render" which is demonstrated in the linked article is not supported)

PR #435, opened 2018-04-26
Recommend functional child method instead of renderProps

Anenth:
**What**:
Remove the use of renderProps and use functional child; Toward #433 

**Why**:

Because of #433

**Checklist**:

* [x] Documentation
* [ ] Tests
* [x] Ready to be merged
      

Anenth:
@kentcdodds yaay! thanks!

PR #451, opened 2018-06-01
fix access error on document for react native

samuelfullerthomas:
**What**:
Solution for this issue: https://github.com/paypal/downshift/issues/334
Handles the other cases where `document` is undefined

**Why**:
Error getting thrown for React Native due to `document` being undefined

I think I could use some help in thinking of the best way to write a test for this.

**Checklist**:

* [x] Tests
* [x] Ready to be merged
      

samuelfullerthomas:
The underlying issue is this line: https://github.com/paypal/downshift/blob/master/src/downshift.js#L95, and then the fact that, from there, there aren't checks to see if the environment is just an empty object or not.

src/downshift.js
```diff
@@ -658,12 +663,10 @@ class Downshift extends Component {
     this.inputId = firstDefined(this.inputId, rest.id, `${this.id}-input`)
     let onChangeKey
     /* istanbul ignore next (preact) */
-    if (preval`module.exports = process.env.BUILD_PREACT === 'true'`) {
+    if (this.props.isPreact) {
```
kentcdodds: I'm sorry, but doing this will prevent these statements from being eliminated during uglification resulting in a bigger bundle. Please restore all of these to what they were before.

README.md
```diff
@@ -59,52 +59,52 @@ harder to contribute to.
 
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
-* [Installation](#installation)
```
kentcdodds: I think your editor may have messed this stuff up. If you could run `npx kcd-scripts format` before you commit that will ensure that everything's formatted properly. Thanks!

src/downshift.js
samuelfullerthomas: 👍

README.md
samuelfullerthomas: It most assuredly did 😬

samuelfullerthomas:
addressed, I think

kentcdodds:
Ok, yeah, so I'm not sure what the preval is not defined issue is all about in the build. Are you getting that locally? Try running `npm run validate` to see.

As for code coverage, we probably should set up some way to test the native build, but unfortunately that would take a fair amount of work. So if you can't cover some code due to react native, then add some well-placed `/* istanbul ignore next */` comments. You can see the coverage report by opening `coverage/lcov-report/index.html` in your browser. Good luck!

samuelfullerthomas:
@kentcdodds passing tests now!

other/react-native/__tests__/__snapshots__/render-tests.js.snap
```diff
@@ -9,9 +9,7 @@ exports[`renders with React Native components 1`] = `
     aria-expanded={false}
     autoComplete="off"
     id="downshift-0-input"
-    onBlur={[Function]}
```
kentcdodds: Hmmm... It feels like this would be a regression.

samuelfullerthomas: Shouldn't this snapshot (the react native snapshot) reflect that the code in the react native build doesn't handle those props / that they won't be there? I could pass through noops but I don't think that'd be right.

kentcdodds: Right, but before your changes, the input had these event handlers for react-native and after your chances it doesn't. That's the regression I think.

src/downshift.js
```diff
@@ -656,30 +658,42 @@ class Downshift extends Component {
...
+          this.input_handleChange,
+        ),
+        onKeyDown: composeEventHandlers(onKeyDown, this.input_handleKeyDown),
+        onBlur: composeEventHandlers(onBlur, this.input_handleBlur),
+      }
     } else if (
       preval`module.exports = process.env.BUILD_REACT_NATIVE === 'true'`
     ) {
-      onChangeKey = 'onChangeText'
+      eventHandlers = {
```
eliperkins: What's the rationale behind removing `onBlur` from the event handlers that we use with React Native?

other/react-native/__tests__/__snapshots__/render-tests.js.snap
eliperkins: Agreed. We should ensure that the onBlur handler is still passed here.

other/react-native/__tests__/__snapshots__/render-tests.js.snap
```diff
@@ -9,9 +9,7 @@ exports[`renders with React Native components 1`] = `
...
     onChangeText={[Function]}
-    onKeyDown={[Function]}
```
eliperkins: I'm fine with removing this one, as `onKeyDown` is unhandled on React Native.

samuelfullerthomas:
Thanks for the review @eliperkins I think I forgot about this PR - I'll make the changes tonight. Apologies!

samuelfullerthomas:
Hey @eliperkins and @kentcdodds - I've updated this PR to work with the master  - it's ready to review again.

kentcdodds:
Thank you so much!

kentcdodds:
Hmmmmm... It appears that the react native snapshots are failing. Care to give this a look?

https://travis-ci.org/paypal/downshift/builds/424888379

We can't release this until the build is working.

samuelfullerthomas:
Hey @kentcdodds I've updated the snapshot - `npm run test:build` works for me locally now

PR #462, opened 2018-06-13
fix(native): Support both onChange and onChangeText props

### LINKED ISSUE #361: Support `onChange` and `onChangeText` props for React Native
[After having some discussion around implementing a test suite for React Native](https://github.com/paypal/downshift/pull/328#discussion_r167035528), we realized there's some tweaks that we should make with regards to the `onChange` prop supplied to a `Downshift` component on React Native.

### Problem description:

The `onChange` prop is meant to emulate the prop of the same name that is used with `<input>` elements. On React Native, `<TextInput>` elements take in either `onChange` or `onChangeText`. The difference is that `onChange` sends an event, with the input value wrapped as part of the event, while `onChangeText` [sends only the input value](https://github.com/facebook/react-native/blob/6f007e8957c9bf5652b0184cba65f385050a8236/Libraries/Components/TextInput/TextInput.js#L925-L927).

Currently, Downshift will call `onChange` with the text input value, rather than an event object, which isn't super intuitive for React Native developers who are familiar with `<TextInput>`'s API.

### Suggested solution:

If we want to support what React Native developers would expect to supply as props here, we should support both `onChange` and `onChangeText` as values supplied to `getInputProps`.

This will likely mean checking for the existence of both keys as input to `getInputProps` when on React Native, and firing both/either `onChange` and/or `onChangeText` with the appropriate payloads, keeping Downshift's state in sync with the component.

eliperkins:
**What**:

This supports both props that <TextInput /> accepts for supporting text
input changed events.

Additionally, this moves onChange related tests to their own module.

**Why**:

Fixes #361 

**How**:

Adds additional argument to `getInputProps` that is only used on React Native builds. This argument is then used to provide an additional event handler.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged
      
### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -639,31 +639,49 @@ class Downshift extends Component {
 
   /////////////////////////////// INPUT
 
-  getInputProps = ({onKeyDown, onBlur, onChange, onInput, ...rest} = {}) => {
+  getInputProps = ({
+    onKeyDown,
+    onBlur,
+    onChange,
+    onInput,
+    onChangeText,
+    ...rest
+  } = {}) => {
     let onChangeKey
+    let eventHandlers = {}
+
     /* istanbul ignore next (preact) */
     if (preval`module.exports = process.env.BUILD_PREACT === 'true'`) {
       onChangeKey = 'onInput'
-      /* istanbul ignore next (react-native) */
-    } else if (
-      preval`module.exports = process.env.BUILD_REACT_NATIVE === 'true'`
-    ) {
-      onChangeKey = 'onChangeText'
     } else {
       onChangeKey = 'onChange'
     }
     const {inputValue, isOpen, highlightedIndex} = this.getState()
-    const eventHandlers = rest.disabled
-      ? {}
-      : {
-          [onChangeKey]: callAllEventHandlers(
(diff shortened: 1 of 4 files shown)


eliperkins:
I PR'ed this into the `next` branch, but I suppose this is also source-compatible with v1.x. Let me know if you want me to PR this into `master` instead!

src/downshift.js
```diff
@@ -639,31 +639,49 @@ class Downshift extends Component {
...
-          [onChangeKey]: callAllEventHandlers(
-            onChange,
-            onInput,
-            this.input_handleChange,
-          ),
-          onKeyDown: callAllEventHandlers(onKeyDown, this.input_handleKeyDown),
-          onBlur: callAllEventHandlers(onBlur, this.input_handleBlur),
-        }
+
+    if (!rest.disabled) {
```
eliperkins: I don't love how this forking logic came out, so I'd love some feedback here.

kentcdodds: Actually, I don't think it's all that bad :+1:

kentcdodds:
Thanks a bunch @eliperkins!

PR #468, opened 2018-06-18
Respect defaultInputValue when provided

### LINKED ISSUE #467: Respect defaultInputValue when provided
Hi, thanks a lot for this UI-less solution: it's been really easy to integrate with our existing UI lib as a substitution of a (really) old jQuery based one!

I'm having one single problem: I provide both a `selectedItem` and a `defaultInputValue` to downshift, and since `inputValue` and `selectedItem` are decoupled, I'd expect that in my initial rendering I would see `defaultInputValue`.

In practice [this snippet of code](https://github.com/paypal/downshift/blob/master/src/downshift.js#L129) in the constructor:
```js
if (state.selectedItem != null) {
  state.inputValue = this.props.itemToString(state.selectedItem)
}
```
prevents that, effectively coupling the two fields.

I'd suggest to also check wether `this.props.defaultInputValue` is `null` in that if condition.

EnoahNetzach:
fixes #467.

This is just a POC for now: tests, docs etc.. will be updated if it's decided to accept the proposal in the main issue.

### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -126,7 +126,7 @@ class Downshift extends Component {
       inputValue: this.props.defaultInputValue,
       selectedItem: this.props.defaultSelectedItem,
     })
-    if (state.selectedItem != null) {
+    if (this.props.defaultInputValue == null && state.selectedItem != null) {
       state.inputValue = this.props.itemToString(state.selectedItem)
     }
     this.state = state


ashishpatelcs:
Can you please tell me why Travic CI build has failed?

kentcdodds:
This is going to be handled by #592

PR #473, opened 2018-06-19
Publish flow typings

EnoahNetzach:
Currently the `flow-typed` dir is not published.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
   "files": [
     "dist",
     "typings",
-    "preact"
+    "preact",
+    "flow-typed"
   ],
   "keywords": [
     "enhanced input",


stereobooster:
It seems flow doesn't catch it this way. I guess we need to do this:

```
downshift.cjs.js
downshift.cjs.js.flow
...
```

This is what I get when I google how to distribute flow signature via npm https://blog.kentcdodds.com/distributing-flow-type-definitions-for-node-and-browser-modules-3952ad38b357

kentcdodds:
Yeah, we probably need a postbuild script to copy the definitions over or add definitions that export those it something 🤔

EnoahNetzach:
it should be set as a lib in the `.flowconfig` file:

```
[libs]
flow-typed
node_modules/downshift/flow-typed
```

I'll get back to you soon to verify if it's working ;)

stereobooster:
well yes, except you don't want to configure it each time, it should work out of the box

PR #479, opened 2018-06-21
Fix memory leak warning (#476)

### LINKED ISSUE #476: memory leak warning
- `downshift` version: 2.0.2
- `node` version: 9.11.1
- `npm` (or `yarn`) version: 1.7.0

Relevant code or config

See problem description

What you did:
Submitted a form with enter when focus is on the autocomplete component

What happened:
React warns that "Can't call setState (or forceUpdate) on an unmounted component"

Problem description:
Every time I submit a form when the focus is on a component that uses downshift and that component has the focus, if the component is unmounted after the submit (for example for a form in a dialog that's dismissed after submission), react warns that "Can't call setState (or forceUpdate) on an unmounted component"

Suggested solution:
setTimeout is used to manage blur events but is not cleared when the component is unmounted. When a form is submitted and unmounted after submission (for example a form in a dialog), this causes a memory leak.
I suggest to clear pending timeouts in componentWillUnmount

alexandernanberg:
**What**:
Don't run setState on unmounted components. Fixes #476

**Why**:
To prevent a possible memory leak plus removing a warning being emitted in the console from React.

**How**:
In #476 you proposed two different solutions for this and I decided to go for the solution with the `_isMounted` flag. That flag was already used for other things so it felt natural to just go with that solutions for this as well. Although I still think that clearing the timeouts is a better and more robust overall solution for the problem, so it might be worth revisiting this later.

~~I'm unsure if you would like me to add tests for this? If you would like that, I'd like some guidance on how you would like that to be implemented.~~

**Checklist**:

- [x] Documentation (Don't think any new documentation is needed)
- [x] Tests
- [x] Ready to be merged
      

alexandernanberg:
I've now added tests for this as well. I've not worked a lot with either Jest or react-testing-library so if there is anything that you want me to edit please tell me and I'll try to fix it 😄

kentcdodds:

PR #484, opened 2018-06-22
fix: handle toggle button keyboard events properly (#471)

### LINKED ISSUE #471: Using spacebar to open menu has issues in Firefox
- `downshift` version: 2.0.2
- `node` version: 8.11.2
- `npm` (or `yarn`) version: 6.1
- `firefox` version: 60.0.2

What you did:
Open Firefox, go to the linked Storybook page, focus the button and then press the spacebar to open (or close) the menu.

What happened:
The menu opens but then closes right after. If you hold the spacebar down the menu will start to flicker between the open and close state.

I also noticed that the `onChange` event fires when pressing the spacebar, but the `selectedItem` argument is undefined. 

```javascript
<Downshift 
  onChange={(selectedItem /* this becomes undefined */) => {}}
/>
```

Reproduction repository:
http://downshift.netlify.com/?selectedKind=Examples&selectedStory=dropdown&full=0&addons=1&stories=1&panelRight=0

Problem description:

Suggested solution:
Not sure what causes the problem so hard to say, Firefox must handle some events differently is my guess.

alexandernanberg:
**What**:
Keyboard events on the toggle button works as you would expect. Pressing space or enter should toggle the menu. Fixes #471

**Why**:
To provide a better experience for users that only uses the keyboard for navigation.

**How**:
I added a `onKeyUp` event listener which only runs `event.preventDefault()` to prevent Firefox from firing both the `keyDown` and the `click` event when you press the spacebar.

I also modified the `Enter` handler in the `keyDownHandlers` so it's now possible the properly toggle menu just by pressing the `Enter` key. Before you would only be able to open the menu. I did this by adding a check if `highlightedIndex` is set before we try to select the highlighted item.

**Checklist**:

- [x] Documentation (Don't think we need to add anything for this?)
- [x] Tests
- [x] Ready to be merged
      
### DIFF
--- a/src/__tests__/downshift.get-button-props.js
+++ b/src/__tests__/downshift.get-button-props.js
@@ -7,10 +7,12 @@ jest.useFakeTimers()
 test('space on button opens and closes the menu', () => {
   const {button, childrenSpy} = setup()
   Simulate.keyDown(button, {key: ' '})
+  Simulate.keyUp(button, {key: ' '})
   expect(childrenSpy).toHaveBeenLastCalledWith(
     expect.objectContaining({isOpen: true}),
   )
   Simulate.keyDown(button, {key: ' '})
+  Simulate.keyUp(button, {key: ' '})
   expect(childrenSpy).toHaveBeenLastCalledWith(
     expect.objectContaining({isOpen: false}),
   )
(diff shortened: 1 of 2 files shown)


alexandernanberg:
I've tested this PR in Edge 42, Chrome 67, Firefox 60 and Safari 11 and I haven't found any issues caused by the new `keyUp` handler. Is there anything else we should test or should this be ok?

PR #487, opened 2018-06-24
fix(utils): Replace utils.scrollIntoView with compute-scroll-into-view

stipsan:
**What**: Picking up where #259 left off with a new version of https://github.com/stipsan/scroll-into-view-if-needed#readme that include only the parts needed in `downshift`: https://github.com/stipsan/compute-scroll-into-view

**Why**: Since then the library have been refactored completely to ponyfill future `Element.scrollIntoView` that are in the CSS draft spec, instead of `Element.scrollIntoViewIfNeeded`.

Additionally the animation part is moved to a separate project https://github.com/stipsan/smooth-scroll-into-view-if-needed so the tricky workarounds from last time is no longer necessary to get the smallest possible bundlesize.

**How**: swap out `utils.scrollIntoView` with `compute-scroll-into-view`

**Checklist**:

- [N/A] Documentation
- [x] Tests
- [x] Ready to be merged

The implementation is using the [boundary](https://github.com/stipsan/scroll-into-view-if-needed#boundary) option to behave as close to the original as possible.
However if we remove this option we can unlock some nice benefits like scrolling the whole chain of parent elements into view correctly:

kapture 2018-06-25 at 0 30 22

This also works for horizontal scrolling 😄


kentcdodds:
Thanks for this! I'm really hesitant to merge this because that dependency sizeable (it's 15% of downshift's current size). But I like the idea of solving this problem once and for all. Also hopefully we could drop the dependency entirely in the future as well.

I'd love to hear input from other people though.

stipsan:
Yeah, so far the priorities have been to make it standards compliant first, get full test coverage second and reduce file size third.
Test coverage reached a point recently where it allows me to take another go at reducing bundle size without fearing regressions, although it’s becoming quite difficult to accomplish that without sacrificing spec compliance.

> Also hopefully we could drop the dependency entirely in the future as well.

This was actually one of the guiding principles going from v1 to v2. The beauty of ponyfills is their promise of future obsolescence 😄

kentcdodds:
See how far you can cut the build down without sacrificing spec compliance. Perhaps in the future when this is actually implemented in some browsers we can drop the dependency and mention that people need to polyfill the API so people can drop the dependency as their browser support requirements allow.

stipsan:
Sure thing, I’ll have a go at it and report back.
Yeah, and using a ponyfill is a good signal to browser vendors that this feature is worth implementing 😃 
block: “nearest” is starting to see adoption (chrome supports it), it would be possible to create a lite/modern version of this library that only implements the scrollMode: “if-needed” part which is far far smaller.

kentcdodds:
Yes, I think I'd be very happy to have a lighter version that only supports the use case we need 👌

stipsan:
Pushed a new version that is reduced as much as possible without breaking the spec. Compare [2.2.8](https://unpkg.com/scroll-into-view-if-needed@2.2.8/umd/) with [2.2.8](https://unpkg.com/scroll-into-view-if-needed@2.2.7/umd/) 😄

kentcdodds:
Awesome! That's definitely an improvement! What do you think it'd take to make a version that is only what we need here?

stipsan:
Sure, why not 😊 i could split out the core logic into its own package and use that directly here. No fancy non-standard features like the boundary option, no smooth scrolling support out of the box. 
It’ll take me a fee days but I think it’s worth it. 

Gotta say Kent, your chase for the most extremely minimal bundlesizes is inspiring 😉

kentcdodds:
What's inspiring is that you're working on making it happen! Thanks @stipsan!

stipsan:
I just released `v.2.2.9`, which is the smallest and most optimized version I could make without introducing breaking changes and the like. The next step will be to extract just the things you need here into https://github.com/stipsan/compute-scroll-into-view. This will also be what powers the more fully featured https://github.com/stipsan/scroll-into-view-if-needed.
I'll keep you posted 😄

stipsan:
@kentcdodds updated the PR with the lighter https://github.com/stipsan/compute-scroll-into-view package (1.18kB gzipped, vs 1.38kB).

I also removed the 4 test case that deviated from the CSSOM spec. Specifically the tests expected that:
1. Given that both the `node`, and `scrollableNode` is scrolled out of view.
2. Scroll `node` into view by adjusting scrollTop on `scrollableNode`.
3. Do not scroll `scrollableNode` into view by adjusting its scrolling box parent's `scrollTop`.

Compare that with the steps specified by the spec: https://drafts.csswg.org/cssom-view/#scroll-an-element-into-view

That's a lot to read so here's the executive summary: each element should be aligned to its `scrolling box` all the way up to the viewport.
In the case of the mentioned tests the `node` is actually within the `scrollableNode` edges, but out of view as `scrollableNode` itself is not in view. According to spec `scrollableNode` is what should be scrolled and that's why the tests expected different results from what `compute-scroll-into-view` was returning.

kentcdodds:
This is awesome! I'm ready to merge as soon as all the tests are passing. Looks like we've got an e2e test failure. Could you give that a look?

stipsan:
Ok cool! Yeah sure, I’m on the go atm but I will give it a look later today 😄

stipsan:
I am a bit confused by this one. I can't seem to be able to reproduce it locally. I pushed an empty commit to see if it was just a one-off but the same error got reported by Travis. Will digg deeper and figure out what's happening tomorrow.

alexandernanberg:
Also confused about the failing e2e test, seems to be from the master branch so not necessarily from this PR. I guess we could just manually focus the element before the blur or use the `force` flag https://docs.cypress.io/api/commands/blur.html#Arguments to solve this.

stipsan:
Thanks @alexandernanberg, it appears you’re correct! 😮

Would you like to make a separate PR with a fix? 😄

@kentcdodds what do you think?

kentcdodds:
Ah, I see. More issues with it being in an iframe. I'm fine hacking around this issue by using `force` and adding a comment. Then we should probably log an issue on cypress because this is probably not desireable behavior.

We can do that in another PR. Thanks for your help friends!

kentcdodds:
Note: unless we're lucky, this probably wont be able to be auto-released until the cypress issue is fixed.

kentcdodds:

PR #488, opened 2018-06-27
Upgrade react-testing-library to v4

alexandernanberg:
**What**:
Upgrade `react-testing-library` to the latest version.

**Why**:
Keeping things up to date. Also the v4 API of `react-testing-library` is a bit cleaner and easier to grasp, so it will be easier for new contributors to add/fix tests.

**How**:
Installed the latest version and migrated the current tests to follow the new API (replace `Simlulate` with `fireEvent` etc).

The only issue I ran into was that I had to run `cleanup` in the `setup` function inside `src/__tests__/utils.reset-id-counter.js`. There might be a better way to solve this that I'm unaware of?

**Checklist**:

- [x] Documentation N/A
- [x] Tests
- [x] Ready to be merged
      
### DIFF
--- a/src/__tests__/downshift.aria.js
+++ b/src/__tests__/downshift.aria.js
@@ -1,5 +1,6 @@
+import 'react-testing-library/cleanup-after-each'
 import React from 'react'
-import {render as renderToDOM} from 'react-testing-library'
+import {render} from 'react-testing-library'
 import Downshift from '../'
 import {resetIdCounter} from '../utils'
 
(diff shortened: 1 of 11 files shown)


src/__tests__/downshift.aria.js
```diff
@@ -1,12 +1,14 @@
...
-import {render as renderToDOM} from 'react-testing-library'
+import {render, cleanup} from 'react-testing-library'
 import Downshift from '../'
 import {resetIdCounter} from '../utils'
 
 beforeEach(() => {
   resetIdCounter()
 })
 
+afterEach(cleanup)
```
kentcdodds: Rather than having an `afterEach(cleanup)` in each file, let's do: `import 'react-testing-library/cleanup-after-each'` at the top of each file. :+1:

Thanks!

alexandernanberg: Yea sounds good 👍  Don't know why it says I dismissed your review after I squashed the commits tho 😅

PR #493, opened 2018-06-29
Upgrade Storybook stories dependencies

alexandernanberg:
**What**:
Update the Storybook stories dependencies to the latest versions. With the exception of `react-popper`, they changed the API in v1 and I was unsuccessful in finding a nice way to integrate it, so I left it for now. See discussion [here](https://github.com/FezVrasta/react-popper/issues/190) and here is a codesandbox of a solution I wasn't really happy with https://codesandbox.io/s/oq9pj2288q

**Why**:
Provide up to date examples for people that's unfamiliar with the library.

**How**:
Updating was pretty straight-forward, the only code I had to edit was the Apollo example. I also removed an unused dependency (`throttle-debounce`).

I was thinking if we should add a `shared.js` similarly to how you did in your `downshift-examples` repo, to keep some style consistant etc. Or should we entirely replace the storybook examples with `downshift-examples`? 🤔 

**Checklist**:

- [x] Documentation NA
- [x] Tests 
- [x] Ready to be merged
      
### DIFF
--- a/stories/examples/apollo.js
+++ b/stories/examples/apollo.js
@@ -1,22 +1,20 @@
 import React from 'react'
-import {
-  ApolloClient,
-  ApolloProvider,
-  createNetworkInterface,
-  gql,
-  graphql,
-} from 'react-apollo'
+import {ApolloProvider, Query} from 'react-apollo'
+import ApolloClient from 'apollo-boost'
+import gql from 'graphql-tag'
 import Downshift from '../../src'
 
-export default Examples
-
-const networkInterface = createNetworkInterface({
+const client = new ApolloClient({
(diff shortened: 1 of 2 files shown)


kentcdodds:
> should we entirely replace the storybook examples with downshift-examples? 🤔

The benefit of having these examples here is for the e2e tests. So I think we should have a few, but most examples should probably be in [`downshift-examples`](https://codesandbox.io/s/github/kentcdodds/downshift-examples) because it's easier for people to play around with/fork/contribute to/etc.

alexandernanberg:
Okay, so would you accept PRs migrating all examples but the ones used for e2e tests to [`downshift-examples`](https://codesandbox.io/s/github/kentcdodds/downshift-examples) then? I'd be willing to do it if that's the case.

kentcdodds:
Definitely! 👍

PR #494, opened 2018-07-09
fix(types): Restored resetIdCounter TS type

Andarist:
I'm not a TS user, but this was reported and it seems to me that this is easy enough to add it right away


wereHamster:
The change looks good. The failure on travis seems unrelated.

PR #495, opened 2018-07-11
Downgrade cypress 

alexandernanberg:
**What**:
Fix failing Cypress tests

**Why**:
Latest version of Cypress introduced issues when calling `.blur()` on elements inside iframes.

**How**:
Downgrade and lock the Cypress version to 3.0.1. We could have used the `force` option as well but I felt like downgrading was the better option for now.
I've created an issue for this on their repo https://github.com/cypress-io/cypress/issues/2124

When this gets fixed or a good workaround is available we should unlock the version number again.

**Checklist**:

- [x] Documentation N/A
- [x] Tests
- [x] Ready to be merged
      
### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "babel-preset-env": "^1.6.1",
     "babel-preset-react-native": "^4.0.0",
     "cross-env": "^5.1.4",
-    "cypress": "^3.0.1",
+    "cypress": "3.0.1",
     "eslint-plugin-cypress": "^2.0.1",
     "flow-bin": "^0.74.0",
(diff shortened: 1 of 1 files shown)


stipsan:
Hey! Downgrading instead of changing our tests is for sure a better option, thanks for contributing this 😄 👍 

I think we should make [a follow up issue](https://github.com/paypal/downshift/issues/new) about this before merging though, to make it easier to track when we can unlock the version again 🙂

kentcdodds:
I'll go ahead and merge this now. If someone wants to make a follow up that's fine 👍 thanks!

PR #505, opened 2018-07-13
fix type for itemToString

ggilmore:
As per your [`README.md`](https://github.com/paypal/downshift#itemtostring), `itemToString` can be called with a null `Item`. 

> `function(item: any)` | defaults to: `i => (i == null ? '' : String(i))`

This PR fixes your typings to reflect that.


ggilmore:
Thanks for merging this so quickly! Do you have an idea when the next release will be?

kentcdodds:
It's been published just now! :)

ggilmore:
👍 Thanks!

PR #506, opened 2018-07-15
refactor: changing getState to improve code readability

tyagow:
Changing Object.keys to Object.entries as it gives to Object package the responsability to desconsruct the object as oposite to direct access the object.

As proposed by @kentcdodds on https://youtu.be/ubXtOROjILU?list=PLV5CVI1eNcJgNqzNwcs4UKrlJdhfDjshf&t=7823 it would be more clear to use Object.entries in getState method.

This makes the code more readable and gives the responsibility to better access the object to the Object module.

Change the Object.keys to Object.entries, making a small change inside the interaction.

**Checklist**:

- [ ] Documentation
- [X] Tests
- [X] Ready to be merged
      
- [X] Added myself to contributors table


tyagow:
@alexandernanberg  I would love to help on making sure that Babel transforms/polyfills it, could you give me some hints on how to test it?

kentcdodds:
Thanks for this! But I'm not sure this change is worth the polyfill. There's not a Babel transform for this.

kentcdodds:
I really do appreciate it @tyagow, but I don't think that this particular case is worth the extra effort to make sure we have polyfills in place. Thanks anyway!

PR #518, opened 2018-07-19
Code Refactor

prashant-andani:
Refactored code : Moved the default parameter of the function to the last.
Added my details to the contrib list.


kentcdodds:
Hi @prashant-andani, thanks for the contribution, but this would be a breaking change and I don't think it's an improvement on the API. So we'll pass on this. Thanks anyway!

PR #522, opened 2018-07-21
Refactor to remove test specific code

alexandernanberg:
**What**:
Remove test specific code and update tests to account for `setTimeout`s

**Why**:
In my opinion the tests should reflect the real-world as closely as possible. I do however agree on the reasoning behind why it was there in the first place (makes testing easier), but I don't think that weights more than the other benefits.

**How**:
Use the `jest.runAllTimers()` wherever we need to wait for a `setTimeout`

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged
      
### DIFF
--- a/src/__tests__/downshift.get-button-props.js
+++ b/src/__tests__/downshift.get-button-props.js
@@ -22,10 +22,12 @@ test('space on button opens and closes the menu', () => {
 test('clicking on the button opens and closes the menu', () => {
   const {button, childrenSpy} = setup()
   fireEvent.click(button)
+  jest.runAllTimers()
   expect(childrenSpy).toHaveBeenLastCalledWith(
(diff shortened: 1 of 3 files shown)


kentcdodds:
We did this for the benefit of _other_ people _using_ downshift and writing tests. Having this in place made it so people didn't have to know the implementation actually uses a setTimeout. If we change this back then people will have to use `jest.runAllTimers` and that is an implementation detail that I don't want people to have to concern themselves about.

Normally I'd be on board with you, but in this case I think we're better off as this is. Thanks anyway @alexandernanberg!

alexandernanberg:
Ah now I understand, makes sense! Thank for clarifying!

PR #525, opened 2018-07-22
Validate that getMenuProps has set the ref correctly

### LINKED ISSUE #490: parent.contains is not a function when used with styled components
- `downshift` version: 2.0.10
- `node` version: 9.5.0
- `npm` (or `yarn`) version: 1.5.1

Relevant code or config

```javascript
        <Downshift
            defaultSelectedItem={selectedOption}
            itemToString={item => (item ? item.message : '')}
        >
            {({
                highlightedIndex,
                itemToString,
                getItemProps,
                getMenuProps,
                getToggleButtonProps,
                isOpen,
                selectedItem,
                toggleMenu,
            }) => (
                <div>
                    <Dropdown.Container>
                        <Button {...getToggleButtonProps()}>Button Toggle</Button>
                        <Dropdown isOpen={isOpen} {...getMenuProps()}>
                            {options.map((option, index) => (
                                 <div
                                    {...getItemProps({
                                        key: option.id,
                                        index,
                                        item: option,
                                        isActive: highlightedIndex === index,
                                    })}
                                />
                            ))}
                        </Dropdown>
                    </Dropdown.Container>
                </div>
            )}
        </Downshift>
```

What you did:
The code above is pseudo code but roughly equates to what I am doing. I am building a dropdown using downshift that toggles when you click a button. I am combining this with styled components and when I click outside of the div I am getting an error from within downshift.

What happened:
The error that I am getting is the following:

screen shot 2018-06-27 at 14 40 47

When digging deeper into this, it's because the parent in isOrContainsNode is coming through as a Styled Component wrapper as you can see in the screenshot below:

screen shot 2018-06-27 at 14 45 08

Problem description:
Clicking outside of the downshift container div when using styled components is throwing an error and the dropdown window stays open.

alexandernanberg:
**What**:
Fixes #490 by using the same approach as we do in `getRootProps`.

**Why**:
Because it could lead to unexpected behavior and to avoid confusion.

**How**:
I'm still unsure of what the best solution for this is. I would like to make the check in the `render` method as we do for the `getRootProps`, but I don't think that's possible because the ref hasn't set `this._menuNode` on the first render. I guess we could loop through all children but that seems unnecessary heavy/slow.

So instead I make the validation/check in `componentDidUpdate`, which looks to work fine.

Another problem I encountered was that there is no way to check if `this._menuNode` is a composite element or not, which makes it hard to use the same validation as we do for `getRootProps`.

I wanted you opinion on this before I continue and add tests and update the documentation. Is there another better way that I've not thought of?

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged
      
### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -780,7 +780,14 @@ class Downshift extends Component {
 
   menuRef = node => (this._menuNode = node)
 
-  getMenuProps = ({refKey = 'ref', ref, ...props} = {}) => {
+  getMenuProps = (
+    {refKey = 'ref', ref, ...props} = {},
+    {suppressRefError = false} = {},
+  ) => {
+    this.getMenuProps.called = true
+    this.getMenuProps.refKey = refKey
+    this.getMenuProps.suppressRefError = suppressRefError
+
     return {
       [refKey]: callAll(ref, this.menuRef),
       role: 'listbox',
@@ -918,6 +925,10 @@ class Downshift extends Component {
   }, 200)
 
   componentDidMount() {
+    if (this.getMenuProps.called && !this.getMenuProps.suppressRefError) {
+      validateGetMenuPropsCalledCorrectly(this._menuNode, this.getMenuProps)
+    }
+
     /* istanbul ignore if (react-native) */
     if (preval`module.exports = process.env.BUILD_REACT_NATIVE === 'true'`) {
       this.cleanup = () => {
@@ -983,6 +994,10 @@ class Downshift extends Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
+    if (this.getMenuProps.called && !this.getMenuProps.suppressRefError) {
+      validateGetMenuPropsCalledCorrectly(this._menuNode, this.getMenuProps)
+    }
+
     if (
       this.isControlledProp('selectedItem') &&
       this.props.selectedItemChanged(
@@ -1031,6 +1046,10 @@ class Downshift extends Component {
     this.getRootProps.called = false
     this.getRootProps.refKey = undefined
     this.getRootProps.suppressRefError = undefined
+    // we do something similar for getMenuProps
+    this.getMenuProps.called = false
+    this.getMenuProps.refKey = undefined
+    this.getMenuProps.suppressRefError = undefined
     // we do something similar for getLabelProps
     this.getLabelProps.called = false
     // and something similar for getInputProps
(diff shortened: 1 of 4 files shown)


src/downshift.js
```diff
@@ -983,6 +990,10 @@ class Downshift extends Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
```
kentcdodds: Should we add this to a `componentDidMount` as well?

alexandernanberg: Yep that would make sense!

alexandernanberg:
Okay, I'll do that! Hopefully I'll have time before I go on vacation on Wednesday. 

That would be a great idea, what do you think of using something like https://www.npmjs.com/package/warning? But I'll probably do this in another PR in that case

kentcdodds:
Sounds perfect :+1:

alexandernanberg:
Added tests and updated the `getMenuProps` section in the `README.md` now. Basically just copied the tests for `getRootProps` and changed them where it was needed

src/downshift.js
```diff
@@ -983,6 +994,10 @@ class Downshift extends Component {
...
+    if (this.getMenuProps.called && !this.getMenuProps.suppressRefError) {
```
kentcdodds: This should probably be:

```javascript
if (
  this.getMenuProps.called &&
  !this.getMenuProps.suppressRefError &&
  preval`module.exports = process.env.BUILD_REACT_NATIVE !== 'true'`
) {
```

so this code can be eliminated for native. Same with the above.

alexandernanberg: I’ll fix that tomorrow! Hmm any reason why we dont do that for ’getRootProps’ as well? Im not so familiar with RN but why wouldnt we want to validate the ref?

kentcdodds: Yeah, I just checked and the only place we're using `_rootNode` is for codepaths that react native doesn't need. So we could probably do the same thing for those.

I'm thinking about making [a custom macro](https://github.com/kentcdodds/babel-plugin-macros) that will shorten things for us/make these kinds of branches easier to maintain. But that'll be for later :)

alexandernanberg: Oh okay! I've pushed the latest changes now, and I think you should be able to use `--autosquash` to squash them together to one commit

I like the sound of that idea!

alexandernanberg:
Hmm this is super weird... the error says that preval is not defined. Tried adding istanbul ignore comments locally but still get the same issue also🤔

kentcdodds:
:-( I'm afraid I don't have time to look into this at the moment :-/

alexandernanberg:
No worries, I’m off a couple of days but will pick this up again when I get back!

alexandernanberg:
Looks like everything is working now, don't really know why `/* istanbul ignore if */` didn't work but `/* istanbul ignore next */` just before preval did 🤔

PR #533, opened 2018-08-03
Upgrade kcd-scripts

alexandernanberg:
**What**:
Upgrades `kcd-scripts` and makes necessary changes (mostly linting)

**Why**:
Keeping up-to-date

**How**:
Upgrade dep and run `npm run validate` to find  eslint errors etc. 

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged 
      
### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "flow-bin": "^0.74.0",
     "flow-coverage-report": "^0.5.0",
     "jest-dom": "^1.3.1",
-    "kcd-scripts": "^0.39.0",
+    "kcd-scripts": "^0.41.0",
     "npm-run-all": "^4.1.2",
     "preact": "^8.2.6",
     "preact-render-to-string": "^3.7.0",
@@ -92,6 +92,7 @@
   "eslintConfig": {
     "extends": "./node_modules/kcd-scripts/eslint.js",
     "rules": {
+      "import/no-useless-path-segments": "off",
       "import/no-unassigned-import": "off",
       "max-lines": "off",
       "no-eq-null": "off",
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -21,11 +21,13 @@ export default class App extends React.Component<Props, State> {
   }
 
   onChange = (selectedItem: Item) => {
-    selectedItem
+    // eslint-disable-next-line no-console
+    console.log('selectedItem', selectedItem)
   }
 
   onUserAction = (changes: StateChangeOptions<Item>) => {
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -92,6 +92,7 @@
   "eslintConfig": {
     "extends": "./node_modules/kcd-scripts/eslint.js",
     "rules": {
+      "import/no-useless-path-segments": "off",
```
alexandernanberg: Your fix on `eslint-config-kentcdodds` didn't publish (CI failing) so I added an temporary rule until the publish

test/basic.test.js
```diff
@@ -21,11 +21,13 @@ export default class App extends React.Component<Props, State> {
   }
 
   onChange = (selectedItem: Item) => {
```
alexandernanberg: Unsure what the original code even did, but I guess an console.log is fine?

test/basic.test.js
```diff
@@ -21,11 +21,13 @@ export default class App extends React.Component<Props, State> {
...
-    selectedItem
+    // eslint-disable-next-line no-console
+    console.log('selectedItem', selectedItem)
   }
 
   onUserAction = (changes: StateChangeOptions<Item>) => {
```
alexandernanberg: `onUserAction` isn't used anywhere AFAIK, so it should be safe to remove that?

test/custom.test.js
```diff
@@ -36,7 +36,8 @@ export default class App extends React.Component<Props, State> {
   }
 
   onChange = (selectedItem: Item) => {
```
alexandernanberg: Same here, can't see that it's being used somewhere?

test/basic.test.js
kentcdodds: I think this file is used for TypeScript tests, let's leave it :+1:

test/custom.test.js
kentcdodds: Used for TypeScript, so I think we can leave it. I kinda leave this stuff up to the TS maintainers.

alexandernanberg:
Yea I really like it, it's a bit bloated with the preact bundles though but I don't think it's worth doing anything about it right now. 

Sure, will do 👍

alexandernanberg:
I've added a note about the size snapshot file. Writing good docs isn't my strong side so feel free to edit it if you like 😅 Wasn't sure where it made most sense to place it either, but I think it kind of works where it is right now.

PR #534, opened 2018-08-07
Improved typings for getMenuProps

franklixuefei:
Fixes the typing part of #490

**Why**: The typings for `getMenuProps` was too broad.

**How**:
Created interfaces for the types of parameters of `getMenuProps`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged
      
### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -111,7 +111,7 @@ export interface GetToggleButtonPropsOptions
 export interface GetMenuPropsOptions<Item>
   extends Record<string, any> {
   refKey?: string;
-};
+}
 
 export interface GetMenuPropsOtherOptions<Item> {
   suppressRefError?: boolean;
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -108,12 +108,12 @@ export interface GetLabelPropsOptions
 export interface GetToggleButtonPropsOptions
   extends React.HTMLProps<HTMLButtonElement> {}
 
-export interface GetMenuPropsOptions<Item>
-  extends Record<string, any> {
+export interface GetMenuPropsOptions {
   refKey?: string;
-}
+  ['aria-label']?: string;
+};
 
-export interface GetMenuPropsOtherOptions<Item> {
+export interface GetMenuPropsOtherOptions {
   suppressRefError?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


franklixuefei:
@kentcdodds Thanks! Who would you recommend for this review?

kentcdodds:
Hopefully they're watching the repo and will get to it when they have an opportunity... We'll give them a day or two.

typings/index.d.ts
```diff
@@ -105,24 +105,29 @@ export interface GetInputPropsOptions
...
   extends React.HTMLProps<HTMLLabelElement> {}
 
-export interface getToggleButtonPropsOptions
+export interface GetToggleButtonPropsOptions
   extends React.HTMLProps<HTMLButtonElement> {}
 
-interface OptionalExtraGetItemPropsOptions {
-  [key: string]: any
+export interface GetMenuPropsOptions<Item>
+  extends Record<string, any> {
```
pbomb: This doesn't seem to be quite right. I also don't understand what the `<Item>` generic type is doing here as it doesn't appear to be used.

From looking over the docs, this is what I would expect this type to look like:
```
export interface GetMenuPropsOptions {
  refKey?: string
  aria-label?: string
}

File: typings/index.d.ts [hunk 2]
```diff
@@ -105,24 +105,29 @@ export interface GetInputPropsOptions
...
+  refKey?: string;
+}
+
+export interface GetMenuPropsOtherOptions<Item> {
```
pbomb: Let's remove the `<Item>` part of this as it isn't adding any value.

typings/index.d.ts
```diff
@@ -105,24 +105,29 @@ export interface GetInputPropsOptions
...
   item: Item
 }
 
 export interface PropGetters<Item> {
   getRootProps: (options: GetRootPropsOptions) => any
-  getToggleButtonProps: (options?: getToggleButtonPropsOptions) => any
+  getToggleButtonProps: (options?: GetToggleButtonPropsOptions) => any
   getLabelProps: (options?: GetLabelPropsOptions) => any
-  getMenuProps: (options?: {}) => any
+  getMenuProps: (options?: GetMenuPropsOptions<Item>, otherOptions?: GetMenuPropsOtherOptions<Item>) => any
```
pbomb: Let's remove the `<Item>` occurrences on this line.

typings/index.d.ts
```diff
@@ -105,24 +105,29 @@ export interface GetInputPropsOptions
 export interface GetLabelPropsOptions
   extends React.HTMLProps<HTMLLabelElement> {}
 
-export interface getToggleButtonPropsOptions
+export interface GetToggleButtonPropsOptions
```
pbomb: :+1:

typings/index.d.ts
franklixuefei: You are completely right. I kind of copied the interface from `GetMenuItemOptions` which has the generic type parameter and I forgot to remove it.

franklixuefei:
I haven't figured out why it's failing.

PR #538, opened 2018-08-10
feat(scrollIntoView): add new prop to override built-in scrollIntoView

### LINKED ISSUE #537: Provide option to customize scrollIntoView.block
- `downshift` version: 2.0.20
- `node` version: 10.5.0
- `npm` (or `yarn`) version: yarn 1.7.0

**Relevant code or config**

```js
<Downshift defaultHighlightedIndex={12} />
```

**What you did**:

Instantiate Downshift with `defaultHighlightedIndex`

**What happened**:

It scrolled to the 13th element, but it ended at the bottom of the Select (because `block: 'nearest'` [here](https://github.com/paypal/downshift/blob/e34f5bcbcfa6024cbaf73dcc0ee1ae65cbfdc029/src/utils.js#L29-L33)).

**Reproduction repository**:

/

**Problem description**:

I would like the highlighted element to be at the top of the Select, instead of being at the bottom.

**Suggested solution**:

I would like to provide an option to change `scrollIntoview.block`. Are you fine with me opening a PR? Thanks!

kentcdodds:
**What**: feat(scrollIntoView): add new prop to override built-in scrollIntoView

**Why**: Closes #537

**How**: React makes this easy. I leverage `defaultProps` to set our own implementation of `scrollIntoView` as the `scrollIntoView` prop and then call it via `this.props.scrollIntoView` so users of downshift can override it to be whatever they want.

This means that I had to change the test slightly because the `scrollIntoView` function is no longer called directly on `utils` so we can't spy on it. Just swapped that to use mocking and it's working fine.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged
      
### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,23 +1,23 @@
 {
   "preact/dist/downshift.cjs.js": {
-    "bundled": 55985,
-    "minified": 23755,
-    "gzipped": 7059
+    "bundled": 56030,
+    "minified": 23804,
+    "gzipped": 7068
   },
   "preact/dist/downshift.umd.min.js": {
-    "bundled": 58032,
-    "minified": 21519,
-    "gzipped": 6612
+    "bundled": 58079,
+    "minified": 21567,
+    "gzipped": 6626
   },
   "preact/dist/downshift.umd.js": {
-    "bundled": 58668,
-    "minified": 21867,
-    "gzipped": 6747
+    "bundled": 58715,
+    "minified": 21915,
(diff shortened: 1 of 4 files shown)


kentcdodds:
Good thinking @eliperkins! It'd be cool to get official support for scrolling to the highlighted index with ReactNative!

franklixuefei:
We probably also needs to change the type definition for this. Will do this in a bit.

PR #539, opened 2018-08-12
chore: upgrade kcd-scripts and enable size-snapshot

alexandernanberg:
**What**:
Upgrade `kcd-scripts` and enable size snapshots

**Why**:
Keeping same functionality

**How**:
Add `--size-snapshot` argument on build

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged
      
### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -9,57 +9,52 @@
     "minified": 21567,
     "gzipped": 6626
   },
+  "dist/downshift.umd.min.js": {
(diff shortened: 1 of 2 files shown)


alexandernanberg:
Yeah probably, I wonder if we should add the `build` script in the `precommit` hook to keep it always updated🤔

kentcdodds:
It is 😅🤷‍♂️

PR #541, opened 2018-08-13
Add example select item that start with key pressed

AmyScript:
**What**:

Added an example that uses setHighlightedIndex() to highlight an item in the dropdown that starts with the letter that the user presses.  Also made a very unique dropdown UI that showcases the flexibility of Downshift.

**Why**:

There doesn't seem to be any example of using setHighlightedIndex() and the implementation is a bit complicated.

**How**:

**Checklist**:
- [X] Documentation
- [ ] Tests N/A
- [X] Ready to be merged
- [X] Added myself to contributors table


AmyScript:
Sure!

AmyScript:
@kentcdodds Here's the PR to that repo: https://github.com/kentcdodds/downshift-examples/pull/8

Let me know if there's anything else I need to add :)

PR #542, opened 2018-08-13
chore: inline build variables with custom simple macro

### LINKED ISSUE #540: Improve build-conditional logic code with a custom macro
There are a few places in downshift that we do stuff like this:

```javascript
preval`module.exports = process.env.BUILD_REACT_NATIVE !== 'true'`
```

For example:

https://github.com/paypal/downshift/blob/4b659d083f3ec0c7709b9bc9a060bab7438fdbc4/src/downshift.js#L249-L255

I'm pretty sure we could write a custom [macro](https://github.com/kentcdodds/babel-plugin-macros) to make this simpler:

```javascript
import {isReactNative} from './env.macro'

// ...

  scrollHighlightedItemIntoView() {
    /* istanbul ignore else (react-native) */
    if (!isReactNative) {
      const node = this.getItemNodeFromIndex(this.getState().highlightedIndex)
      this.props.scrollIntoView(node, this._rootNode)
    }
  }
```

I'm fairly certain all we'd need to do is replace `isReactNative` with `true` or `false` and rely on rollup's dead code elimination, or perhaps add a babel plugin for dead code elimination. Either that, or we could just eliminate the code ourselves, though I feel like that would make it harder to maintain, so I'd prefer the alternatives.

Andarist:
fixes #540

Some snapshot sizes are bigger, some are smaller. This PR should remove some bytes in all the cases, gonna investigate this in a moment - I suspect those differences might be caused by newer versions of downshift's deps and not downshift itself, but cant be sure right now.

EDIT:// yeah, running the build on master with fresh deps updates the snapshot, so size regressions are not caused by this PR

@TrySound would it be possible to output additional snapshot sizes for the lib itself with all deps as external? Both numbers are valuable (with externals bundled in and without them)

### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -2,7 +2,7 @@
 
 import React, {Component} from 'react'
 import PropTypes from 'prop-types'
-import preval from 'preval.macro'
+import {isPreact, isReactNative} from './is.macro'
 import setA11yStatus from './set-a11y-status'
 import * as stateChangeTypes from './stateChangeTypes'
 import {
@@ -232,7 +232,7 @@ class Downshift extends Component {
 
   scrollHighlightedItemIntoView() {
     /* istanbul ignore else (react-native) */
-    if (preval`module.exports = process.env.BUILD_REACT_NATIVE !== 'true'`) {
+    if (!isReactNative) {
       const node = this.getItemNodeFromIndex(this.getState().highlightedIndex)
       this.props.scrollIntoView(node, this._rootNode)
     }
@@ -574,7 +574,7 @@ class Downshift extends Component {
     ...rest
   } = {}) => {
     const {isOpen} = this.getState()
-    const enabledEventHandlers = preval`module.exports = process.env.BUILD_REACT_NATIVE === 'true'`
+    const enabledEventHandlers = isReactNative
       ? /* istanbul ignore next (react-native) */
         {
           onPress: callAllEventHandlers(onPress, this.button_handleClick),
@@ -672,7 +672,7 @@ class Downshift extends Component {
     let eventHandlers = {}
 
     /* istanbul ignore next (preact) */
-    if (preval`module.exports = process.env.BUILD_PREACT === 'true'`) {
+    if (isPreact) {
       onChangeKey = 'onInput'
     } else {
       onChangeKey = 'onChange'
@@ -692,7 +692,7 @@ class Downshift extends Component {
     }
 
     /* istanbul ignore if (react-native) */
-    if (preval`module.exports = process.env.BUILD_REACT_NATIVE === 'true'`) {
+    if (isReactNative) {
       eventHandlers = {
         ...eventHandlers,
         onChangeText: callAllEventHandlers(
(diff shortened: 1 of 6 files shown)


TrySound:
@Andarist umd is a result with bundled dependencies, esm is a result without node_modules treated as externals.

TrySound:
Lockfile could make size snapshot more stable. Yarn is the best at this. Npm sometimes works quite weird with locking dependencies.

Andarist:
> @Andarist umd is a result with bundled dependencies (except peers usually), esm is a result without node_modules treated as externals.

Oh, makes a lot of sense that 😅 I've come to wrong conclusion when I've pinged u, thanks for clarifying this!

The true snapshot diff for this PR can be seen [here](https://github.com/paypal/downshift/compare/e3d23e3d585f19ed8181fdb4ac36a701685988fa...baa76f387b88f37970fad0a1aba40c604ba212f5#diff-29fd455fe4a1c8d19c5e5b1420ce14fa)

src/downshift.js
```diff
@@ -248,7 +248,7 @@ class Downshift extends Component {
 
   scrollHighlightedItemIntoView() {
     /* istanbul ignore else (react-native) */
-    if (preval`module.exports = process.env.BUILD_REACT_NATIVE !== 'true'`) {
+    if (!isReactNative) {
```
kentcdodds: Out of curiosity... With this new macro, can we remove these Istanbul ignore comments?

Andarist: Maybe, i kinda dont know how those comments work **exactly**, not sure how to check if they can be removed (like i dont know what kind of outcome should i expect and where)

kentcdodds: Try to remove the comment, then run `npm run test -- --coverage` and see if we still get a 100% coverage report.

Andarist:
Ok, so the coverage should be fixed now (I can confirm that istanbul comments cannot be removed). I've also rebased the work on top of the master so there is no conflict now.

src/downshift.js
kentcdodds: Rats! Istanbul uses a babel plugin to instrument the code for coverage and I've found it's been a little tough to make it play nicely with other babel plugins at times :-(

Andarist: https://github.com/paypal/downshift/pull/542#issuecomment-412681801

it seems to work ok now, CI reports TS task failing, will have to investigate this in following days

Andarist:
I've check what is going with Travis - the same validate script is working for me on my machine (well... actually I get some linting errors cause by eslint's `import/no-unresolved`). 

The build is failing with 137 code and based on found info in the internet the most likely cause is Out of memory error. Not sure what we should do about it right now.

kentcdodds:
I think I've fixed it with this: https://github.com/paypal/downshift/commit/d733cf74f94e6a5a8eabb9ae820e753b8d952b4b

kentcdodds:
There could be something we could do in a way that's built-into kcd-scripts. Maybe we could use jest-worker for the validate script: https://github.com/facebook/jest/tree/master/packages/jest-worker

kentcdodds:
I _think_ this is good to go. Could you rebase onto master to be sure? Thanks!

Andarist:
Rebased and green on CI, thanks! 🎉

Issue #452, opened 2018-06-04
Where can downshift be improved?


kentcdodds:
Hey friends! I'm looking for feedback. Where can downshift be improved? This is an open-ended question and I'm looking for ideas. How can we improve the docs? What can we improve about the API? How can the implementation be improved?

A few ideas I have:

1. A legit website with an easier way to add or contribute to examples (I have some ideas here... Maybe it could just be a repo that could be loaded into codesandbox which we'd publish via netlify... Very easy to contribute to via codesandbox's github integration).
2. Some auxiliary component packages that implement a simpler, more opinionated and intrinsically useful autocomplete/dropdown/etc. on top of the downshift implementation.
3. Release 2.0.0 which will remove the `breakingChanges` prop and also bring improved support for TypeScript and Flow.

Any other ideas? Remember that downshift itself is intended to be extremely flexible, so if what you're looking for is a simpler API, then consider whether you could build a component on top of downshift that exposes that API. If that's possible then let's do it and publish that!

TrySound:
Remove `jsnext:main` field in the next major. It was initiated by rollup and deprecated (just with an issue) in it in favour of `module` field. It's better to not provide dead stuff which others copy in their projects.

kentcdodds:
Cool! Go ahead and make a PR for that whenever 👌

9jaswag:
I used downshift for the first time today and I had a need to find out what element was changed in my `onChange` method. I don't know if there's already a way to get this info but I had to get the `id` from the `stateAndHelpers` and add `-input` to the end of the string before going ahead to find the element in the DOM.
If a better solution for this exists, I'd like to know. If not, it'll be cool to have. (I'll be glad to have my first open source contribution 😄  )

kentcdodds:
Hi @9jaswag, could you describe your use case further? Why do you need to know what element was changed? And what do you mean by "element was changed?"

9jaswag:
@kentcdodds I have multiple downshift input fields on a page and in the `onChange` method, I'm setting `selectedItem` to state. I want to use the same `onChange` method for all the for all the input fields on the page. 
From my little React knowledge, I do something like this `this.setState({ [event.target.name]: event.target.value }) and it sets the matching item in the state.
So, I wanted to do the same thing and I had to get the "name" attribute of the input field whose value was just selected. I ended up doing something like this:
```sh
onChange(selectedItem, stateAndHelpers) {
[...]

PR #544, opened 2018-08-13
chore: remove runtime validations for prod builds

Andarist:
Exploration of https://github.com/paypal/downshift/issues/452#issuecomment-412457953

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,59 +1,59 @@
 {
   "preact/dist/downshift.cjs.js": {
-    "bundled": 48810,
-    "minified": 21477,
-    "gzipped": 6159
+    "bundled": 49253,
+    "minified": 21627,
+    "gzipped": 6187
   },
   "preact/dist/downshift.umd.min.js": {
-    "bundled": 58137,
-    "minified": 21466,
-    "gzipped": 6629
+    "bundled": 55922,
+    "minified": 20357,
+    "gzipped": 6315
   },
   "dist/downshift.umd.min.js": {
-    "bundled": 63167,
-    "minified": 23204,
-    "gzipped": 7404
+    "bundled": 58774,
+    "minified": 21562,
+    "gzipped": 6869
   },
   "preact/dist/downshift.umd.js": {
-    "bundled": 58773,
-    "minified": 21814,
-    "gzipped": 6764
+    "bundled": 58945,
+    "minified": 21771,
+    "gzipped": 6785
   },
   "dist/downshift.cjs.js": {
-    "bundled": 50161,
-    "minified": 22557,
-    "gzipped": 6351
+    "bundled": 50604,
+    "minified": 22707,
+    "gzipped": 6380
   },
   "dist/downshift.umd.js": {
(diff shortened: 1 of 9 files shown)


.size-snapshot.json
```diff
@@ -1,59 +1,59 @@
...
-    "bundled": 55770,
-    "minified": 23596,
-    "gzipped": 7003,
+    "bundled": 55990,
+    "minified": 23767,
+    "gzipped": 7026,
     "treeshaked": {
       "rollup": {
-        "code": 21245,
+        "code": 20136,
```
Andarist: those treeshaked values look the best - 1kb removed

Andarist:
This will require changing the tests - can work on that, but not today.

alexandernanberg:
IMO it would be nice if we'd have a `isProduction` macro or something similar, but I guess that is dependent on #542

alexandernanberg:
Oh that's true, totally forgot about that 😅

Andarist:
I'm using `console.error` now (instead of throwing errors) & updated the tests accordingly

kentcdodds:
I just merged the other PR and now there are merge conflicts :-( Sorry

Andarist:
Worry not, I've rebased this branch :)

src/downshift.js
```diff
@@ -1090,16 +1115,19 @@ function validateGetRootPropsCalledCorrectly(element, {refKey}) {
...
   } else if (!isComposite && refKeySpecified) {
-    throw new Error(
+    // eslint-disable-next-line no-console
+    console.error(
       `downshift: You returned a DOM element. You should not specify a refKey in getRootProps. You specified "${refKey}"`,
     )
   }
   if (!getElementProps(element)[refKey]) {
-    throw new Error(
+    // eslint-disable-next-line no-console
```
kentcdodds: Let's disable the `no-console` rule entirely. It's kinda silly at this point.

Thanks!

src/__tests__/downshift.misc.js
```diff
@@ -7,6 +7,16 @@ import ReactDOM from 'react-dom'
...
+const oldError = console.error
+
+beforeEach(() => {
+  console.error = jest.fn()
+})
+
+afterEach(() => {
+  console.error = oldError
+})
+
```
kentcdodds: I think this could be easier:

```javascript
beforeEach(() => jest.spyOn(console, 'error').mockImplementation(() => {}))
afterEach(() => console.error.mockRestore())
```

src/downshift.js
Andarist: IMHO it still makes sense to have this rule and disable it on per call site basis - this guards from unintended logs being shipped by accident, but still allows for granular control for cases when we actually want those console logs

at least this is what I always do at work

let me know what do you think about it

src/__tests__/downshift.misc.js
Andarist: done

src/downshift.js
kentcdodds: That's fine with me 👌

PR #545, opened 2018-08-13
chore: use stateChangeTypes namespace module internally instead of st…

Andarist:
This saves around 50 bytes in prod build, each internal reference to stateChangeTypes can be "inlined" by rollup as a simple variable - instead of being a property of a **nested** object.

While this isn't impressive I've made a test for another (sadly breaking change) - removing `stateChangeTypes` from the `Downshift` class entirely & exporting them from the root entry instead, like this:
```js
import * as stateChangeTypes from './stateChangeTypes'
export { stateChangeTypes }
```

and this gives far better results:
```diff
   "dist/downshift.esm.js": {
-    "bundled": 88000,
-    "minified": 35472,
-    "gzipped": 10364,
+    "bundled": 88140,
+    "minified": 35481,
+    "gzipped": 10445,
     "treeshaked": {
       "rollup": {
-        "code": 22990,
+        "code": 22251,
         "import_statements": 37
       },
       "webpack": {
-        "code": 23997
+        "code": 23256
       }
     }
   }
```

and this is 741 bytes saved for webpack for the most common use case (not needing `stateChangeTypes`). 

It would be even better than that to create a secondary entry point (`import * as stateChangeTypes from 'downshift/state-change-types'`) so people wanting to use a single state change type wouldnt have to import all of them.

I've also done **separate** test which builds on top of this PR - I've replaced all stateChangeTypes with increasing integers and this allowed minifier to completely inline those numbers at each "usage site" which lead to **nice** savings:
```diff
   "dist/downshift.esm.js": {
-    "bundled": 88120,
-    "minified": 35508,
-    "gzipped": 10444,
+    "bundled": 87643,
+    "minified": 35031,
+    "gzipped": 10359,
     "treeshaked": {
       "rollup": {
-        "code": 22934,
+        "code": 22048,
         "import_statements": 37
       },
       "webpack": {
-        "code": 23941
+        "code": 23055
       }
     }
   }
```
those are independent from the previous test.

And obviously we can generate strings or numbers depending on NODE_ENV (with the help of macros)


Andarist:
What do u think about the two tests that I have made?

kentcdodds:
I like them, but we'll need to save the one for a breaking change. Let's do the other now though (via a macro I think). Thanks!

PR #548, opened 2018-08-14
chore: wrap stateChangeTypes in NODE_ENV and make them numbers for pr…

Andarist:
…oduction

Implemented what has been discussed in https://github.com/paypal/downshift/pull/545

- [x] Ready to be merged

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,59 +1,59 @@
 {
   "preact/dist/downshift.cjs.js": {
-    "bundled": 49253,
-    "minified": 21627,
-    "gzipped": 6187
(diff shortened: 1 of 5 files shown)


chrisdhanaraj:
@kentcdodds Just a quick comment! I haven't looked back here for a while, and I was being super lazy with a component and didn't want to log out the Downshift.stateChangeTypes; I ended up just using the string values and went from there. Was _super_ confused when I shipped it to production and it didn't work; would you be okay with a PR to update the README with both a mapping of the stateChangeTypes and a warning about this?

kentcdodds:
Yeah, that's a good idea 👍

PR #556, opened 2018-08-20
add missing 'suppressRefError' option typings

rifler:
**What**:
Missing TypeScript (TS) typings added - second argument to `getRootProps` and `getMenuProps` functions

**Checklist**:

- [ ] Documentation
- [ ] Tests N/A
- [x] Ready to be merged
      

README.md
```diff
@@ -495,6 +496,12 @@ node you render in downshift). Internally we use
 so if you use that package then you wont be adding any additional bytes to your
 bundle :)
 
+### suppressRefError
```
kentcdodds: This is excluded from the docs intentionally. It's a pretty far edge case so I don't think it's worth including in the docs.

rifler:
fixes done
I just noticed that there is very similar PR - https://github.com/paypal/downshift/pull/534
They overlap a bit

rifler:
@kentcdodds LGTM, I'll rebase my PR after merging #534

rifler:
done
Do you run prettier/linters on precommit? It's a bit hard to manage semicolons manually

kentcdodds:
Yes... Do you have dependencies installed?

kentcdodds:
I'm not sure why the precommit script is not running for you. It should be automatically generating the contributors table in the README as well as running prettier on all changed files... 🤔

rifler:
It's because I didn't install npm deps, as you mentioned earlier. It was obvious, but I didn't catch it :) Now it is ok.
About CI I meant the following - CI checks are more reliable, e.g. if someone (like me) forget to install deps (or use --no-verify), than CI fails
For example my PR contains style fixes (semi removal) from previous contributor
https://github.com/paypal/downshift/pull/556/files#diff-494415c3af35245ecae73f5d42d8b331R117
I suppose it is because he forgets to install deps too.
Also we could avoid errors like this - https://github.com/paypal/downshift/commit/5acaf862e39b14afc422d5f2c38e334d1d9ad63d

kentcdodds:
How do you suggest we update the pull request someone else opened? I know it's possible, but it'd be a bit more work than I would like to do myself.

kentcdodds:
Oh, we definitely do that via the `validate` script. I'm not sure why that last one got through. I probably merged it before the build was finished or ignored the failing build because we'd been having intermittent build failures anyway. Either way, it was my fault 😅

Issue #561, opened 2018-08-22
Idea: Overload getItemProps for Advanced Props


nathanforce:
- `downshift` version: 2.1.5

How would you feel about a PR implementing the added functionality below to `getItemProps`? These props are simple for Downshift to calculate internally but can be difficult or impossible to determine in userland.

```javascript
const props = getItemProps(({ index, itemCount, isSelected, isHighlighted }) => ({
    style: {
        fontWeight: isSelected ? 'bold' : 'normal',
        background: isHighlighted ? 'purple' : 'transparent',
    }
}))
```

Let me know what you think! Happy to make the PR if we think people might want this.

Thanks!

kentcdodds:
To be clear, prop getters could accept an object of props or a function that accepts the same thing as children and returns an object of props.

If that's true, yes that sounds great. We'll need docs, tests, and implemention. Do you want to do that?

nathanforce:
Not quite.  The function would accept the same thing plus some other helpful props like `isHighlighted` or `isSelected` that are trivial for Downshift to determine internally.

For example..

```
isSelected: this.getState().selectedItem === item,
isHighlighted: this.getState().highlightedIndex === index,
itemCount: this.items.length,
```

kentcdodds:
Ah, gotcha.

Yes, that would be acceptable I think. Though the itemCount may not be accurate. The items are reset in every render and added to a you call `getItemProps`, so for the first item the itemCount would be 0...

In addition, I would like all other prop getters to support the function capabilities!

nathanforce:
Okay I can do that.

Do you have any ideas on which values each prop getter would expose?

kentcdodds:
Let's start with `this.getStateAndHelpers()` for all of them + `isHighlighted` and `isSelected` for `getItemProps`

nathanforce:
@kentcdodds This is what I've got for `getItemProps` currently.  Wanted to get some feedback before doing the others.

Open discussion:

- This introduces a second argument to `getItemProps`. I could not come up with a way to do the overloading, as `{ item, index? }` are required to determine `isSelected` and `isHighlighted`
- What should happen if you pass an argument other than `{ item, index? }` while also passing a transform function? Should we pass through? Do we merge the results of the two? What about conflicts?


![image](https://user-images.githubusercontent.com/6694194/44596009-c95d4280-a798-11e8-8f2f-b773e3be6a3c.png)

kentcdodds:
[...]

PR #563, opened 2018-08-24
Feature: Add mapStateToProps function to all Prop Getters

nathanforce:
Adds `mapStateToProps` functions to each Prop Getter.

All are called with `stateAndHelpers`.  `getItemProps` is called with 

```
{
    ...stateAndHelpers,
    isSelected,
    isHighlighted,
    index
}
```

See #561 

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


nathanforce:
Thanks Kent

1.  I think I agree here. I was on the fence when adding them anyhow so I'll update those.
2.  My thought was that the consistency across Prop Getters was more valuable for consumers, and since `getItemProps` mandates that item is passed first I opted for all getters to take the function as their last arg.  This also opens us up for adding values to the other Getters in the future, same as we did here for `getItemProps`.
3. More documentation than [I already did](https://github.com/paypal/downshift/pull/563/files#diff-04c6e90faac2675aa89e2176d2eec7d8R569)? Happy to add more if you think its not clear.

README.md
```diff
@@ -1047,21 +1049,22 @@ Thanks goes to these people ([emoji key][emojis]):
...
...0?v=4" width="100px;"/><br /><sub><b>Jared Forsyth</b></sub>](http://jaredforsyth.com)<br />[🤔](#ideas-jaredly "Ideas, Planning, & Feedback") [📖](https://github.com/paypal/downshift/commits?author=jaredly "Documentation") | [<img src="https://avatars1.githubusercontent.com/u/8162598?v=4" width="100px;"/><br /><sub><b>Jack Moore</b></sub>](https://github.com/jtmthf)<br />[💡](#example-jtmthf "Examples") | [<img src="https://avatars1.githubusercontent.com/u/2762082?v=4" width="100px;"/><br /><sub><b>Travis Arnold</b></sub>](http://travisrayarnold.com)<br />[💻](https://github.com/paypal/downshift/commits?author=souporserious "Code") [📖](https://github.com/paypal/downshift/commits?author=souporserious "Documentation") | [<img src="https://avatars0.githubusercontent.com/u/1045233?v=4" width="100px;"/><br /><sub><b>Marcy Sutton</b></sub>](http://marcysutton.com)<br />[🐛](https://github.com/paypal/downshift/issues?q=author%3Amarcysutton "Bug reports") [🤔](#ideas-marcysutton "Ideas, Planning, & Feedback") | [<img src="https://avatars2.githubusercontent.com/u/244704?v=4" width="100px;"/><br /><sub><b>Jeremy Gayed</b></sub>](http://www.jeremygayed.com)<br />[💡](#example-tizmagik "Examples") |
```
kentcdodds: Hmmm... Somehow this table got all messed up... Maybe by your editor? Could you revert these changes?

src/__tests__/downshift.get-button-props.js
```diff
@@ -71,6 +71,27 @@ test('getToggleButtonProps returns all given props', () => {
   )
 })
 
+test('getToggleButtonProps accepts a mapStateToProps function', () => {
+  render(
```
kentcdodds: Could we rewrite this (and the other similar tests) to work more like the one in `downshift.get-input-props`?

README.md
```diff
@@ -566,16 +566,18 @@ rather than applying them on the element yourself to avoid your props being
...
-| `getLabelProps`        | `function({})`    | returns the props you should apply to the `label` element that you render.                     |
-| `getMenuProps`         | `function({},{})` | returns the props you should apply to the `ul` element (or root of your menu) that you render. |
-| `getRootProps`         | `function({},{})` | returns the props you should apply to the root element that you render. It can be optional.    |
+| property               | type                               | description                                                                                                                                                                           |
+| ---------------------- | ---------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `getToggleButtonProps` | `function({}, mapStateToProps)`    | returns the props you should apply to any menu toggle button element you render.                                                                                                      |
```
kentcdodds: I'd really prefer the `getToggleButtonProps`, `getInputProps`, `getLabelProps`, `getMenuProps`, and `getRootProps` to accept an object _or_ `mapStateToProps` as the first argument. I'm fine with `getItemProps` requiring an object as the first argument and being slightly different than the rest. I'm thinking 99% of the time people will want one or the other but not both, so let's optimize for that case.

src/__tests__/downshift.get-item-props.js
```diff
@@ -178,6 +178,109 @@ test(`disabled item can't be selected by pressing enter`, () => {
...
 })
 
+test(`getItemProps mapStateToProps function provides isSelected`, () => {
+  render(
+    <Downshift defaultSelectedItem={0}>
+      {({getItemProps}) => (
+        <div>
+          <span
+            {...getItemProps({item: 0}, ({isSelected}) => {
+              expect(isSelected).toBe(true)
```
kentcdodds: We should probably use [expect.assertions](https://jestjs.io/docs/en/expect#expectassertionsnumber) for any of these tests that put assertions within callbacks to make sure those assertions run.

kentcdodds:
I think that's an acceptable workaround. Let's go ahead and close this. Thanks for all the work @nathanforce!

I think we should add you to the contributors table for the idea! Would you like to do that?

PR #570, opened 2018-08-31
Made types more accurate

franklixuefei:
**What**:
Made types more accurate.

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged
      
### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -105,13 +105,17 @@ export interface GetRootPropsOptions {
 }
 
 export interface GetInputPropsOptions
-  extends React.HTMLProps<HTMLInputElement> {}
+  extends React.HTMLProps<HTMLInputElement> {
+    disabled?: boolean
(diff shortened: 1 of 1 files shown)


kentcdodds:
Build failure:

```
[test:ts] test/custom.test.tsx(52,21): error TS2345: Argument of type '{ item: string; index: number; isSelected: boolean; }' is not assignable to parameter of type 'GetItemPropsOptions<string>'.
[test:ts]   Object literal may only specify known properties, but 'isSelected' does not exist in type 'GetItemPropsOptions<string>'. Did you mean to write 'selected'?
```

franklixuefei:
I see why it failed. Previously, that interface allowed any key value pairs, so “isSelected” was not treated as unrecognized even if it was not explicitly defined in that interface. I’ll go ahead and make the change

PR #573, opened 2018-09-05
Update React Native snapshot

samuelfullerthomas:
See: https://github.com/paypal/downshift/pull/451

Also re-add `onChange`. Question - should there be something in the README about the RN only tests? Is there? Did I miss it?

### DIFF
--- a/other/react-native/__tests__/__snapshots__/render-tests.js.snap
+++ b/other/react-native/__tests__/__snapshots__/render-tests.js.snap
@@ -19,7 +19,6 @@ exports[`renders with React Native components 1`] = `
     onBlur={[Function]}
     onChange={[Function]}
     onChangeText={[Function]}
-    onKeyDown={[Function]}
     underlineColorAndroid="transparent"
     value=""
   />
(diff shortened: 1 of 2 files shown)


samuelfullerthomas:
@kentcdodds

other/react-native/__tests__/__snapshots__/render-tests.js.snap
```diff
@@ -19,7 +19,6 @@ exports[`renders with React Native components 1`] = `
     onBlur={[Function]}
     onChange={[Function]}
     onChangeText={[Function]}
-    onKeyDown={[Function]}
```
kentcdodds: Do we not need this?

samuelfullerthomas: Nope!

See https://github.com/paypal/downshift/pull/451#discussion_r215278302

samuelfullerthomas: This lists the valid props for RN Text Input - no onKeyDown: https://facebook.github.io/react-native/docs/0.56/view#props

PR #581, opened 2018-09-20
Actually make flow types available in npm packages

stereobooster:
**What**:

It seems I figured out how properly distribute Flow types in npm

**Why**:

**How**:

I saw this pattern in https://github.com/philipp-spiess/react-recomponent/tree/master/dist

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Ready to be merged
      

ZhangYiJiang:
The way I managed this was to add this to our `flowconfig` 

```
[libs]
node_modules/downshift/flow-typed/npm
``` 

I'm not entirely sure if adding the Flow libdefs to dist directly will work since I'm not sure if Flow is configured to look for libdefs in node_modules directly

stereobooster:
For every import/require `file.js` Flow checks `file.js.flow` first

silviuavram:
it's stale, to say the least. you can open an issue and we can follow up on that.

PR #586, opened 2018-09-26
Fix React Native onBlur crash

audiolion:
**What**: Check for existence of `this.props.environment.document` before trying to access further properties on `document`

**Why**: `onBlur` when called in react native will try to access `this.props.environment.document.activeElement`, and `environment.document` does not exist in React Native, resulting in Downshift crashing RN.

**How**: Added a failing test case to showcase the bug, then added a change that fixes the failing test

**Checklist**:

- [N/A] Documentation
- [x] Tests
- [ ] Ready to be merged
      

audiolion:
I am not really sure how to capture the error in a failing test, but you can see the error getting logged out as the test passes:

https://travis-ci.org/paypal/downshift/builds/433607691#L649

image

I tried using `expect(() => firstArg.onBlur(fakeEvent)).not.toThrow()` but the error doesn't bubble up from that event, it occurs after the `internalSetTimeout` occurs

kentcdodds:
That's fine. Thanks!

Issue #589, opened 2018-10-04
[Enhancement][Accessibility] When Listbox is closed, Up/Down should also highlight last/first option 


silviuavram:
- `downshift` version: 2.2.3
- `node` version: 8.11.3
- `npm` (or `yarn`) version: npm, 5.6.0

**Enhancement description**:
According to ARIA 1.1 guidelines related to the Combobox, when Listbox is closed and you hit Up / Down as focus is on the Input, the Listbox should open, and the highlighted option should be last / first item. More on that [here](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html#Textbox).

At the moment, the scenario just opens the Listbox, without highlighting anything.

Since I am using Downshift for a prototype in the Stardust project, I managed to fix it from our side, but I believe this accessibility enhancement should also be implemented here, in the source.

**Suggested solution**:
If Listbox is not open and Up/Down is pressed, highlight last/first option after opening the Listbox.

Now working on a PR with a proposed solution. Updating some unit tests to reflect the changes.

kentcdodds:
I'm pretty sure that if you set the `defaultHighlightedIndex` to `0` that's the behavior you'll get.

silviuavram:
https://github.com/paypal/downshift/pull/590

Related to your answer, if I do that, the dropdown will always open with first item highlighted.

I only want for it to 1) open with first highlighted when I press Arrow Down and 2) open with last highlighted when I press Arrow Up.

When opened at click / text input, behaviour should remain the same. Do you agree?

PR #590, opened 2018-10-05
[Enhancement] Added ARIA specification of First and Last option highlight

silviuavram:
Addresses https://github.com/paypal/downshift/issues/589

**What**:
Added highlight on first and last options when user opens the dropdown via Up/Down Arrows.

**Why**:
According to ARIA, if dropdown is collapsed and focus is on the text input:
- Arrow Down should open menu **and** highlight on option.
- Arrow Up should open menu **and** highlight last option.

**How**:

Modified `moveHighlightedIndex` in dropdown.js, specifically the `else` branch. There I open the menu, and then check if the opening action was made by Arrow up or down. I also check if there are any items in the dropdown.

If yes, I call `setHighlightedIndex` with either `itemCount - 1` or `0`, depending on the value of the event type (arrow up or arrow down), in order to comply to ARIA specifications.

I also modified some unit tests and end-to-end tests in order to reflect the changes. Also added 2 more end-to-end tests that test only this new behaviour.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      

src/__tests__/downshift.misc-with-utils-mocked.js
```diff
@@ -37,7 +37,7 @@ test('does not scroll from an onMouseMove event', () => {
   }
   const {queryByTestId} = render(<HighlightedIndexController />)
   const input = queryByTestId('input')
-  const item = queryByTestId('item-1')
+  const item = queryByTestId('item-2')
```
silviuavram: I managed to 'fix' this test, although I need some more insight on what it tests. I used that example as a story and I still could not figure out how it works and how I managed to break it. Some help, please?

kentcdodds: This is actually fine. Expected for the change that's been made.

src/downshift.js
```diff
@@ -247,7 +247,24 @@ class Downshift extends Component {
     if (this.getState().isOpen) {
       this.changeHighlightedIndex(amount, otherStateToSet)
     } else {
-      this.setHighlightedIndex(undefined, {isOpen: true, ...otherStateToSet})
+      // if menu was closed, will open it.
```
kentcdodds: In this case we know it's closed because this is the else case of an isOpen check. We can remove the comment.

kentcdodds:
Oh, and looks like some cypress tests are failing on CI: https://travis-ci.org/paypal/downshift/builds/437712230

silviuavram:
Thank you for the quick feedback.

Those tests are the ones I added, will take a look and find out what the problem is.

Also will address the other comments and come back to you once the PR is polished.

kentcdodds:
I'm going to manually merge this because I'm doing weird things. Standby...

kentcdodds:
Sweet. Hopefully I can get this released in the major release this afternoon.

kentcdodds:
Let's keep it open until it's actually released 👍

kentcdodds:
This has been released in 3.0.0!

PR #592, opened 2018-10-05
release 3.0.0

### LINKED ISSUE #591: Release 3.0.0
Alrighty, we've got a few breaking changes that need to happen:

- [x] Respect defaultInputValue when provided #467
- [x] inputValue is set to empty string on select of already selected item #512
- [x] [Enhancement][Accessibility] When Listbox is closed, Up/Down should also highlight last/first option #589

And I have a few things I want to do:

- [x] upgrade all deps
- [x] move from storybook to [docz](https://docz.site) (will make cypress tests simpler and we may be able to actually use it for a legit site in the future... maybe now?) (related: https://github.com/paypal/downshift/issues/499)

As always, there are a handful of bugs that need investigating (I just went through and cleaned up the backlog).

I am confident that we can get this done today hopefully.

kentcdodds:
**What**: All the changes needed for 3.0.0

**Why**: Closes #591

**How**: Step-by-step...


.travis.yml
```diff
@@ -10,13 +10,11 @@ node_js: '8'
...
   - npx cypress verify &
-  - cd stories
-  - npm install
-  - cd ..
 script:
   - kcd-scripts validate lint,build-and-test,test:cover
   - kcd-scripts validate test:ssr,test:cypress
-  - kcd-scripts validate test:flow:coverage,test:ts
+  # TODO: fix typescript/flow
+  # - kcd-scripts validate test:flow:coverage,test:ts
```
kentcdodds: Could I get a TypeScript maintainer to give this a look please? It's failing the build: https://travis-ci.org/paypal/downshift/builds/437768573

package.json
```diff
@@ -19,18 +19,18 @@
     "test:ts": "tsc --noEmit -p ./tsconfig.json",
     "test:flow": "flow",
     "test:flow:coverage": "flow-coverage-report",
-    "test:build": "jest --projects other/misc-tests other/react-native",
-    "test:cypress:dev": "npm-run-all --parallel --race storybook cy:open",
-    "pretest:cypress": "npm run storybook:build --silent",
-    "test:cypress": "start-server-and-test storybook:serve http://localhost:6006 cy:run",
+    "test:build": "jest --config other/misc-tests",
+    "// FIXME: test:build": "jest --projects other/misc-tests other/react-native",
```
kentcdodds: The react native tests are broken. Could a react native maintainer help figure out what's wrong? travis-ci.org/paypal/downshift/builds/437797785

.travis.yml
franklixuefei: @kentcdodds I did some research, and it turned out that this is a bug introduced by an outdated version of `@types/jquery` that is currently referenced by `cypress/cli` (https://github.com/cypress-io/cypress/issues/2363). They have merged the fix (by bumping up the version of `@types/jquery`, but not yet published that fix yet (https://github.com/cypress-io/cypress/pull/2369). I think in their next release, this fix will be published. However, before that, the quick and dirty fix for us is to do `npm i --save-dev @types/jquery@3.3.6`, or reference `@types/jquery@3.3.6` directly in `package.json`. What do you think?

package.json
franklixuefei: Seems like SFCs are treated like Stateful Components because it was looking for `render`...

.travis.yml
franklixuefei: If you think this fix is too hacky, then I'd say just go ahead and publish 3.0.0 without fixing this, until cypress publishes their next release.

PR #596, opened 2018-10-09
Add Ctrl+P/N as alternatives to Up/Down arrows

sdegutis:
**What**: Added Ctrl-P and Ctrl-N as aliases to Up/Down arrows for list navigation.

**Why**: Tens of thousands of programmers use these shortcuts in much of their software, and it's a nice little attention to detail that they will appreciate and which raises their perception of the quality of the software they're using when they find it supported.

**How**: I added a non-invaisive if-statement in the normalizeArrowKey() util function.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged
      
- [ ] Added myself to contributors table N/A (didn't want to download the repo just to do this)

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -257,6 +257,10 @@ function normalizeArrowKey(event) {
   if (keyCode >= 37 && keyCode <= 40 && key.indexOf('Arrow') !== 0) {
     return `Arrow${key}`
   }
+  if (keyCode.ctrlKey) {
+    if (key === 'n') return 'ArrowDown';
+    if (key === 'p') return 'ArrowUp';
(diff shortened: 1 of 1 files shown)


kentcdodds:
Hmmm... I'm not sure I'm in favor of this. It seems pretty non-standard to me...

darshanrampatel:
Agreed - I'd imagine that most people now use Ctrl + P for Print...

sdegutis:
Ah yes I forgot most devs use Windows and not Mac so this shortcut probably makes no sense in that context.

PR #608, opened 2018-10-17
Make Downshift class accept generic Item type

willcaul:
**What**:

Make Downshift class accept generic Item type (TypeScript)

**Why**:

`Downshift` was being declared as `DownshiftInterface<any>`, which prevents passing a generic `Item` type to `Downshift`. Was this being done for a particular reason?  With the existing typings I had to work around the missing generic argument by doing something like: `const TypedDownshift: DownshiftInterface<string> = Downshift;` and then using `<TypedDownshift>...</TypedDownshift>` in my JSX.  After this change I can simplify this to `<Downshift<string>>...</Downshift>` in my JSX.

**How**:

I exported the `Downshift` class with a generic type arg (that defaults to `any`).  I also deleted `DownshiftInterface` because it doesn't seem to have much use after this change, but I can add it back as an alias for the `Downshift` class if we are concerned with its removal being a breaking change.
(Note: JSX generics were introduced in TS 2.9.)

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged
      

typings/index.d.ts
```diff
@@ -201,10 +201,10 @@ export type ChildrenFunction<Item> = (
   options: ControllerStateAndHelpers<Item>,
 ) => React.ReactNode
 
-export type DownshiftInterface<Item> = React.ComponentClass<
+export class Downshift<Item = any> extends React.Component<
```
NMinhNguyen: Curious if this could've been the following for backwards compatibility?

```suggestion
export type DownshiftInterface<Item = any> = React.ComponentClass<
```

willcaul: @NMinhNguyen I think we want to export `Downshift` accepting a generic `Item` type (as was not the case prior to this PR).  But we could probably export `DownshiftInterface` as well for backwards compatibility.  Maybe like `export type DownshiftInterface<Item> = Downshift<Item>` or something along those lines?  Thoughts?

PR #611, opened 2018-10-21
types: Item can be null in onChange and onSelect

epeli:
**What**:

I noticed that the item in `onChange` and `onSelect` handlers can be sometimes null. In TS strict mode Typescript usually warns about these situations but here I hit a runtime error.

Here's an example where the `item` is null in `onChange`: 

https://codesandbox.io/s/ywzpvv3jjz

Just select a fruit from the list and see `ordered-examples/01-basic-autocomplete`.

**Why**:

Typescript typings should reflect the real Javascript types closely as possible. 

**How**:

Item is now an union with null in `onChance` and `onSelect` which forces Typescript users to check for it in strict mode to avoid the runtime error.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged
      
<strike>
The tests "kinda pass". I wasn't able to execute the test successfully even with a plain master branch without any of my changes.

From `npm run validate` I get 

```
[build-and-test] [react.cjs] /var/www/git/downshift/src/index.js → dist/downshift.cjs.js...
[test:cypress] events.js:183
[test:cypress]       throw er; // Unhandled 'error' event
[test:cypress]       ^
[test:cypress]
[test:cypress] Error: watch /var/www/git/downshift/doczrc.js ENOSPC
```

Also `npm run test:ts` complain about jquery types on the master branch but that's not related to the changes neither.
</strike>

Update: Weird. The tests pass on Travis 😃

### DIFF
--- a/test/basic.test.tsx
+++ b/test/basic.test.tsx
@@ -15,7 +15,7 @@ export default class App extends React.Component<Props, State> {
     items: ['apple', 'orange', 'carrot'],
   }
 
-  onChange = (selectedItem: Item) => {
+  onChange = (selectedItem: Item | null) => {
     console.log('selectedItem', selectedItem)
   }
 
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -38,11 +38,11 @@ export interface DownshiftProps<Item> {
   selectedItemChanged?: (prevItem: Item, item: Item) => boolean
   getA11yStatusMessage?: (options: A11yStatusMessageOptions<Item>) => string
   onChange?: (
-    selectedItem: Item,
+    selectedItem: Item | null,
     stateAndHelpers: ControllerStateAndHelpers<Item>,
   ) => void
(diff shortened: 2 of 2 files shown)


silviuavram:
null can stand for an object. maybe you are describing a TSLint rule for strict null checks.

closing this.

epeli:
Nope. Strict null checks is a core compiler feature which is more often recommended to enable than not. It's not a TSLint thing.

And in strict mode null cannot stand for objects.

Just enable `strictNullChecks` from the [playground](https://www.typescriptlang.org/play/) options and paste this:

```ts
const foo: {} = null
```

epeli:
Also this change has no effect for users who do not have `strictNullChecks` enabled but prevents runtime crashes to those who have.

silviuavram:
thanks for explaining this in detail, got it now.

Issue #621, opened 2018-11-09
Weird Voiceover behavior in latest version of Chrome


dstapleton92:
- `downshift` version: 1.31.16
- `node` version: 10.11.0
- `npm` (or `yarn`) version: yarn 1.10.1

Has anyone else noticed weird behavior when using Voiceover on the autocomplete suggestions in Chrome in macOS Mojave? Our autocomplete suggestions are read partially, and then read again. For example, if you type "apple" into the search field, and scroll down to a suggestion for "apple sauce", VO will read "apple apple sauce". Or for "chicken", scrolling to "chicken noodle soup" reads "chicken noo chicken noodle soup".

After some troubleshooting, I suspect there has been a change in behavior and the suggestions are being read directly by VO **in addition to the live region** that Downshift is updating with the A11y status message.

Note that this does not occur with Safari, and I'm not sure if this is something that changed in Chrome with the latest version, or with VO in Mojave, as I updated both around the same time. Can anyone confirm this behavior? Am I crazy or is this real?

Thanks!

silviuavram:
confirmed here as well, after tried to fix it from my end for about an hour.

silviuavram:
Reproduced it with other screen readers as well, only that Jaws and NVDA will narrate the autocomplete suggestion twice. Narrator just stops while narrating first and then narrates it again.

Issue comes from `getA11yStatusMessage` because it returns `itemToString(highlightedItem)` when you go through the suggestions in the list. Screen readers read the <li> option naturally, without us needing to explicitly set an `aria-live` message for it.

As I don't think anything changed in that area in `Downshift`, I am curious how did this issue just occurred? Was it always there and we did not notice?

A quick workaround would be to override `getA11yStatusMessage` with our own one via props, but we might also need to change the default one in `Downshift`, as it is not providing a correct behaviour. What do you guys think @kentcdodds @dstapleton92 ?

kentcdodds:
Interesting! Could you open a PR for the fix and I'll give it a look. Thank you so much for all the time you've spent helping make the corner of the internet that uses downshift more accessible 🤗

dstapleton92:
For our use case, overriding `getAllyStatusMessage` with a similar function that doesn't read the selected item was a no-go. The reason why is because Safari does **not** automatically read the selected item of the aria combobox, so this would exclude Safari users from being able to use a screen reader.

[...]

PR #626, opened 2018-11-25
Disable aria-activedescendant by default

dstapleton92:
**What**:
Addresses #621. Downshift places an `aria-activedescendant` attribute on the input when a highlighted index is defined. The problem with this is that some browsers instruct screen readers to read the active descendant aloud, while others do not. On browsers that read the descendant, the reading occurs at the same time that Downshift's own aria live region is read, leading to a "double-reading" effect: "apple sa apple sauce".

**Why**:
This change is necessary so that the default behavior of this utility will not result in double reading on some browsers.

**How**:
Added a config property called `useAriaActiveDescendant` with a default value of `false`. If a developer wants to re-enable the aria active descendant, they need to pass `{ useAriaActiveDescendant: true }` into `getInputProps`.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1003,6 +1003,17 @@
         "bug",
         "code"
       ]
+    },
+    {
+      "login": "dstapleton92",
+      "name": "dstapleton92",
+      "avatar_url": "https://avatars3.githubusercontent.com/u/11726925?v=4",
+      "profile": "https://github.com/dstapleton92",
+      "contributions": [
+        "code",
+        "doc",
+        "test"
+      ]
     }
   ]
 }
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ autocomplete/dropdown/select/combobox components</p>
 [![downloads][downloads-badge]][npmcharts] [![version][version-badge]][package]
 [![MIT License][license-badge]][license]
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-102-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-103-orange.svg?style=flat-square)](#contributors)
 [![PRs Welcome][prs-badge]][prs] [![Chat][chat-badge]][chat]
 [![Code of Conduct][coc-badge]][coc]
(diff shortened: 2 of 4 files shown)


kentcdodds:
Thanks! How much work would someone have to do if we didn't support the `useAriaActiveDescendant` option? I feel like it's unlikely that anyone would use that option and if they did it would be trivial to implement. What do you think?

kentcdodds:
Also, could you please figure out why CI is failing? It looks like a Cypress test failure, but I'm not sure what is going on there.

silviuavram:
I still have some mixed feelings about this way to fix the issue.

Firstly, ARIA clearly specifies that `aria-activedescendant=IDREF` should be added on the Edit Text (input). [spec](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html)
Secondly, having `aria-activedescendant` will ensure a screen reader behaviour that is consistent with other dropdowns.
Lastly, the fact that a browser-screen reader combination does not read is a browser/screen reader problem and we should not hotfix everything else with the `aria-live`.

That being said, wouldn't it be better to fix Safari specifically?
- remove the aria-live update for listbox options. This will solve the double-reading on the majority of browsers/screen readers.
- fix Safari some other way, like keeping the old behaviour specifically for Safari.. For instance, update aria-live with highlightOption toString only if we detect Safari. We can find a better fix, open for debate.

This way we don't have a breaking change in the DOM. I really don't think removing `aria-activedescendant` by default is a good idea. That attribute is used not only for screen reader option reading. There are many implementation that use it for CSS changes, for example. I think we should do our changes in the already open for change `getA11yStatusMessage`.

kentcdodds:
I think I like what you're saying @silviuavram. I'm interested in seeing what you're suggesting. Would you like to open a pull request? We can still include @dstapleton92 on the contributors table for this contribution 💯

silviuavram:
I will create a PR tomorrow. I intended to do so, but I had busy days at the office. Will create the PR with suggested solution and will also test with browsers and screen readers afterwards. Sorry for the delay.

kentcdodds:
Take your time and thank you!

Issue #630, opened 2018-12-04
Handling Events, prevent default, stop propagation


johnjesse:
This is a question, and this may not be the correct forum for it as it's more about events themselves than Downshift.

Downshift handles the keyDown event and calls preventDefault on the event  https://github.com/paypal/downshift/blob/27a19a31fb3f3f9475d29080817c2ad866c6eed6/src/downshift.js#L556

It doesn't call `stopPropagation` - I'm assuming that's by design.
This causes an issue for me as I use a downshift select input inside another element that is also listening to keyDown events.

My question is why does it only call `preventDefault` and not `stopPropagation too`? Downshift has handled the event so why does it allow it to bubble up and be potentially handled by someone else. I'm really interested in the rationale for only using `preventDefault` in this case and how this applies generally to handling events.

johnjesse:
haha, fair enough,

so I'd argue that it should be using `stopPropagation` too, it's handled the event and no one else should be. By not calling `stopPropagation` anything further up the Dom tree would have to check the event to see if it's already been handled.

I'd be happy to submit a PR

kentcdodds:
That sounds good to me, so long as there's a way to prevent that (like we have with preventDefault). Thanks for your help and understanding <3

Also, your username is the same name as my father-in-law....

Dad?

timvracer:
NOOOOO This is wrong and has broken my app in 3.8, with NO WORKAROUND.  If you want to stop propagation, just create a key handler and stop stop it for the key events you wish to not propagate.  Some of us badly need to know when ESCAPE is hit, and the element that "needs to know" is not the immediate parent element.  Things were fine as they were, but now broken in 3.8.  PLEASE remove the stopPropagation on 'ESCAPE', and for those who want the event to stop, it is very easily handled with a keyEvent handler per the documentation.

kentcdodds:
Sorry to break your app. I agree. I think it would be better to propagate. Feel free to open a pull request to fix this.

timvracer:
Thanks Kent, I submitted a new issue for this... is that good enough, or do I need to go and make the change in the source?  (I did not inspect the source, I just saw this issue and realized what had happened).

Happy to try and figure it out in the source and create a PR if that is what you would like.

johnjesse:
@timvracer sorry that this broke your app, can't you use `preventDownshiftDefault` to stop this behaviour if you don't want it?

johnjesse:
[...]

PR #633, opened 2018-12-13
Stop handled key events propagating

johnjesse:
**What**: Stop the propagation of keydown events that have been handled by downshift

**Why**: If I am using downshift inside a component that is also handling key events then it shouldn't be receiving them if downshift has already taken action

**How**: using `event.stopPropagation()`

**Checklist**:

- [ ] Documentation "N/A" - I think there's nothing to add to the documentation about this, the user can still disable this behaviour (as they could before) by setting `event.preventDownshiftDefault`, so I don't think there needs to be a new addition to the dos?
- [x] Tests
- [x] Ready to be merged
      
- [ ] Added myself to contributors table - the `add-contributors` script failed on my windows machine
      

This fixes issue #630

### DIFF
--- a/src/__tests__/downshift.get-button-props.js
+++ b/src/__tests__/downshift.get-button-props.js
@@ -90,6 +90,26 @@ test(`getToggleButtonProps doesn't include event handlers when disabled is passe
   }
 })
 
+test('stops key events that downshift has handled from propagating', () => {
+  const keyDownSpy = jest.fn()
+  const {container} = render(
+    <div onKeyDown={keyDownSpy}>
+      <Downshift isOpen highlightedIndex={1}>
+        {({getToggleButtonProps}) => <button {...getToggleButtonProps()} />}
+      </Downshift>
+    </div>,
+  )
+
+  const button = container.querySelector('button')
+
+  fireEvent.keyDown(button, {key: 'ArrowDown'})
+  fireEvent.keyDown(button, {key: 'ArrowUp'})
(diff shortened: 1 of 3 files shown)


johnjesse:
Also, running the tests locally
`npm run build-and-test` failed for me... within the build-and-test script, not sure what the error was
`npm run test:cypress` had one failing test

but I think this might just be my machine, as they both failed locally before my changes,

kentcdodds:
I wonder if we should use `stopPropagation` for every event 🤔 (and not just these specific keys).

johnjesse:
Potentially, I think if we are calling `preventDefault` it's a good indicator that we probably want to call `stopPropagation`.

I would be a little wary of overusing `stopPropagation`. Calling stopPropagation is effectively cancelling the event and saying that no-one else can handle it. I have also sees varying accounts for when/if you should call stop propagation, see https://css-tricks.com/dangers-stopping-event-propagation/ for instance.

I think we can safely use `stopPropagation` when you are convinced that nothing else should be doing something with that event (it feels like this is probably generally true). In our case, you are using the keyboard interaction to navigate the downshift control/select an element/close the downshift control. In all cases the keyboard interaction is within the context of the downshift control and doing something to that control. The same keys should not be being re-used for some other functionality whilst we're within the context of downshift. Maybe we review the other event handlers and decide if they fit the same criteria?

kentcdodds:
That makes sense. I think we'll stick with this until someone complains 👍 Thanks!

PR #634, opened 2018-12-18
Fix grammar error in getRootProps error message

petetnt:
This PR fixes grammar in otherwise bit weird error message, a bit much applying and using there (I think...) 😊


petetnt:
Whoops, snappies need to be updated so GH flow kinda fails there, will do asap.

PR #639, opened 2019-01-03
[TypeScript] Fixed type for "GetInputPropsOptions"

franklixuefei:
According to the latest `react` types, we should also allow `HTMLTextAreaElement` as part of the props for `GetInputPropsOptions`

**What**:
Typescript type fix.

**Why**:
to make the type more accurate and prevent broken builds.

**How**:
I added `HTMLTextAreaElement` as part of the props to `GetInputPropsOptions`, according to the latest `react` type def.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged
      
### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -107,7 +107,7 @@ export interface GetRootPropsOptions {
 }
 
 export interface GetInputPropsOptions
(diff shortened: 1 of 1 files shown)


franklixuefei:
Thanks! I’ll try to fix them soon.

PR #641, opened 2019-01-05
(revert) [TypeScript] Fixed type for "GetInputPropsOptions"

franklixuefei:
Revert #639 
I decided to revert this change for the following reasons:
1. For now, downshift is not designed for textareas, so it doesn't make sense to account for the props that a textarea has in typings, which adds to the complexity of the type definitions.
2. this will no longer work (typescript compiler will throw):
```jsx
<input {...getInputProps(...)} />
```
because the returned type `React.HTMLProps<HTMLInputElement | HTMLTextAreaElement>` cannot be assigned to `React.HTMLProps<HTMLInputElement>`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      
### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -108,7 +108,7 @@ export interface GetRootPropsOptions {
 }
 
 export interface GetInputPropsOptions
-  extends React.HTMLProps<HTMLInputElement | HTMLTextAreaElement> {
+  extends React.HTMLProps<HTMLInputElement> {
   disabled?: boolean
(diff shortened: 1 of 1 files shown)


franklixuefei:
Test passed locally:
```
PS C:\opensource\downshift> yarn test:ts
yarn run v1.12.3
$ tsc --noEmit -p ./tsconfig.json
Done in 13.50s.
```

Issue #642, opened 2019-01-05
Issue #630 regarding key event propagation is a breaking change.  I cannot find a workaround that is reasonable


timvracer:
- `downshift` version: 3.8
- `node` version: 
- `npm` (or `yarn`) version:

**Relevant code or config**

```javascript
```

**What you did**:

**What happened**:

**Reproduction repository**:

**Problem description**:

Issue #680 was implemented in 3.8, and unfortunately is a breaking change for anybody who needs the ESCAPE key further up in the component chain.    I'd like to suggest that the previous behavior be returned or a controlling property added.  My rationale:
a) It is trivial to stop propagation on keycodes you wish not be propagated with an input key handler
b) It is non-trivial to workaround the key event being stopped by Downshift when the listener is several layers up in the component chain.
c) an alternative would be to add a property to control this for ESCAPE, such as "escapeDoesNotPropagate", which is similar to what I have seen in some material-ui components

I could understand this change when the menu box is open (Escape closes the box), but there are many application situations for which Escape has a more global impact desired, and is relevant beyond the Downshift context.  (for example, escape always exits a particular edit state for an entire form while within an edit control).  Escape has always been a pretty important method to exit an edit state, so it is necessary to receive the event from controls that may be in a form.  Sometimes the form is complicated, and the input control that will receive the events is several components in from the parent component that listens to the event, and the onKeyDown is not chained throughout all those components.

**Suggested solution**:

johnjesse:
sorry that this broke your app @timvracer , I take it you can't use preventDownshiftDefault to stop this behaviour because you still want downshift to handle it too.

Also... It was my understanding that the escape event handler only ran if the dropdown was open... i.e. if it actually reset the control. If the control was already reset and escape would do nothing I agree it should not be marking the event as handled or stopping propagation (or preventing default) 

I'm not sure about the work around you suggest about it being "trivial" to stop propagation on keycodes you don't want
1) This ties anything I'm wrapping downshift in to understand downshifts implementation details to decide whether it even needs to bother stopping propagation events
2) As above, propagation should depend on the state of the downshift control... as I don't always want to stop the events - only when downshift has used them... so whatever was wrapping would need to know things like the open state etc... 

For the case you suggest, where you always want escape to do something, could you potentially use the capture phase of the event instead? I'm not sure that's the desired solution but it might help you get around the problem

kentcdodds:
[...]

PR #643, opened 2019-01-06
Solution for Issue #642, add new property 'propagateControlKeys'

timvracer:
**What**: a property for the Downshift component has been added, propagateControlKeys, to allow for direct control over whether key events are propagated for special Downshift control keys: Enter, Escape, ArrowUp, and ArrowDown

**Why**: Because some applications need propagation to function correctly, especially with the Escape key, in which they need a higher level component to process the key for additional functions beyond the Downshift default behavior.  Simply 'stealing' the event won't suffice because the Downshift behavior is still needed in this use case

**How**: added a check in downshift.js for both input and button key handling, but sharing the same code which will execute the stopPropagation on the event as appropriate.  I added tests for both button-props and input-props to ensure each control key specification functions, as well as testing for a combination of them.

**Checklist**:

- [x ] Documentation
- [x ] Tests
- [x ] Ready to be merged
      
- [x ] Added myself to contributors table


timvracer:
Ok, the pull request is active.  Please review.  Thank you.
Not sure why it says "some checks haven't completed yet"... maybe waiting for Travis.

timvracer:
Travis is showing an error in 'combobox.js', but you can see in my checkins that I did not touch that file, so I don't know why that one test is failing.

README.md
```diff
@@ -523,6 +524,13 @@ node you render in downshift). Internally we use
 so if you use that package then you wont be adding any additional bytes to your
 bundle :)
 
+### propagateControlKeys
```
kentcdodds: I'm not so sure I like this solution for two reasons:

1. I feel like it's less flexible than it could be
2. I don't think this prop will be used enough to justify its existence.

I think that I'd rather just disable event propagation for when our event handlers actually run. I believe that should solve the situation that we're dealing with here.

johnjesse: Cool, so the solution would be pretty much just this bit of code https://github.com/paypal/downshift/issues/642#issuecomment-451673844, without the prop check.
May I suggest we put the prevent default inside the conditional too?

kentcdodds: > May I suggest we put the prevent default inside the conditional too?

Hmmm... That seems reasonable, but I'm not sure of the implications. Let's start with changing propagation first.

timvracer: Hi Kent and John,
Going back to disabling propagation will certainly work for me. There are lots of cases where an advanced UI may need to see these keys, so I think this makes sense.

That said, to make smart decisions on keys like Enter, Escape, and others, I require knowing if the menu is open or not.  (e.g. If the menu is open, then ESCAPE should mean, "close the menu".. done.  If the menu is not open, then Escape needs to be propagated, because it most likely means "close the dialog").

The only gotcha here was that if the list becomes empty (no matches for what the user has typed), the menu "closes" (the new render state of Downshift with no items is that there is no menu), but there is no notification of this sent via onStateChange (a state change is noted, but isOpen is undefined)

Like I said, I am already handling this in my keyHandler code, but it requires some awkward code in onStateChange to manage my own "isOpen" local state to try and keep track of the menu state.  It looks like this:

```
            onStateChange={(
              {
                highlightedIndex,
                // inputValue,
                isOpen,
                // selectedItem,
              }
            ) => {
              // NOTE, Downshift does not accurately tell us when the list
              // closes if it closes due to there being no matches.  So we proxy that here
              // by setting isOpen to reflect whether the current list being rendered has any
              // items at all (in the cases where isOpen is undefined)
              //
              if (isOpen !== undefined) {
                this.setState({ isOpen: Boolean(isOpen) });
              } else {
                this.setState({ isOpen: this.state.listItems && this.state.listItems.size > 0 });
              }
              this.setState({ currSel: highlightedIndex });

            }}

```
Maybe this is a separate Issue report... isOpen is not sent as 'false' when the state changes from open to closed due to items becoming an empty list.

timvracer: OH NO, I said that wrong!
I meant "Going back to propagating all events", not disabling propagation.  SO SORRY!

timvracer: That said, doing what you suggest by adding a property to the event to indicate one way or the other would be fine.  I would vote that Downshift propagate all events by default, and that the user can set stopDownshiftPropagation to request that for that event Downshift stops the propagation.  Good solution Kent!

kentcdodds: > I would vote that Downshift propagate all events by default, and that the user can set stopDownshiftPropagation to request that for that event Downshift stops the propagation.

I will merge a PR that does this :+1: Thanks!

timvracer: ummm... ok, thinking about this... I don't think we need a property in the event.  If I have to build a keyHandler in order to set that property, then that handler can simply call stopPropagation() directly on the event which will suffice (and what I am doing today).   Or am I missing something?  I think we just need to go back to the version that always propagates events.

kentcdodds: Honestly I think I'm in favor of that myself. @johnjesse was saying that would make it difficult for what they're doing though. I'd like to know more about why because all the downshift state should be available within the closure of the event handler.

timvracer: "all the downshift state should be available within the closure of the event handler".    AFAIK the event handler passed to Downshift cannot access the state of the Downshift closure (although  I'd love to learn how).  However, I use onStateChange to keep track of relevant state changes from Downshift, and use that info in my event handlers.

johnjesse: So - I'm sorry this is a bit verbose, writing it down will make sure I understand it properly. My use-case is using downshift as an autocomplete/dropdown inside [react-day-picker](http://react-day-picker.js.org/examples/elements-year-navigation), which could be used potentially inside another pop-up within my app. The day-picker does great stuff with keyboard handling for arrow keys, changing months and years etc. Similarly any popup is likely to want to handle escape events to close themselves.

But it's a weird UX if keyboard buttons do more than one thing, in my example using the down arrow key would both open the downshift dropdown and change the year in the day-picker. Similarly pressing escape when the downshift dropdown was open not only exits the downshift control, but also the popup.

This is what I was trying to address with #633.

So, let's say we revert that pr, how do I solve my problems -
1) I could wrap downshift in an element and handle onKeyDown events and handle anything that downshift has handled. I could keep track of downshift's state via `onStateChange`, use that state to decide if downshift had done anything with the event and then decide whether to stopPropagation of the event/
Issues with this are, I need to know what events and keys downshift is using internally and I need to duplicate downshift state, which is cumbersome. I think this is what Tim is already having to do.
2) I could, as suggested above, pass an event handler go `getInputProps` and `getToggleButtonProps` that has access to the downshift state through the propGetters.
This gets around me keeping track of the state myself, but means I still have to know exactly what events downshift is using, on which elements and how that changes depending on state.

This means whatever I do is very heavily coupled to the internal implementation of downshift. And it's tricky to know in advance if I need to stop evnt propagation or not, as it depends on where my wrapped downshift control is being used in the rest of the application.

I'd argue that downshift should do the most sensible thing by default. For instance, I think we'd all agree that if escape has caused the downshift to remove an open popup it usually doesn't want to cause anything else to happen. I think the way to do that would be stopping propagation when it has handled the event in a way that causes a change to the UI. Although it might be that downshift is unable to know for certain if it has done that.

So I think it comes down to what should/can the default behaviour of downshift be. I think we want to either revert #633 or do something like this PR (as #633 was incomplete).

timvracer: I always wrap 3rd party components like this when I have a customized form solution and need to manage these types of events.  I guess my view is that this is why the whole event handler architecture exists, to allow folks to customize behavior for their specific needs.

BTW - you can see the code above for tracking the menuIsOpen state, while not ideal, it is not that difficult.  This is in the "wrapper component" that wraps the Downshift component.

That said, if we make the default for Enter and Escape to stop propagation when the menu is open (since those events will cause the menu to close), that will work fine in my case and saves me a few lines of code (I no longer need to stopPropagation() on those keys when the menu is open).   

The one concern I can see is that making it this way and requiring a event property (event.doPropagateEvent) to ask for the event to be propagated is a non-standard pattern, and will require someone to find the solution in the docs, but probably only after hours of frustration - but agree that for that vast majority of users, they would want Enter/Escape to no propagate if they were used when the menu was open.

johnjesse: :smile: I'm ok whichever solution we pick, it's more if many people need to add those extra lines of code to manage events then it feels like it should be the default.

I think you're right about it feeling a bit odd to have a `propagateEvent` property though...

kentcdodds: Honestly friends, I'm not sure. I haven't used downshift in over a year and I just don't think that I'm the right person to maintain the project anymore. I'm just too detached from the use cases.

https://github.com/paypal/downshift/issues/647

johnjesse: So if it just comes down to what myself and Tim then I'm coming around to "we should just revert #633".

As much as I think the default behavior I've suggested is correct, I'd hold off on making that change because right now we only know two of us who it would help... it feels a little premature to make that the default, especially given there are workarounds. It's probably better to revert the broken always stop propagation, and wait and see if this arises again later?

What do you think Tim?

johnjesse: I suggest we close this then and revert, in a different PR

johnjesse: #649

PR #646, opened 2019-01-07
Handle End and Home Keys in list navigation

silviuavram:
**What**:
This PR aims to improve kb navigation inside the listbox by adding support for `Home` and `End`. These keys should highlight first / last element, if the listbox is visible.

**Why**:
Improves the component by adding another ARIA specification regarding kb navigation.

**How**:
Added event listeners for keydown on both `Home` and `End`, and added handlers for each, in which I've used variations for the `setHighlightedIndex` function. Also added event names for both kets.

Also added UTs for these changes, and planning to add E2E Cypress tests too. These are already written, however Cypress does not support Home and End at the moment. I've created a PR for implementing this in their repo [here](https://github.com/cypress-io/cypress/pull/3071). Those tests are currently commented.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      
### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,65 +1,60 @@
 {
+  "dist/downshift.cjs.js": {
+    "bundled": 54264,
+    "minified": 24037,
+    "gzipped": 6628
+  },
   "preact/dist/downshift.cjs.js": {
-    "bundled": 51792,
-    "minified": 22290,
-    "gzipped": 6337
+    "bundled": 52833,
+    "minified": 22887,
+    "gzipped": 6433
   },
   "preact/dist/downshift.umd.min.js": {
-    "bundled": 64448,
-    "minified": 21905,
-    "gzipped": 7058
+    "bundled": 63298,
+    "minified": 22029,
+    "gzipped": 7062
   },
   "dist/downshift.umd.min.js": {
-    "bundled": 68904,
-    "minified": 23079,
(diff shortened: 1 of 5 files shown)


kentcdodds:
Ah, looks like we're just shy of our coverage threshold. Could you give that a look please?

silviuavram:
sure, will take a look tomorrow and come back to you. Thanks for the feedback!

kentcdodds:
Hey @silviuavram, would you be at all interested in being a primary maintainer of the project? Of everyone I feel really comfortable handing this project to you. See #647

silviuavram:
Hello Kent,

Thank you very much for your considering me. It will be both a pleasure and
a challenge to maintain and improve the project.

We are using it currently in our own project, Stardust UI Library, and I am
planning to contribute more to Downshift, as we are finding use cases that
need to be handled and it's way better to improve the source rather than
fixing stuff at our end. It will also be valuable for everyone using the
library.

If you agree to go forward with your decision, I will kindly ask for all
the information you consider necessary and good for me to know. I will also
sync with people from my organisation that have strong knowledge of project
ownership, but your help will also be very valuable.

I look forward to hearing from you.

Best regards,
Silviu

kentcdodds:
Sounds awesome! Could you shoot me an email and we'll chat about what needs to happen for this to work :) kent@doddsfamily.us

PR #648, opened 2019-01-08
Change `touchStart` to `touchEnd` in the TypeScript definitions for `StateChangeTypes`.

mkaradeniz:
**What**:

I replaced `touchStart` with `touchEnd` in the `StateChangeTypes` enum of the TypeScript definitions. 

**Why**:

Currently, the `StateChangeTypes` enum of the TypeScript definitions defines `touchStart` as a possible state change that the `stateReducer` is called on (with `__autocomplete_touchstart__`), but this is not case. Instead the `stateReducer` is called on `touchEnd` with `__autocomplete_touchend__`.

**How**:

Updated the [index.d.ts](https://github.com/paypal/downshift/blob/master/typings/index.d.ts).

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Ready to be merged
      
One `cypress` test failed right after cloning and also after my changes. Interestingly, it is not always the same test. On windows the `can use the mouse to click an item` test failed and on MacOS `resets the item on blur` failed once and on another run the `can down arrow to select an item` failed instead. Everytime it was this error `Uncaught SyntaxError: Unexpected token <`. So I'm not sure if this actually read to be merged right away.


capsula4:
This is awesome, I was looking for this!

silviuavram:
hello @mkaradeniz thank you for this PR. will take a look after I get some context, and will come back to you asap.

silviuavram:
ok cool, can you please merge latest master in you branch @mkaradeniz ?

mkaradeniz:
Any idea why the tests are failing?

silviuavram:
some new lint rule enforced. will need to fix that first before commiting this.

PR #649, opened 2019-01-08
Revert "fix(events): stop handled key events propagating (#633)"

johnjesse:
This reverts commit 5084cd05d0ca5ef7ff9f95de5a8dab9e882d73d8.

**What** and **Why**

The changes as part of #633 were incomplete, did not provide a way to allow downshift to handle event and optionally stop propagation, and incorrectly always stopped event propagation on Escape Keydown events.

See discussion in #642 and #643

Whilst we may want to do some stopping of propagation, for now this just reverts that PR so that the events are always propagated.

### DIFF
--- a/src/__tests__/downshift.get-button-props.js
+++ b/src/__tests__/downshift.get-button-props.js
@@ -90,26 +90,6 @@ test(`getToggleButtonProps doesn't include event handlers when disabled is passe
   }
 })
 
-test('stops key events that downshift has handled from propagating', () => {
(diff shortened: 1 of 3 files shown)


johnjesse:
Note that since then, Home and End key events have been added, which also stopPropagation (maybe because the other events did), should that be removed to

kentcdodds:
>  should that be removed to?

Yes, let's remove it from there as well.

timvracer:
Wooot!!!!!  Thanks all!!

PR #658, opened 2019-01-29
chore: replace "paypal/downshift" with "downshift-js/downshift"

kentcdodds:
**What**:

replace "paypal/downshift" with "downshift-js/downshift"

**Why**:

Because downshift is a community project. Even though PayPal is still a heavy user, it makes more sense to be a project on its own considering the primary maintainer is not a PayPal employee.

**How**:

find/replace


README.md
```diff
@@ -1,7 +1,7 @@
 <h1 align="center">
   downshift 🏎
   <br>
-  <img src="https://cdn.rawgit.com/paypal/downshift/d9e94139/other/logo/downshift.svg" alt="downshift logo" title="downshift logo" width="300">
+  <img src="https://downshift.netlify.com/logo/downshift.svg" alt="downshift logo" title="downshift logo" width="300">
```
kentcdodds: rawgit is shutting down, so we're going to host this on our netlify site.

package.json
```diff
@@ -28,6 +28,7 @@
     "cy:open": "cypress open",
     "build-and-test": "npm run build -s && npm run test:build -s",
     "docs:build": "docz build",
+    "postdocs:build": "cpy \"**/*\" ../../.docz/dist --cwd=\"other/public\" --parents",
```
kentcdodds: Because rawgit is shutting down, we need to find another place to host static files. So I created a `other/public` directory. All the files in there will be copied as-is to the `.docz/dist` directory which is what netlify is configured to publish.

PR #669, opened 2019-02-13
Allow `null` item to be selected

brabeji:
**What**:

I'm currently unable to select `null` as an item.

**Why**:

A `null` happens to be a valid item in my use-case.

**How**:

I've updated item existence check in `selectItemAtIndex` to `item === undefined`

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged
      
- [ ] Added myself to contributors table N/A - just a tiny change

### DIFF
--- a/src/__tests__/downshift.misc.js
+++ b/src/__tests__/downshift.misc.js
@@ -55,6 +55,24 @@ test('selectItemAtIndex can select item that is an empty string', () => {
   )
 })
 
+test('selectItemAtIndex can select item that is a null', () => {
+  const items = ['Chess', null]
+  const children = ({getItemProps}) => (
+    <div>
+      {items.map((item, index) => (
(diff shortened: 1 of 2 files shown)


silviuavram:
I don't know about this. I cannot see why would you ever need `null` as a valid item. how do you display it? how do we narrate it for accessibility? and I bet there is more that one line to change in our logic. but really I cannot see why we would ever do this. @kentcdodds ?

brabeji:
@silviuavram This change is similar to https://github.com/downshift-js/downshift/pull/242 where I believe all of your arguments apply. How to display it should be answered in userland imo.

brabeji:
I thought about this a bit more and I don't really like `null` being of different type than rest of items in a common scenario.

PR #672, opened 2019-02-14
fix(unnecessary-state-call): remove call to set highlighted index on toggle if there is no default

GonchuB:
**What**:

When the toggle button gets clicked, after setting `isOpen` on internal state, another call to set internal state was fired with the default highlighted index. If that prop is not defined, the state will be set either way with null.

**Why**:

If you want to customize the behavior with `stateReducer` you will get two calls if you want to hook in on open and you are not specifying `defaultHightlightedIndex` prop

**How**:

Checking if the prop is defined before attempting to set internal state. Add tests to cover the scenario.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged


silviuavram:
Thank you for your help on this. Will take a look when I get to find some time and we will take it from there.

silviuavram:
I've been think about this a bit more, and I think some of our state changes are a bit chaotic. Always, after each type of action, it should be a single state change, with every property that has changed. For instance when you open the list with Down/Up it will trigger 2 state changes. I am planning to create a bigger PR to address these issues, and will add you as a contributor there, and will ask for your review. Thanks!

GonchuB:
@silviuavram let me know if you need help with the actual PR implementation (or testing, or whatever)

silviuavram:
will try to handle this case in https://github.com/downshift-js/downshift/pull/674 @GonchuB please let me know what you think. Does it include your fix? You can also help me with reviewing it and testing. Thanks a bunch and for pointing out that we miss some stuff in the highlighting department, we really needed it.

silviuavram:
closing this, I merged my PR. forgot to add you to contributors, will do that as soon as possible. or if you want you can create another PR with just the add contributor and I will merge it. Thanks!

PR #674, opened 2019-02-20
fixed index highlights in toggle and keydown

silviuavram:
**What**:

Fixes item highlights in the case of passing `highlightedIndex`, `defaultHighlightedIndex` and `initialHighlightedIndex`. Handles both toggle and opening by keyDown and keyUp. Also has some code refactoring + unit tests for `getInputProps`. Planning to also add unit tests for `getToggleButtonProps`.

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      
### DIFF
--- a/src/__tests__/downshift.get-input-props.js
+++ b/src/__tests__/downshift.get-input-props.js
@@ -269,6 +269,7 @@ test('navigation key down events do nothing when no items are rendered', () => {
     arrowUpInput,
     endOnInput,
     homeOnInput,
+    escapeOnInput,
     childrenSpy,
   } = renderDownshift({items: []})
   const keysOnInput = [arrowDownInput, arrowUpInput, endOnInput, homeOnInput]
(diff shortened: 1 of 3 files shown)


src/downshift.js
```diff
@@ -561,18 +533,50 @@ class Downshift extends Component {
...
-      })
+
+      if (this.getState().isOpen) {
+        const amount = event.shiftKey ? 5 : 1
+        this.moveHighlightedIndex(amount, {
+          type: stateChangeTypes.keyDownArrowDown,
+        })
+      } else {
+        const self = this
+        this.toggleMenu({type: stateChangeTypes.keyDownArrowDown}, () => {
```
GonchuB: nice 👌

src/__tests__/downshift.get-input-props.js
```diff
@@ -90,55 +90,177 @@ test('manages arrow up and down behavior', () => {
...
+    )
+  })
+
+  test('initialHighlightedIndex - 1', () => {
+    const initialHighlightedIndex = 3
+    const {arrowUpInput, childrenSpy} = renderDownshift({
+      // provide only highlightedIndex
+      props: {initialHighlightedIndex},
+    })
+    // ↓
```
GonchuB: should be `// ↑` in this and the following

src/downshift.js
```diff
@@ -561,18 +533,50 @@ class Downshift extends Component {
...
     },
 
     ArrowUp(event) {
       event.preventDefault()
-      const amount = event.shiftKey ? -5 : -1
-      this.moveHighlightedIndex(amount, {
-        type: stateChangeTypes.keyDownArrowUp,
-      })
+
+      if (this.getState().isOpen) {
```
GonchuB: The only difference I see between these handler and ArrowDown is the `stateChangeType` and `newIndex`. Does it make sense to move to a method and pass the correct newIndex value?

Maybe it complicates things as `newIndex` is calculated from state after the menu gets toggled. And repeating something twice could be better than a wrong abstraction.

src/__tests__/downshift.get-input-props.js
silviuavram: thanks, will change!

Issue #664, opened 2019-02-11
Feature Request: Reset downshift state when inputValue changes


bkilrain:
- `downshift` version: 3.1.12
- `node` version: 8.11.1
- `npm` (or `yarn`) version: 5.6.0

**Relevant code or config**

```javascript
```

**What you did**:

I'm using Downshift to render the dropdown for an autocomplete component. The `defaultHighlightedIndex` is set to `0` so that there is always a 'highlighted' item in the dropdown. 

**What happened**:

If I hover an item lower in the list (index 4 or 5), however, and type to filter out possible items to a number less than the highlightedIndex, the `highlightedIndex` state does not update to ensure a highlighted item is still rendered in the dropdown.

**Steps to Reproduce**:

[Code Sandbox Link](https://codesandbox.io/s/74y7y4j6rx)

1. Type 'a' into the input
1. Hover over the last item in the list to set the highlightedIndex
1. Type 'p'
1. Notice the highlightedIndex doesn't update to highlight the top item

**Suggested solution**:

Right now, I'm calling `reset()` on every input change to get around this. As far as changes to the library, perhaps we could compare the highlightedIndex with the itemCount to make sure a highlighted item is being rendered at all times? Or add the inputChange event to cases where downshift's state is reset. Can anyone think of a usecase where they don't want downshift state to reset on inputChange?

silviuavram:
Thank you for the improvement request @bkilrain . I'd be happy to see this solved in our library so every user can benefit from it. Are you up for a Pull Request? :)

bkilrain:
Yep - I'm happy to take this on. To be clear, we just want to reset Downshift state when `inputValue` changes?

silviuavram:
I don't think we need to do all of that. As I understand, the issue is to reset only the highlightedIndex to the defaultHighlightedIndex each time the input value changes. Am I correct? 

Resetting the whole state will probably mean other state stuff reset which we may not want to reset, like the selected item.

bkilrain:
Ok - fair enough.

kilrain:
Hey there - I've completed the fix but I'm getting [this error](https://github.com/pedronauck/docz/issues/596) when I run the `setup` script. I'll cut a PR anyways... just want to let you know.

Also, I'll submit from my work github account. Thanks!

PR #675, opened 2019-02-20
Set highlightedIndex to defaultHighlightedIndex on input value change

kilrain:
**What**:

This PR ensures the highlightedIndex is reset to defaultHighlightedIndex when the inputValue is changed. 

**Why**:

documented here: #664 

**How**:

In both `inputHandleChange` and `inputHandleTextChange`, I included the following bit of state to update:
```
{
  highlightedIndex: this.props.defaultHighlightedIndex,
}
```
I've also added a test and updated the docs to reflect this change.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged
      
Additionally, 
* I ran into an error when running the `setup` script which is documented in the [feature request ticket](#664 )
* I added `.vscode` to `.gitignore`


silviuavram:
I've added some changes related to these highlighted indexes. Make sure to rebase and see that it works.

I will try to help you as much as possible to get these changes in, as it will be a great improvement for our accessibility. Thanks again for contributing!

kilrain:
Got it... I'll pull down the changes and confirm we're good to go.

kilrain:
@silviuavram My smoke testing worked as expected and all unit tests pass as well.

src/__tests__/downshift.get-input-props.js
```diff
@@ -583,6 +583,25 @@ test(`getInputProps doesn't include event handlers when disabled is passed (for
...
+  const props = {defaultHighlightedIndex: 0}
+  const {childrenSpy, arrowDownInput, changeInputValue} = renderDownshift({
+    props,
+  })
+
+  childrenSpy.mockClear()
+
+  arrowDownInput() // highlightedIndex = 1
+
+  changeInputValue('c')
```
silviuavram: let's input `r` instead, to make sure the number of resulting items is more than one (in this case it will be 4). that will actually test that the highlight is going to the right index, not just because there was only one option (if you type `c` you will get only one option)

src/__tests__/downshift.get-input-props.js
```diff
@@ -583,6 +583,25 @@ test(`getInputProps doesn't include event handlers when disabled is passed (for
   }
 })
 
+test('highlightedIndex is reset to defaultHighlightedIndex when inputValue changes', () => {
+  const props = {defaultHighlightedIndex: 0}
```
silviuavram: maybe add `const defaultHighlightedIndex = 0` and use that in props and in assertion. Not a big deal but it can be better.

.gitignore
```diff
@@ -7,6 +7,7 @@ dist
 .next
 .eslintcache
 .docz
+.vscode/
```
silviuavram: let's not add stuff unrelated to this PR scope.

src/__tests__/downshift.get-input-props.js
kilrain: good point

src/__tests__/downshift.get-input-props.js
kilrain: That's fair

.gitignore
kilrain: no problem

kilrain:
Ok - change requests resolved.

Issue #680, opened 2019-02-21
Escape should clear text in input


silviuavram:
https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

Escape key inside input should clear the text from it. Currently we are resetting it to the string value of the item.

babzcraig:
If no one is on this yet, I'd like to take this on @silviuavram. I've read the requirements on the w3 page and can make a PR for this

silviuavram:
That is great to hear! It's all yours, let me know when I can help out. Thanks!

silviuavram:
@babzcraig how is the progress on this one? need any help? :)

gaskar:
hi @silviuavram I can take this one.

babzcraig:
@gaskar please feel free to. I couldn't get downshift's `validate-setup` command to work and I've been too swamped to debug for a while. @silviuavram perhaps you can help me with this and I can take on another issue. Thanks!

gaskar:
@babzcraig seems there is a problem with cypress tests, I've tried different versions nothing is working beside of just commenting out these tests, seems that @silviuavram created PR for this problem but it is failing as well
https://github.com/downshift-js/downshift/pull/661

babzcraig:
I see. I had just changed my development machine so I though maybe I hadn't set up things well yet. I might have to do the same then. If you want to take this issue, please do. I won't be available until Sunday so I'll just pick another issue then if you're on this

brandonkal:
This should be reconsidered. Escape is usually used to blur an element. Clearing the textbox is an unexpected side effect.

See: https://github.com/w3c/aria-practices/issues/1066
Reverting this is already planned for aria-practices 1.2

Just to provide some additional examples, try pressing escape in the "Search or jump to..." box above or any search engine.

The only other place I have seen this pattern is in the Mac System Preferences search box, but this is done in tandem with removing focus on the input, which is the expected behavior.

filipjnc:
Please reopen the issue, as there is a new official guideline by ARIA:
The Escape key should only clear the input, if the dropdown is **not** currently open.
Source: https://github.com/w3c/aria-practices/pull/1129

PR #689, opened 2019-03-15
clear input text on escape key #680

gaskar:
**What**: Fixed issue concerning escape key handling for comboboxes from w3 standards
https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

Fixes https://github.com/downshift-js/downshift/issues/680

**How**: Set inputValue to empty string on escape key handler, before calling reset method.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged
      

src/downshift.js
```diff
@@ -610,7 +610,10 @@ class Downshift extends Component {
 
     Escape(event) {
       event.preventDefault()
-      this.reset({type: stateChangeTypes.keyDownEscape})
+      this.reset({
+        type: stateChangeTypes.keyDownEscape,
+        inputValue: '',
```
silviuavram: I have been thinking about this. if we clear the input, we should also remove the selected item, if any. in the ARIA docs, the whole point of the combobox is to fill this input with a string value, doesn't matter what the value you compute it from is. if ARIA says the input is cleared at Escape, then there is not point in having a selected item.

also, I think our reset function needs to clear the selected item by default. but unfortunately right now it's being used in our blur handles. ARIA says that on blur, if there is a highlighted item, it should be selected. I created an issue for this https://github.com/downshift-js/downshift/issues/679.

that being said, if you agree, we can add selectedItem as null here, update the documentation on the Escape handler, and then we can start working on having selectedItem: null in reset function and then selecting the item on blur, thus removing some uses of the reset function. in a separate PR. what do you think?

gaskar: @silviuavram agree, sorry I didn't see your comment here, will do as you suggested and update PR.

gaskar: @silviuavram can you revisit this PR?

Issue #694, opened 2019-03-23
Controlling when updateStatus is called


Nic128:
- `downshift` version: 3.2.7
- `node` version: 10
- `npm` (or `yarn`) version: 1.13

**Relevant code or config**

```javascript
updateStatus = debounce(() => {
    const state = this.getState()
    const item = this.items[state.highlightedIndex]
    const resultCount = this.getItemCount()
    const status = this.props.getA11yStatusMessage({
      itemToString: this.props.itemToString,
      previousResultCount: this.previousResultCount,
      resultCount,
      highlightedItem: item,
      ...state,
    })
    this.previousResultCount = resultCount

    setA11yStatus(status)
}, 200)
```

**What you did**:

I have a very similar integration to this asynchronous example: https://codesandbox.io/s/github/kentcdodds/downshift-examples/tree/master/?module=%2Fsrc%2Fother-examples%2Fgmail%2Findex.js&moduleview=1

**What happened**:

I tested with a screen reader (VoiceOver in my case, but can be reproduced with any). When typing one letter, the `updateStatus` gets called. You get the `No results are available` announcement while the asynchronous fetch is happening. Then after the result are loaded, you can't rectify the announcement.

**Reproduction repository**:

**Problem description**:

I can customize the message announced, great. I can set state after the results are loaded, great. But I can't trigger the `updateStatus` when I want it to. It feels like you are one step behind every state change.

**Suggested solution**:

I'd like to prevent this default behaviour of immediately firing `updateStatus` upon input change, and a children prop of some kind to fire it when I want. Because of the synchronous nature of the event, currently it doesn't really work with an asynchronous setup.

PR #695, opened 2019-03-25
feat(loading): Prevent default triggering of …

Nic128:
…getA11yStatusMessage via updateStatus (#694)

 - Rename updateStatus to updateA11yStatus
 - Expose setLoading via children props

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged
      

silviuavram:
Thank you for taking on this issue with async load. We also had it in Stardust-ui, and we needed to handle it. 

What I am wondering is whether we can handle it without this much code. For instance, in stardust, we have a `loading` prop that, when true, says the data is loading. if that prop is true, the a11y status is not read. then when you are done loading, you also turn that prop to false, and a11y status will be read.

we will just need that prop and add it to the a11y reading function. exposing a11y reading in the children is not necessarily that value imo.

what do you think? :)

Nic128:
Sounds a lot simpler, I will give it a shot.

silviuavram:
let me know if you need further help with this. it will be cool to have this prop and also it would be awesome to have a working example in the docs. rendering a spinner while loading and then displaying results after the mock load.

Nic128:
Now with the prop, you get a consistent API. I refactored my integration (With a declarative fetch) to use the state reducer, works well.

```
  stateReducer = (state, changes) => {
    // If opening the menu, it means we're calling async calls
    // Set loading internal state to true to avoid triggering the status
    const isOpening = !state.isOpen && changes.isOpen === true

    switch (changes.type) {
      case Downshift.stateChangeTypes.changeInput:
        return {
          ...changes,
          loading: true
        }
      default:
        return {
          ...changes,
          loading:
            isOpening ? true : changes.loading
        }
    }
  }
```

Also, a trick to possibly document once examples are done. My integration has two downshift instances with the async call. Switching between the two would not trigger the status if the message is the same. It's a bug with some screen readers. Found a trick here. https://core.trac.wordpress.org/ticket/36853#comment:10

```
  // These are the messages announced to screen reader
  // If you return empty, nothing will be announced
  // If you return the same message twice, nothing will be announced
  // That is kind of a bug depending on the screen reader being tested
  // The `\u00A0` (no-break space character) is to force a non-empty new entry
  // in the aria-live region so the next message gets read
  getA11yStatusMessage = ({
    isOpen,
    selectedItem,
    resultCount,
    previousResultCount,
    itemToString
  }) => {
    const { translator } = this.props

    if (!isOpen) {
      return selectedItem ? itemToString(selectedItem) : '\u00A0'
    }

    if (!resultCount) {
      return 'No results are available.'
    }

    if (resultCount !== previousResultCount) {
      return `${resultCount} result${
        resultCount === 1 ? ' is' : 's are'
      } available, use up and down arrow keys to navigate. Press Enter key to select.`
    }

    return '\u00A0'
  }
```

silviuavram:
I love these changes! Looks good, this seems ready to merge.

One last (really last) thought, regarding the a11y message. Right now we skip the update if the prop is loading, right?

How about we pass the `loading` prop to the `getA11yStatusMessage` function and based on that it will generate a message like `Loading results` or something similar. That way, the user gets feedback on the progress, which should be a nice touch.

Overall, it should be input change -> loading message narrated -> results message narrated. Do you think it's a good idea to add this to the PR? If so, the last unit test may be deleted, and instead a new one added to `get-a11y-status-message.js` checking for loading message if prop is passed and true.

Thank you again for the contribution to Downshift!

Nic128:
I find voicing the loading message brings more noise than it's worth. As a blind user, it becomes overwhelming to hear on each key typing:

- (The user focuses on the field)
- Loading...
- 20 results are available...
- (The user types `a`, the screen reader voices `a`)
- Loading...
- 3 results are available...
- And so on in rapid succession, what's important are the results.

Plus, the `updateA11yStatus` is debounced, if your ajax call is very fast, you might voice the loading message but lose the debounced execution when the results are back.

We can revisit the use case in a separate issue if we can validate the need for it.

silviuavram:
Good point. Thanks again for contributing, I now managed to take a better look at the change.

Sorry for taking so long, I am still learning how to manage this library and don't want to rush into adding changes.

That being said, since you want only to suppress the message while you load, I think your initial implementation was enough. The `loading` thing does not bring us much value as I see it now, and there is no reason to add more state and all that stuff.

I created a PR https://github.com/downshift-js/downshift/pull/701 that adds a prop to prevent the message to be added. Please check if it solves your case, review it, approve it, and once it is done I will add you as contributor, change the docs and merge it asap.

Please do continue to be active with issues and contributing in the future :)

silviuavram:
After talking about it with others, we agreed that we will not introduce any changes for this case, for now. I always felt that we are adding more stuff to the API just to treat a corner case, that can be achieved with changing the prop `getA11yStatusMessage` from the loading case (you will return empty string) to the normal case (return normal function with status) and toggle between them as needed. You can also hack it further, have it use your `loading` variable, apart from the options parameter it already takes, and when it's loading, return empty string.

It's not ideal, but for now this is what we decided. Thank you very much for being involved in this!

Nic128:
I like the suggestion. I will try it. Thanks.

PR #701, opened 2019-04-13
feat(a11yStatus): add preventA11yStatusMessage prop

silviuavram:
**What**:

Adds `preventA11yStatusMessage` prop to Downshift.

Closes https://github.com/downshift-js/downshift/issues/694.

**Why**:

In case a user wants to prevent our a11y message generation, he should be able to do so by supplying this prop.

**How**:

Added prop to the list of props. Still remains to be added: documentation.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged
      
### DIFF
--- a/src/__tests__/downshift.misc.js
+++ b/src/__tests__/downshift.misc.js
@@ -6,6 +6,10 @@ import React from 'react'
 import ReactDOM from 'react-dom'
 import {render} from 'react-testing-library'
 import Downshift from '../'
+import setA11yStatus from '../set-a11y-status'
+
+jest.mock('../set-a11y-status')
+jest.useFakeTimers()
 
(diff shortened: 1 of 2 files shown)


Nic128:
If we add `preventA11yStatusMessage`, we need to be able to call `updateA11yStatus` manually somehow. Otherwise, we have to toggle that `preventA11yStatusMessage` prop.

It was my previous implementation, but does introduce a non-conventional API, to where you can call `updateA11yStatus`, but onlt from children prop. https://github.com/downshift-js/downshift/commit/72ef2f961c8505d4eab62c4a104c0badc6e01ff7

The `loading` prop approach is more conventional and can used through children prop and regular prop.

PR #703, opened 2019-04-14
fix: use correct state change type for End key (#702)

### LINKED ISSUE #702: End key comes up as stateChangeTypes.keyDownEnter in stateReducer
- `downshift` version: 3.2.7
- `node` version: 11.14.0
- `npm` version: 6.9.0

**Relevant code snippet**
```javascript
stateReducer = (state, changes) => {
  if (changes.type === Downshift.stateChangeTypes.keyDownEnter) {
    // will also print "pressed enter key" when the "End" key was pressed
    console.log('pressed enter key');
  }
};

<Downshift stateReducer={stateReducer}>...</Downshift>
```
In the state reducer, pressing the "End" key is registered as `keyDownEnter`. This is because of a copy-paste error in `stateChangeTypes.js`. 

I'll provide a PR to fix this.

rincedd:
**What**:
Fixes #702 

**How**:
Fixed `keyDownEnd` and `keyDownEnter` constants having the same value. Also, the Cypress tests for "End" and "Home" keys were still commented out, so I enabled them, too. 

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged
      

cypress/integration/combobox.js
```diff
@@ -37,7 +37,6 @@ describe('combobox', () => {
       .should('have.value', 'Red')
   })
 
-  /* should be enabled if https://github.com/cypress-io/cypress/pull/3071 gets released.
```
silviuavram: leave these commented please. We need cypress 3.2.0 for them. If you want to enable them please create another PR in which you also upgrade the cypress version and check they run locally as expected.

since our cypress tests don't run on CI anymore for a long time, maybe you can take a look at https://github.com/downshift-js/downshift/issues/704 and see if you have any experience in solving this. You will be a hero!

rincedd: Alright, I'll leave them as is for now and perhaps address them in a separate PR. I wasn't aware that the cypress tests aren't run on CI at the moment. I'll update the PR. I'll also see if I can do anything about #704, although I do not have much experience with cypress itself.

PR #706, opened 2019-04-18
Update typings for toggle button

everdimension:
Since @types/react had been updated
(https://github.com/DefinitelyTyped/DefinitelyTyped/commit/4d371be185ddd77264a8d7f30a7f7f8912738ed8),
it is required to use ButtonHTMLAttributes generic to avoid typescript
errors

**Checklist**:

- [x] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -116,7 +116,7 @@ export interface GetLabelPropsOptions
   extends React.HTMLProps<HTMLLabelElement> {}
 
 export interface GetToggleButtonPropsOptions
-  extends React.HTMLProps<HTMLButtonElement> {
+  extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   disabled?: boolean
 }
 


everdimension:
@kentcdodds Ping 🙂 
I think this is still relevant for ts codebases

kentcdodds:
I'm no longer the maintainer of downshift. But don't worry, there are *89* people "watching" this repository and getting notifications when you comment :) They'll get to this when they can.

silviuaavram:
I'm not a TS guru, so I am hoping someone can review this PR better.

ncesar:
just to leave a quick note, this is what `ButtonHTMLAttributes` looks like:
```ts
    interface ButtonHTMLAttributes<T> extends HTMLAttributes<T> {
        autoFocus?: boolean;
        disabled?: boolean;
        form?: string;
        formAction?: string;
        formEncType?: string;
        formMethod?: string;
        formNoValidate?: boolean;
        formTarget?: string;
        name?: string;
        type?: 'submit' | 'reset' | 'button';
        value?: string | string[] | number;
    }
```

silviuaavram:
Coming back to this. So, I've been thinking about this change, and I don't think it should go in.

Being an HTML element is something we can take for granted. But if it's a button, I'm not sure about that. That's up to the developer creating the UI. Maybe he wants a `<span>` instead, it's his call.

Since Downshift's scope is not how the UI is created, I would not change this. If you know, for sure, that your element is a button, you can cast it. But I would not make this assumption for all use cases.

Let me know what you think.

PR #720, opened 2019-06-19
Adding tests supporting function refs

atomicpages:
**What**: Adding test to show how a `ref` can be passed as a function to `getMenuProps`. 

**Why**: Docs are not clear about this feature. This is particularly useful when you need access to the menu element for positioning purposes using `getBoundingClientRect`.

**How**: Change exists already -- just adding a test to highlight the feature.

**Checklist**: Maybe?

- [ ] Documentation - N/A
- [x] Tests
- [x] Ready to be merged


silviuavram:
Hey, I think you are right, I probably encountered the same thing while working on stardust-ui with Downshift.

Should we also update the docs so it is more clear?

atomicpages:
Sure. I can contribute those as well

silviuaavram:
@all-contributors please add @atomicpages for test

PR #737, opened 2019-08-07
Add type definitions for scrollIntoView prop

benjamminj:
**What**:

Added a type definition for the `scrollIntoView` prop.

**Why**:

While the docs describe the type signature of `scrollIntoView`, not having it in the typings file makes TypeScript fail when using this prop.

**How**:

Added a new key with type definitions to the `DownshiftProps` definition in `typings`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged
      
Also, big thanks to everyone that maintains this library, using downshift has  made my life a _ton_ easier so very happy to give back a little 😎


silviuavram:
thank you for the contribution!

PR #745, opened 2019-08-16
Downshift prevents form submission on enter when defaultHighlightIndex is set

lewisl9029:
**What**:

`defaultHighlightIndex={0}` is useful for allowing users to quickly select the first suggestion in a typeahead just by pressing enter. However, we found that it has an unfortunate side-effect of preventing form submission on enter when there are no items in the suggestions list (i.e. when the user is trying to enter something that's not part of the suggestions list).

Here's a slightly modified version of the typeahead sandbox that changes the outer div to a form with a submit handler to help demonstrate the issue: 

https://codesandbox.io/s/downshift-examples-xgk4r?fontsize=14

1. First try removing `defaultHighlightedIndex={0}` and typing in 123 (or anything else that causes the suggestions list to become empty) and then pressing enter. You'll see the "not-submitted" text at the top of the page turn into "submitted", which is triggered by the custom onSubmit handler on the form.
2. Then try adding `defaultHighlightedIndex={0}` back in and doing the same. You'll see that not-submitted remains unchanged, meaning the onSubmit was never called.

**Why**:

This happens because the enter key handling for downshift will call `event.preventDefault()` to prevent form submission behavior and set the selected item even if the item at the highlighted index doesn't exist: https://github.com/downshift-js/downshift/blob/master/src/downshift.js#L599

**How**:

This PR simply adds an additional check to ensure the item at the highlighted index actually exists before preventing default and trying to change selection to it on enter.

Here's a version of the sandbox using a fork of downshift with the fix here applied: https://codesandbox.io/s/downshift-examples-1r460?fontsize=14

**Additional Considerations**:

I also just noticed a test that seems to be incompatible with the behavior change implemented in this PR: https://github.com/downshift-js/downshift/blob/master/src/__tests__/downshift.get-input-props.js#L537

And I'm not sure how to reconcile the two seemingly conflicting use cases. Would love to hear more thoughts/ideas on this.

**Checklist**:

- [ ] Documentation
N/A
- [ ] Tests
I tried to add a test for this but couldn't get it to pass (the onSubmit handler on the form never ends up getting called). Would love to get some 👀 on it to see if I'm doing something wrong or if there's another way to test this.
- [ ] Ready to be merged
      
### DIFF
--- a/README.md
+++ b/README.md
@@ -164,8 +164,8 @@ should be installed as one of your project's `dependencies`:
 npm install --save downshift
 ```
 
-> This package also depends on `react` and `prop-types`. Please make sure you
-> have those installed as well.
+> This package also depends on `react`. Please make sure you
+> have it installed as well.
 
 > Note also this library supports `preact` out of the box. If you are using
 > `preact` then use the corresponding module in the `preact/dist` folder. You


src/__tests__/downshift.get-input-props.js
```diff
@@ -385,6 +385,19 @@ test('enter on an input with an open menu does nothing without a highlightedInde
   expect(childrenSpy).not.toHaveBeenCalled()
 })
 
+test.only('enter on an input with an open menu with 0 items calls submit', () => {
```
weyert: Are you sure you want to use `.only` here?

lewisl9029: Nope, this definitely shouldn't be there. This was just here for ease of testing.

I can't seem to get this test to pass though. Would love to get some more eyes on this to figure out why, or if anyone has other suggestions on how to test this, I'd love to hear them too.

PR #747, opened 2019-08-18
feat(hooks): add useSelect hook

silviuavram:
Adding `useSelect`, the first hook of `Downshift`.

`Downshift` is often used to create custom `<select>` widgets. Even if it's designed as an autocomplete solution first, it proves it's a versatile component and we should aim to keep it this way.

As a follow-up to https://github.com/downshift-js/downshift/issues/683 I started the change to the hooks-based API by writing `useSelect`.

It aims to provide the same functionality and accessibility as a native HTML `<select>`. It should follow the [ARIA Dropdown Pattern](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html) for consistency reasons (it should work on Windows/Mac, with Jaws/NVDA/VoiceOver/ChromeVox no matter if the browser is Chrome/Safari/Firefox/Edge/IE).

The API is meant to resemble Downshift's.

Functionality provided out of the box:
- focus switch between the toggle button and the list. Combined with proper `aria` attributes on both these elements, widget should be accessible via all screen readers.
- toggle functionality with mouse, Enter, Arrow Keys and Space.
- selection with click, Enter and blur (by Tab, Shift-Tab and outside click).
- highlight option by typing character keys.
- a11y `aria-live` message on menu open and item select.
- highlight with Arrow keys that can be combined with Shift, End and Home.


rollup.config.js
```diff
@@ -8,6 +8,18 @@ config.plugins[cjsPluginIndex] = commonjs({
   include: 'node_modules/**',
   namedExports: {
     'react-is': ['isForwardRef'],
+    react: ['useReducer', 'useEffect', 'useRef', 'useState'],
+    'node_modules/keyboard-key/src/keyboardKey.js': ['getKey'],
```
mnajdova: Why was this added? Wouldn't this require all rollup clients to include this?

typings/index.d.ts
```diff
@@ -228,3 +228,136 @@ export type DownshiftInterface<Item> = React.ComponentClass<
...
+  FunctionOpenMenu = '__function_open_menu__',
+  FunctionCloseMenu = '__function_close_menu__',
+  FunctionSetHighlightedIndex = '__function_set_highlighted_index__',
+  FunctionSelectItem = '__function_select_item__',
+  FunctionClearKeysSoFar = '__function_clear_keys_so_far__',
+  FunctionReset = '__function_reset__',
+}
+
+export interface UseSelectProps<Item> {
+  items: Item[]
```
Daniel15: I think this could actually be `readonly Item[]` as Downshift doesn't mutate this array, right? What do you think about changing it?

PR #748, opened 2019-08-18
Simplify TS typings

### LINKED ISSUE #693: How to use DownshiftInterface?
It seems very convenient to have a type-parameterized Downshift, so I can check that the items I pass in/out of Downshift are of the correct item type.  Seems like there should be a simple answer, but what's the correct way to use this / to get a type-parameterized Downshift component?

ksocha:
This simplifies TypeScript typings and allows to use `Downshift` instance with generic parameter.

It's possibly a breaking change, because I've removed `DownshiftInterface` as it's no longer needed.

Fixes #693 

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/test/basic.test.tsx
+++ b/test/basic.test.tsx
@@ -1,8 +1,7 @@
 import * as React from 'react'
-import Downshift, {StateChangeOptions, DownshiftInterface} from '../'
+import Downshift, {StateChangeOptions} from '../'
 
 type Item = string
-const TypedDownShift: DownshiftInterface<Item> = Downshift
 
 interface Props {}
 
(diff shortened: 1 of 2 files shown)


silviuavram:
Can we fix this using something similar to the workaround in the issue attached?

I don't want to cause a breaking change for a type.

ksocha:
Current typings are just wrong and in my opinion they should be fixed. The change is breaking for people using TS, but adjusting code after the change is trivial.

silviuavram:
I will consider this for a future major release. Thank you for this!

CodeWitchBella:
Was asked for opinion :slightly_smiling_face:  I like this change :+1:

PR #761, opened 2019-09-11
fix(rollup): imports and revert added namedExports

silviuavram:
replaced `import * as X` with `import X from`. Also reverted added rollup named exports added by https://github.com/downshift-js/downshift/pull/747


src/hooks/useSelect/__tests__/getMenuProps.test.js
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable jest/no-disabled-tests */
-import * as keyboardKey from 'keyboard-key'
+import keyboardKey from 'keyboard-key'
```
layershifter: This is a correct change as `keyboard-key` is CJS module. It's the same issue as here: https://github.com/rollup/rollup/issues/1267#issuecomment-446681320

PR #781, opened 2019-10-04
Cleanup status message on unmount

beckerei:
**What**:
A fix for https://github.com/downshift-js/downshift/issues/769

**Why**:
The status message was not always cleaned up.

**How**:
running a cleanup function when component unmounts

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged
      
Running `npm run add-contributor` throws an error

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,59 +1,59 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 81533,
-    "minified": 37657,
-    "gzipped": 9489
+    "bundled": 81812,
+    "minified": 37895,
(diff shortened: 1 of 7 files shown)


silviuavram:
Hey, I looked over and this is breaking the cleanup. You removed the actual call of the cleanup debounced function and called the function (un-debounced) in willUnmount. Calling the `debounce(fn)` like that does nothing, as that call just returns another function (the debounced one) which can be called.

I thought that the problem is that the actual HTML element stays in the DOM, now I actually understand. Let's sync on the issue.

Issue #708, opened 2019-04-24
stateChangeTypes.keyDownArrowDown fires after initial up/down arrow press


roginfarrer:
- `downshift` version: 3.2.8

**What you did**:

I upgraded my version of downshift to v3.2.8 (up from 2.1.4), and discovered this bug.

**What happened**:

 I use `stateReducer` to modify the highlightedIndex, and I found that when I tabbed to the input with a selected item, and pressed up or down, the highlightedIndex was incorrect.

**Reproduction repository**:

https://codesandbox.io/s/vnyxvwk673

**Problem description**:

![Gif of recreated issue](https://cl.ly/c06ae6cb4116/ezgif.com-optimize.gif)

When you tab to the input, and the menu is not opened, and press the down arrow, the change type `Downshift.stateChangeTypes.keyDownArrowDown` fires twice. When the up arrow is pressed, it's still fired when after the `keyDownUpArrow` event fires. I expect it to both not fire when the up arrow is pressed or only once when the down arrow is pressed.

**Suggested solution**:

Haven't had a chance to debug this yet.

silviuavram:
Hello. The fact that you get two events in the reducer is expected. One comes with the `isOpen` changed, and the other one comes with the `highlightedIndex`. The reason for this is the fact that on opening we don't know anything about the items. Probably this can be improved because I also prefer only one state changed, with both `isOpen` and `highlightedIndex`. If you can look into this that would be nice.

The other part with on Up Arrow the second event comes as Arrow Down is probably a bug. This probably is easier to fix.

From your previous version to this one, the behaviour that changed is that on opening the dropdown by arrow up/down, we will also highlight the last/first item. It's in the ARIA spec on how the widget should work.

Let me know if you need any more info and if you're willing to fix this. That would be awesome! Thanks!

claytron5000:
I have a potential PR to deal with the double `keyDownArrowDown` or `Down` statechange. It includes a new stateChangeType. I'm having trouble running `setup`, that's another issue. When I get that setup I can offer a fix.

silviuavram:
please see my comment in https://github.com/downshift-js/downshift/pull/715

pranesh239:
I have raised a PR on this issue, please check.

silviuavram:
I will leave it as it is. Please use the new `useCombobox` if you want this behavior, as it will work.

Will be available as soon as I release 4.0.0.

https://downshift.netlify.com/hooks/use-combobox

PR #791, opened 2019-10-13
fix - duplicate and wrong event trigger on ArrowDown and ArrowUp

pranesh239:
**What**:
Fix for duplicate and wrong event call on ArrowDown and ArrowUp respectively.  #708 

**Why**:
If user has an event handler for ArrowDown it will be invoked twice and the event handler for ArrowUp will both ArrowUp and ArrowDown handlers. 

**How**:
Solved by removing the event type set at `setHighlightedIndex` of ArrowDown and ArrowUp handlers.

Still those events are called while setting state

```js
this.internalSetState(
          {
            isOpen: true,
            // changing the state of event type
            type: stateChangeTypes.keyDownArrowUp, 
          },() => // cb
```
Changes made: 
```diff
@@ -554,7 +554,6 @@ class Downshift extends Component {
                   this.getState().highlightedIndex,
                   itemCount,
                 ),
-                {type: stateChangeTypes.keyDownArrowDown},
               )
             }
           },

@@ -585,7 +584,6 @@ class Downshift extends Component {
                   this.getState().highlightedIndex,
                   itemCount,
                 ),
-                {type: stateChangeTypes.keyDownArrowDown},
               )
             }
           },
```

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged
      
Note: *this is my **first** open source contribution on fix, so please be patience if I made any mistake*

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,59 +1,59 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 81533,
-    "minified": 37657,
-    "gzipped": 9489
+    "bundled": 81215,
+    "minified": 37707,
+    "gzipped": 9505
   },
   "preact/dist/downshift.cjs.js": {
-    "bundled": 80158,
-    "minified": 36556,
-    "gzipped": 9379
+    "bundled": 79840,
+    "minified": 36606,
+    "gzipped": 9397
   },
   "preact/dist/downshift.umd.min.js": {
-    "bundled": 93608,
-    "minified": 31791,
-    "gzipped": 9759
+    "bundled": 93276,
+    "minified": 31879,
+    "gzipped": 9780
   },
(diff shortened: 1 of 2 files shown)


pranesh239:
*we need to have two state calls (one that will open, one that will set the highlighted index)* - 
In this case, rather setting the ArrowDown event twice, we can add two new types **OPEN** and **CLOSE**. When first time you're doing down arrow key press, it sets **OPEN** as type then ArrowDown type later. And sets **CLOSE** as type, when ever you close the dropdown.

In such way we can get rid of duplicate type set and also we will have more precise types.

silviuaavram:
I don't want to refactor the way it was working in `Downshift`. If you want this behavior please update to `useCombobox` which, because you are passing an `items` reference, will only make one state change for open + highlight.

PR #795, opened 2019-10-16
fix - Localize clearTimeout per useSelect mounted instances

### LINKED ISSUE #794: `keysSoFar` is only cleared on the last instance in the page
- `downshift` version: 3.3.4
- `node` version: 10.15.3
- `yarn` version:  1.13.0

**What you did**:

I placed more than one instance of dropdown using the useSelect hook and interacted with the first one. While collapsed and focused, I pressed A, it selected `Americium`, awaited a second and then pressed `P`

**What happened**:

`Americium` remained selected while it should select `Plutonium` (and it does on last instance)

**Reproduction repository**:

https://codesandbox.io/s/useselect-usage-cnv1p

**Problem description**:
`clearTimeout` should be inherent to the instance (by useRef), currently it holds the last mounted component's debounced `clearTimeout`.
https://github.com/downshift-js/downshift/blob/master/src/hooks/useSelect/index.js#L121-L135

**Suggested solution**:
```js
  const clearTimeout = useRef(null)
  useEffect(() => {
    // init the clean function here as we need access to dispatch.
    if (isInitialMount.current) {
      clearTimeout.current = debounce(() => {
        dispatch({
          type: stateChangeTypes.FunctionClearKeysSoFar,
        })
      }, 500)
    }
    if (!keysSoFar) {
      return
    }
    clearTimeout.current()
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [keysSoFar])
```

edygar:
**What**:

Localize clearTimeout per useSelect mounted instances

**Why**:

Closes #794 

Previously, clearTimeout was previously at root level closure, causing each new mounted component that used the useSelect hook to override the previous reference, so every instance of the component was alway calling only the last instance's clearTimeout, never calling their own clearTimeout and therefore, never clearing the `keysSoFar` state. 

**How**:

Assigned resulting clearTimeout to a ref, instead of the root closure variable

**Checklist**:

- [ ] Documentation (N/A)
- [ ] Tests (N/A)
- [x] Ready to be merged

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1074,6 +1074,16 @@
         "code",
         "test"
       ]
+    },
+    {
+      "login": "edygar",
+      "name": "Edygar Oliveira"
+      "avatar_url": "https://avatars.githubusercontent.com/u/566280?v=3",
+      "profile": "https://github.com/edygar",
+      "contributions": [
+        "code",
+        "bug"
+      ]
     }
   ]
 }
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,59 +1,59 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 81533,
-    "minified": 37657,
-    "gzipped": 9489
+    "bundled": 81604,
+    "minified": 37781,
+    "gzipped": 9518
   },
   "preact/dist/downshift.cjs.js": {
-    "bundled": 80158,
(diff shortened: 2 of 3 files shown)


silviuavram:
super cool, it's a great find, thanks very much! will merge this asap, but first run build to update the size-snapshot and commit that. and also a small nit, can you put the ref next in line with the other refs?

PR #799, opened 2019-10-22
fix(highlightedIndex): do not highlight disabled items by keyboard

silviuavram:
**What**:

Allows disabled items to be skipped by keyboard highlight (arrow keys, home, end, character keys).

**Why**:

To have this scenario properly working.

**How**:

Refactored the utils methods that handle the index generation logic for highlighting. These methods also request `getItemNodeFromIndex` that will tell them if the item is marked as disabled or not.

This method comes from the hooks / component and is sent as parameter to the utils functions.

Test cases:

- find next non-disabled if next is disabled on arrow down.
- find prev non-disabled if prev is disabled on arrow up.
- if not circular and no non-disabled item can be highlighted, index should stay the same.
- if circular, the search for non-disabled item is resumed from first / last item.

- on home/end, if first/last are disabled, then start searching downwards/upwards for first non-disabled.

- on character key (useSelect) it will not highlight if item is disabled, it will search instead for the next item that starts with the key(s) and is not disabled.

Additional changes:
- `useSelect` has `false` default prop for `circularNavigation`, while `useCombobox` has it `true.
- unit tests in getItemProps for `<Downshift>`.
- `useCombobox` now works properly with `initialHighlightedIndex` and `defaultHighlightedIndex`.
- full testing for the `getNextWrappingIndex` and `getItemIndexByCharacterKey`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged
      

Started to work based on https://github.com/downshift-js/downshift/pull/752

Closes https://github.com/downshift-js/downshift/issues/728 https://github.com/downshift-js/downshift/issues/825

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,59 +1,59 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 107604,
-    "minified": 50106,
-    "gzipped": 11240
+    "bundled": 111980,
+    "minified": 51395,
+    "gzipped": 11470
   },
   "preact/dist/downshift.cjs.js": {
-    "bundled": 106322,
-    "minified": 49068,
-    "gzipped": 11125
+    "bundled": 110698,
+    "minified": 50357,
+    "gzipped": 11366
   },
   "preact/dist/downshift.umd.min.js": {
-    "bundled": 118383,
-    "minified": 38903,
-    "gzipped": 11087
+    "bundled": 122933,
+    "minified": 39994,
+    "gzipped": 11322
   },
   "preact/dist/downshift.umd.js": {
-    "bundled": 134506,
-    "minified": 46407,
-    "gzipped": 12880
+    "bundled": 139056,
+    "minified": 47496,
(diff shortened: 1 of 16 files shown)


weyert:
How can we progress this PR? Would it be possible to merge this PR with the logic for vanilla downshift and have the useSelect scenarios as enhancement?

silviuavram:
@weyert we should be close with this. I only need to add tests to the hooks and finish the PR description.

weyert:
Which tests need to be added? Looks like it already has 100% coverage?

silviuavram:
I was not 100% confident even with that coverage so added separate tests for the index generating function. Now it's better.

weyert:
Amazing, thanks @silviuavram :D

weyert:
Awesome! Thank you, excited about this 😀

silviuavram:
@all-contributors please add @jgodi for code

PR #801, opened 2019-10-22
Remove status div when component unmounts

beckerei:
**What**:
Fixes https://github.com/downshift-js/downshift/issues/769

**Why**:
Leaving a clean state after Downshift is no longer present

**How**:

Removes the node from the DOM on unmount.

I decided to use `removeChild` instead of `remove`, otherwise a polyfill for IE would be required: https://caniuse.com/#feat=childnode-remove
Not sure if IE is a target of downshift, but using `remove` would be a breaking change.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged
      
   That still throws an error for me ;)


src/__tests__/set-a11y-status.js
```diff
@@ -5,38 +5,45 @@ beforeEach(() => {
...
+test('removes the status div', () => {
+  const {default: setA11yStatus, removeStatusDiv} = setup()
+  setA11yStatus('hello')
+  removeStatusDiv()
+  expect(document.body.firstChild).toBe(null)
+})
+
 test('replaces the status with a different one', () => {
-  const setA11yStatus = setup()
+  const {default: setA11yStatus} = setup()
```
beckerei: this is a bit ugly, but I don't know how to import the default and the named export with require.

PR #808, opened 2019-10-27
fix(preact): remove hooks from build

francoischalifour:
## What

This removes the Downshift hooks from the Preact build because they're not compatible with the main Preact package.

## Why

The [React ESM build](https://unpkg.com/browse/downshift@3.4.0/dist/downshift.esm.js) generates this import:

```js
import { cloneElement, Component, useCallback, useReducer, useState, useEffect, useRef } from 'preact';
```

The current [Preact build](https://unpkg.com/browse/downshift@3.4.0/preact/dist/downshift.esm.js) generated this import:

```js
import { cloneElement, Component, useCallback, useReducer, useState, useEffect, useRef } from 'preact';
```

However, Preact hooks are exported from `preact/hooks` (not `preact`). Since they don't exist in `preact`, it throws in some environments (e.g. building a library with Rollup).

## How

This PR results in this import in the Preact ESM bundle:

```js
import { cloneElement, Component } from 'preact';
```

Not exporting Downshift hooks in the Preact for now has two benefits:

- **Makes the Preact build work in all environments**
- **Decreases the bundle size from 80Kb to 52Kb**

> Note that having added hooks to the main build without bumping the `react` peer dependency makes it error for people using React < 16 as well as Preact users.

## Checklist

- [x] Ready to be merged

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -5,55 +5,55 @@
     "gzipped": 9598
   },
   "preact/dist/downshift.cjs.js": {
-    "bundled": 81090,
-    "minified": 36994,
-    "gzipped": 9498
+    "bundled": 53373,
+    "minified": 22913,
+    "gzipped": 6425
   },
   "preact/dist/downshift.umd.min.js": {
-    "bundled": 93792,
-    "minified": 32003,
-    "gzipped": 9813
+    "bundled": 66232,
+    "minified": 22631,
+    "gzipped": 7213
   },
   "preact/dist/downshift.umd.js": {
-    "bundled": 107824,
-    "minified": 38043,
-    "gzipped": 11343
+    "bundled": 79745,
+    "minified": 28135,
+    "gzipped": 8660
   },
   "dist/downshift.umd.min.js": {
-    "bundled": 98431,
-    "minified": 33321,
-    "gzipped": 10377
+    "bundled": 98773,
+    "minified": 33403,
+    "gzipped": 10417
   },
   "dist/downshift.umd.js": {
-    "bundled": 136984,
-    "minified": 46941,
-    "gzipped": 13963
+    "bundled": 137346,
+    "minified": 47043,
+    "gzipped": 14012
   },
   "dist/downshift.esm.js": {
-    "bundled": 81619,
-    "minified": 37532,
-    "gzipped": 9478,
+    "bundled": 82000,
+    "minified": 37733,
+    "gzipped": 9539,
     "treeshaked": {
       "rollup": {
         "code": 629,
         "import_statements": 303
       },
       "webpack": {
-        "code": 27541
+        "code": 27621
       }
     }
   },
   "preact/dist/downshift.esm.js": {
-    "bundled": 80309,
-    "minified": 36465,
-    "gzipped": 9374,
+    "bundled": 53071,
+    "minified": 22681,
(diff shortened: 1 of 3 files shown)


silviuavram:
Instead of removing the hook from preact, can you please describe what we need to do in order to make the hook compatible with preact?

francoischalifour:
@silviuavram We would need to transform the import to:

```js
import { cloneElement, Component } from 'preact';
import { useCallback, useReducer, useState, useEffect, useRef } from 'preact/hooks';
```

I'm not too sure how to transform the import to that though. Since it doesn't fully work with Preact anyway, I reckoned this solution would be temporary.

What do you think about having the hooks in another Downshift package? (e.g. `downshift/hooks`)

silviuavram:
Then why remove the hooks if the issue is also with:

`import { cloneElement, Component } from 'preact';`

Which is not related to hooks.

silviuavram:
There must be another way, I need that hook in Downshift for now. They say Preact X alpha works with hooks, so as long as `cloneElement, Component` are handled correctly (react->preact), the hooks should too.

francoischalifour:
The issue is not with:

```js
import { cloneElement, Component } from 'preact';
```

It's with:

```js
import { useCallback, useReducer, useState, useEffect, useRef } from 'preact';
```

Because `use*` are not exported by `preact` but by `preact/hooks`.

francoischalifour:
The reason I didn't dig into a better solution for now is because the Rollup config comes from [`kcd-scripts`](https://github.com/kentcdodds/kcd-scripts/blob/a3131186ce02756aa919779eb25f8ebca7da33fe/src/config/rollup.config.js) which abstracts everything. It's hard to understand how to transform the imports from there. I figured instead of taking too much time digging into that dependency, removing them temporarily is fine because they don't work in the Preact build anyway.

silviuavram:
https://github.com/developit/preact-compat can this help us?

Yes, it comes from kcd-scripts, but we can also add what we need to it.

We should have the hooks in Preact as well, as I mentioned before, I want to release one more hook soon.

francoischalifour:
Yes, another solution would be to depend on `preact/compat` (the new version of `preact-compat`). Preact Compat is a Preact bundle with the same interface as React, so it works as expected.

The reason I didn't use this import it is because we don't get the benefits of Preact being very lightweight.

As an example:

- My library + (Downshift + `preact`): 46.2 Kb 
- My library + (Downshift + `preact/compat`): 48.4 Kb

It might be negligible compared to the value that hooks brings, so we can go with that solution if you prefer!

silviuavram:
I would go with the `preact/compat`. I think exposing what this library has to offer is more important than size, for now. We can improve on size afterwards.

silviuavram:
We should contact @kentcdodds and see what he thinks. Is there another option we can apply the changes only here?

francoischalifour:
I guess it should be possible to batch the config to use `preact/compat` as an alias with [`babel-plugin-module-resolver`](https://github.com/tleunen/babel-plugin-module-resolver).

Let's see what @kentcdodds thinks about it!

kentcdodds:
I prefer the babel plugin approach 👍

vpzomtrrfrt:
Does this mean Downshift is dropping native Preact support?

kentcdodds:
No

PR #816, opened 2019-11-02
Add onItemSelected callback prop for useSelect

gstauffer-toast:
**What**: This PR addresses https://github.com/downshift-js/downshift/issues/793

**Why**: This provides a callback for users to see when an item is selected in the dropdown (even if it was the same item).

**How**: I added a new parameter to `useEnhancedReducer` called `performParentCallbacks`.

This allows the parent hook to pass in some of its own callbacks into the enhanced reducer. It takes two parameters, because that's all I need to use in the current implementation:
- `action`: The action performed, defined by parent reducer. This allows us to the action's type, which are `stateChangeTypes` defined within `useSelect`
- `newState`: The outgoing state from the reducer

In hooks/useSelect, I added a new file: `propCallbacks.js`. This file will be passed into useEnhancedReducer, and allows prop callbacks to be defined on `useSelect` that the enhanced reducer may not know about

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged
      

I feel like this is a little messy, so I'm happy to hear thoughts on this. It seems a little funky to have both `onSelectedItemChange` and `onItemSelected` defined on the same component.

I also didn't use Commitizen for the commits.. should I have done that? I can copy over with new commit messages if we feel these changes are OK.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -15,9 +15,9 @@
     "gzipped": 9935
   },
   "preact/dist/downshift.umd.js": {
-    "bundled": 109019,
-    "minified": 38372,
-    "gzipped": 11471
+    "bundled": 108855,
+    "minified": 38364,
+    "gzipped": 11469
   },
   "dist/downshift.umd.min.js": {
     "bundled": 99442,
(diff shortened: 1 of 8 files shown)


silviuavram:
I don't understand all the changes. I thought you needed `onSelect` that will get triggered at any select, so add that as prop and call it in `utils.js` next to `onStateChange`, if there is any `selectedItem` in changes.

PR #821, opened 2019-11-12
docs(useSelect hook): add `refKey` prop description for getItemProps

mamackenzie:
**What**: In the docs for the `useSelect` hook, I've added an item to the optional properties list of `getItemProps`, as it was not clear from the documentation at the time that this was necessary, and I had to read the `useSelect` hook source code to find this solution.

**Why**: In developing an accessible dropdown menu with Downshift using composite components, I noticed that I needed to set the `refKey` attribute not only on the composite _menu_ component, but also on the composite _item_ components as well. Without setting this in both places, the code did not work and the docs did not explain anything about this.

**How**: A small edit to the docs.

**Checklist**:

- [x] Documentation
- [ ] Tests (N/A)
- [x] Ready to be merged

### DIFF
--- a/src/hooks/useSelect/README.md
+++ b/src/hooks/useSelect/README.md
@@ -519,6 +519,7 @@ The main difference from vanilla `Downshift` is that we require the items as pro
 Optional properties:
 
 - `disabled`: If this is set to `true`, then all of the downshift item event handlers will be omitted. Items will not be highlighted when hovered, and items will not be selected when clicked.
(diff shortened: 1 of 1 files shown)


silviuavram:
I have a feeling that we should generalise this. Probably we should have a separate section in the Readme with that description, and from the getter props where this applies (almost all of them) you will write a few words why `refKey` would be needed and then have a link to that section with a `see more` text or something.

silviuaavram:
This comment has been added in the meantime. Thank you for the PR anyway! Will close it.

PR #828, opened 2019-11-15
Doczrc Issue

EOstling:
The Doczrc is in a broken state so added regex to pattern match the specific extension.

The current state is broken

Simple Regex

**Checklist**:

- [ X] Documentation
- [ X] Tests
- [ X?] Ready to be merged
      

This is definitely not easy to test, so seeing how it behaves after the change should give us some clue as to where the issue is if it does persist.


silviuavram:
what do you mean by broken?

PR #830, opened 2019-11-19
Run stateReducer over stateToSet passed to onInputValueChange

### LINKED ISSUE #819: "stateAndHelpers" param within "onInputValueChange" function does not align with desired state
- `downshift` version: 3.4.1
- `node` version: 10.9.0
- `yarn` version: 1.17.3

Potentially related to #759 

**Problem description**:

The `stateAndHelpers` param from the `onInputValueChange` function doesn't actually represent my desired state _in certain circumstances_

This was introduced in #217 to fix the cursor position issue however the state reducer was never called against the `stateToSet`

```js
if (!isStateToSetFunction && stateToSet.hasOwnProperty('inputValue')) {
  this.props.onInputValueChange(stateToSet.inputValue, {
    ...this.getStateAndHelpers(),
    ...stateToSet,
  })
}
```

Further down, if the `stateToSet` is a function then the `onInputValueChange` function isn't called until later

```js
this.props.onInputValueChange(newStateToSet.inputValue, {
  ...this.getStateAndHelpers(),
  ...newStateToSet,
})
```

Where `newStateToSet` has been passed through the `stateReducer`

```js
newStateToSet = this.props.stateReducer(state, newStateToSet)
```

This means that depending on what is passed to the `internalSetState` function (a state object or state function) then the `stateAndHelpers` of the `onInputValueChange` function will have different values.

**Suggested solution**:

Call `stateReducer` against the `stateToSet` before passing it to the `onInputValueChange` function

```js


if (!isStateToSetFunction) {
  const newStateToSet = this.stateReducer(this.state, stateToSet)
  if (newStateToSet.hasOwnProperty('inputValue')) {
    this.props.onInputValueChange(newStateToSet.inputValue, {
      ...this.getStateAndHelpers(),
      ...newStateToSet,
    })
  }
}
```

This will allow me to continue to rely on the downshift state as a source of truth (I am manipulating the `selectedItem` of a multi-select to **always** be an array)

If there is no other considerations that I may have missed with this fix then I can submit a PR.

aaronnuu:
**What**: Fixes #819 

**Why**: Right now depending on how `internalSetState` is being called (as a state object or function) the `stateAndHelpers` param has either been run through the `stateReducer` or it hasn't, this PR addresses this and makes both implementations work the same.

**How**: I run the `stateToSet` through the `stateReducer` and pass that state to the `onInputValueChange` function, exactly how it's being done within the functional `setState` just below.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged
      

I have added a couple of tests that make sure the bugs that I identified aren't a problem anymore, not sure if they're done in the best way or not though so any feedback on that one would be appreciated.

### DIFF
--- a/src/__tests__/downshift.lifecycle.js
+++ b/src/__tests__/downshift.lifecycle.js
@@ -261,6 +261,7 @@ test('selecting an item calls onInputValueChange with reduced state', () => {
     onInputValueChange: onInputValueChangeSpy,
   })
   selectItem('bar')
+  expect(onInputValueChangeSpy).toHaveBeenCalledTimes(1)
   expect(onInputValueChangeSpy).toHaveReturnedWith(
     expect.objectContaining({selectedItem: 'foo'}),
   )
@@ -281,6 +282,7 @@ test('onInputValueChange is called when reduced state includes inputValue', () =
   })
   openMenu()
   expect(onInputValueChangeSpy).toHaveBeenCalledTimes(1)
+  expect(onInputValueChangeSpy).toHaveBeenCalledWith('baz', expect.any(Object))
 })
 
 test('onInputValueChange is not called when reduced state does not include inputValue', () => {
@@ -295,7 +297,7 @@ test('onInputValueChange is not called when reduced state does not include input
     onInputValueChange: onInputValueChangeSpy,
   })
   selectItem('bar')
-  expect(onInputValueChangeSpy).toHaveBeenCalledTimes(0)
+  expect(onInputValueChangeSpy).not.toHaveBeenCalled()
 })
 
 function mouseDownAndUp(node) {
--- a/src/__tests__/downshift.lifecycle.js
+++ b/src/__tests__/downshift.lifecycle.js
@@ -250,6 +250,54 @@ test('controlled highlighted index change scrolls the item into view', () => {
   )
 })
 
+test('selecting an item calls onInputValueChange with reduced state', () => {
+  const stateReducer = (_, changes) => ({
+    ...changes,
+    selectedItem: 'foo',
+  })
+  const onInputValueChangeSpy = jest.fn((_, stateAndHelpers) => stateAndHelpers)
+  const {selectItem} = setup({
+    stateReducer,
+    onInputValueChange: onInputValueChangeSpy,
+  })
+  selectItem('bar')
+  expect(onInputValueChangeSpy).toHaveReturnedWith(
+    expect.objectContaining({selectedItem: 'foo'}),
+  )
+})
+
+test('onInputValueChange is called when reduced state includes inputValue', () => {
+  const stateReducer = (_, changes) => {
+    return {
+      ...changes,
+      inputValue: 'baz',
+    }
+  }
+  const onInputValueChangeSpy = jest.fn(() => null)
+  const {openMenu} = setup({
+    stateReducer,
+    inputValue: 'foo',
+    onInputValueChange: onInputValueChangeSpy,
+  })
+  openMenu()
+  expect(onInputValueChangeSpy).toHaveBeenCalledTimes(1)
+})
+
+test('onInputValueChange is not called when reduced state does not include inputValue', () => {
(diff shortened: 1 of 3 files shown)


silviuaavram:
It took me too much, but arrived at this issue and aiming to close it. @aaronnuu are you still interested in merging this? If so I have a few comments, if not i will create a PR myself. issue makes sense, so we should merge the reviewed changes.

aaronnuu:
I haven't been following the hooks development very closely so if it makes more sense to open a new PR I'm not against it, otherwise yeah still interested in helping out.

src/__tests__/downshift.lifecycle.js
```diff
@@ -250,6 +250,54 @@ test('controlled highlighted index change scrolls the item into view', () => {
...
+    ...changes,
+    selectedItem: 'foo',
+  })
+  const onInputValueChangeSpy = jest.fn((_, stateAndHelpers) => stateAndHelpers)
+  const {selectItem} = setup({
+    stateReducer,
+    onInputValueChange: onInputValueChangeSpy,
+  })
+  selectItem('bar')
+  expect(onInputValueChangeSpy).toHaveReturnedWith(
```
silviuaavram, line 264: we should also check the times it got called.

src/__tests__/downshift.lifecycle.js
```diff
@@ -250,6 +250,54 @@ test('controlled highlighted index change scrolls the item into view', () => {
...
+    }
+  }
+  const onInputValueChangeSpy = jest.fn(() => null)
+  const {openMenu} = setup({
+    stateReducer,
+    inputValue: 'foo',
+    onInputValueChange: onInputValueChangeSpy,
+  })
+  openMenu()
+  expect(onInputValueChangeSpy).toHaveBeenCalledTimes(1)
```
silviuaavram, line 283: we should also check the parameters it got called with.

src/__tests__/downshift.lifecycle.js
```diff
@@ -250,6 +250,54 @@ test('controlled highlighted index change scrolls the item into view', () => {
...
+    return changes
+  }
+  const onInputValueChangeSpy = jest.fn(() => null)
+  const {selectItem} = setup({
+    stateReducer,
+    inputValue: 'foo',
+    onInputValueChange: onInputValueChangeSpy,
+  })
+  selectItem('bar')
+  expect(onInputValueChangeSpy).toHaveBeenCalledTimes(0)
```
silviuaavram, line 298: .not.toHaveBeenCalled()

src/downshift.js
```diff
@@ -341,11 +341,14 @@ class Downshift extends Component {
     // the input change as soon as possible. This avoids issues with
     // preserving the cursor position.
     // See https://github.com/downshift-js/downshift/issues/217 for more info.
-    if (!isStateToSetFunction && stateToSet.hasOwnProperty('inputValue')) {
-      this.props.onInputValueChange(stateToSet.inputValue, {
-        ...this.getStateAndHelpers(),
-        ...stateToSet,
-      })
+    if (!isStateToSetFunction) {
+      const newStateToSet = this.props.stateReducer(this.state, stateToSet)
```
silviuaavram, line 345: no reason to call this unless `newStateToSet.hasOwnProperty('inputValue')` so I would leave the condition as it was and call this before calling `onInputValueChange`

aaronnuu: I had it this way originally but ran into some issues since the user can add or remove `inputValue` with the `stateReducer` this ensures that `onInputValueChange` will be called (or not called) if they add / remove it

aaronnuu: I might add a comment for this because it's a bit of a gotcha

aaronnuu:
@silviuaavram Sorry it took a couple of weeks, have made those changes and re-run the snapshot. Wasn't sure how to replicate the helper object that gets sent as the second param to the `onInputValueChange` function so just verifying it's any object.

PR #829, opened 2019-11-17
fix(useSelect): respect disabled option passed to `getToggleButtonProps`

maxmalov:
**What**: bug

**Why**: [documentation](https://github.com/downshift-js/downshift/tree/master/src/hooks/useSelect#gettogglebuttonprops) states `getToggleButtonProps` accepts `disabled` option, but actually it doesn't

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged

PR #833, opened 2019-11-23
fix(useSelect): respect disabled option passed to `getItemProps`

mateuszkrzak:
**What**:
`disabled` property is ignored when passed to `getItemProps`

**Why**:
Not respecting `disabled` property causes unwanted handlers to trigger on hover and click (opposite to the what [docs](https://github.com/downshift-js/downshift/tree/master/src/hooks/useSelect#getitemprops) says)

**How**:
I followed the convention that @maxmalov used to fix similar bug (#829), but in `getToggleButtonProps`. 

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,59 +1,59 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 82078,
-    "minified": 38025,
-    "gzipped": 9601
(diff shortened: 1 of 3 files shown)


silviuavram:
Looks great! Merging!

PR #841, opened 2019-11-28
perf(getItemProps): memoize the event handlers for item

silviuavram:
The main reason for doing this is because we are calling `getItemProps` to apply attributes / handlers and also to supply Downshift with items. If we want to avoid calling it so that we don't re-render the item, Downshift on re-render will throw away the items references and will not know about them anymore.

Since there is no way to avoid calling this with current `Downshift` implementation, we should at least provide the event handlers as memoized functions, so if your item is a pure component and it does not have to change, it will not.

Probably we should use this approach with the rest of the handlers as well, will need to look into it soon.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,59 +1,59 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 82171,
-    "minified": 38051,
-    "gzipped": 9612
+    "bundled": 83479,
+    "minified": 38653,
+    "gzipped": 9793
   },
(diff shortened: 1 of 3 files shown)


silviuavram:
thanks @smashercosmo but we have many items in the array, and memoize-one memoizes just one, it invalidates the result all the time. But I decided to use fast-memoize instead, should be better. What do you think?

smashercosmo:
yes, in terms of size fast-memoize is much better (3x smaller than lodash.memoize)

PR #850, opened 2019-12-09
fix: treeshake Downshift and useSelect

silviuavram:
`defaultProps` and `stateChangeTypes` prevent Downshift from being tree-shakeable. Also the prop type checks in `useSelect`.

Use `no-side-effect-class-properties` in `.babelrc.js` along with the original configs from `kcd-scripts`. https://github.com/naver/babel-plugin-no-side-effect-class-properties @kentcdodds maybe you want to include this as well or mark it to be added once a more official version is released.

Updated `rollup.config.js` to use the babel config file.

Only check hook prop-types if not on production.

Result: treeshakeable Downshift.

Most grateful to @layershifter for contributing with this.

https://github.com/babel/babel/issues/8592
https://github.com/babel/babel/pull/6963

### DIFF
--- a/src/hooks/useSelect/index.js
+++ b/src/hooks/useSelect/index.js
@@ -17,6 +17,7 @@ import {
   debounce,
   scrollIntoView as defaultScrollIntoView,
   normalizeArrowKey,
+  noop,
 } from '../../utils'
 import downshiftSelectReducer from './reducer'
 import {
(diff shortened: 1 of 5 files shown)


src/hooks/useSelect/index.js
```diff
@@ -27,7 +28,10 @@ import {
 } from './utils'
 import * as stateChangeTypes from './stateChangeTypes'
 
-const validatePropTypes = getPropTypesValidator(useSelect, propTypes)
+const validatePropTypes =
+  process.env.NODE_ENV === 'production'
+    ? noop
+    : getPropTypesValidator(useSelect, propTypes)
```
layershifter: I rather recommend to use a condition in `useSelect` as in this case `validatePropTypes` function will be unused and removed:

```jsx
if (process.env.NODE_ENV !== 'production') {
  validatePropTypes(userProps)
}
```

silviuavram:
@all-contributors please add @layershifter with code, infra, ideas

silviuavram:
@all-contributors please add @epiqueras with ideas

PR #857, opened 2019-12-12
release 4.0.0 (#812)

silviuavram:
Release `useCombobox` hook, adapts the TS typings and improves the docsite.

The `useCombobox` should be the way forward when it comes to building an autocomplete component.
Structure the docsite as a Hooks menu with a submenu for each hook. Each hook should be represented on a full page.
Details in https://github.com/downshift-js/downshift/pull/812.

**BREAKING CHANGE**:  Remove `DownshiftInterface` as it's not required anymore to make `Downshift` props generic.You can pass your <Item> generic to the Downshift props as `<Downshift<MyItem> {...props} />`. https://github.com/downshift-js/downshift/pull/748

**BREAKING CHANGE**: Change TS typings for `useSelect`, in getter props, to specify that most of them will require `refKey` in case the user renders a custom React component with a ref prop that has a different name than `ref` (like `innerRef` for instance). https://github.com/downshift-js/downshift/pull/812
```js
  getToggleButtonProps: (options?: UseSelectGetMenuPropsOptions) => any
  getLabelProps: (options?: UseSelectGetLabelPropsOptions) => any
  getMenuProps: (options?: UseSelectGetToggleButtonPropsOptions) => any
  getItemProps: (options: UseSelectGetItemPropsOptions<Item>) => any
```

 Each new `UseSelect` type is like:
```js
export interface UseSelectGetToggleButtonPropsOptions
  extends GetPropsWithRefKey,
    GetToggleButtonPropsOptions {}
```


silviuavram:
Will re-release this to 4.0.0.

PR #869, opened 2019-12-25
fix(useSelect): add support for space button on menu

lukyth:
**What**:

Making the menu on `useSelect` behave the same way when pressing spacebar or enter.

**Why**:

`<select>` also behave like that. See https://github.com/downshift-js/downshift/issues/853 for more detail.

**How**:

- Add a new `stateChangeTypes`, `MenuKeyDownSpaceButton`
- Dispatch it when receiving `' '` key on menu
- Handle that action in the reducer the same way `MenuKeyDownEnter` is handling.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged

### DIFF
--- a/src/hooks/useSelect/index.js
+++ b/src/hooks/useSelect/index.js
@@ -198,6 +198,12 @@ function useSelect(userProps = {}) {
         type: stateChangeTypes.MenuKeyDownEnter,
       })
     },
+    ' '(event) {
+      event.preventDefault()
+      dispatch({
+        type: stateChangeTypes.MenuKeyDownSpaceButton,
+      })
+    },
     Tab(event) {
       // The exception that calls MenuBlur.
       // istanbul ignore next
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,38 +1,38 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 107604,
-    "minified": 50106,
-    "gzipped": 11240
+    "bundled": 107924,
+    "minified": 50343,
+    "gzipped": 11289
   },
   "preact/dist/downshift.cjs.js": {
-    "bundled": 106322,
-    "minified": 49068,
-    "gzipped": 11125
+    "bundled": 106642,
+    "minified": 49305,
+    "gzipped": 11176
   },
   "preact/dist/downshift.umd.min.js": {
-    "bundled": 118383,
-    "minified": 38903,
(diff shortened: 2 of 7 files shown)


src/hooks/useSelect/index.js
```diff
@@ -198,6 +198,12 @@ function useSelect(userProps = {}) {
         type: stateChangeTypes.MenuKeyDownEnter,
       })
     },
+    ' '(event) {
```
weyert: I am wondering if whether it would be better to use `Space` instead of `' '` here?

silviuavram: we should be safe here according to https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values

Although I would probably name it `Spacebar` instead of `SpaceButton` in all the occurrences in this PR. For example `stateChangeTypes.MenuKeyDownSpacebar`. What do you think?

silviuavram:
Awesome work, I love it! Take a look at our comments and let us know what you think.

src/hooks/useSelect/index.js
lukyth: `Space` wouldn't work for some reason. I guess that's why it's `' '` on another place as well.

https://github.com/downshift-js/downshift/blob/84b6ff6046507a8ea3cd783e005778e97d2a9674/src/downshift.js#L626

I also use `SpaceButton` because I found that name being used else where.

https://github.com/downshift-js/downshift/blob/84b6ff6046507a8ea3cd783e005778e97d2a9674/src/downshift.js#L628

But `Spacebar` does make more sense to me TBH. I could update them all if you think that's okay.

silviuavram: It does make more sense. But thinking about it a bit more, in order to be consistent, we need to deliver a breaking change, and change those occurrences as well.

Nobody complained about it so far, let's leave it like this.

silviuavram: Can you merge latest master, remove size snapshot, run build locally and update that size snapshot and commit it?

lukyth: Sorry I just read your reply. Seems like you already merge master yourself?

silviuavram:
@all-comtributors add @lukyth for code, bug

silviuavram:
@all-contributors please add @lukyth for bug, code

PR #875, opened 2020-01-04
perf(getItemProps): getMemoizedItemHandlers prop to return memoized handlers

silviuavram:
**What**:
Prop `getMemoizedItemHandlers` that should be used to memoize the event handlers returned by `getItemProps`.

**Why**:
`getItemProps` should be called at each re-render, but receiving new event handlers means that the items themselves need to re-render all the time. If the item is a PureComponent and its position in the list has not changed, then it makes sense for it not to re-render, so providing a memoized version of those handlers makes sense.

Having many items as React components that are expensive to render will make Downshift appear as a slow library. We are not giving any choice so far in this matter, as we always return new callbacks and the items will always be re-rendered. But not anymore.

**How**:
`getMemoizedItemHandlers` is a prop that is called with a callback to get the handlers, with the item itself and its index in the list. The last two items are enough to make a decision on whether to return the handlers memoized or new ones. The serializer can be `${item.id}${index}` as it is important for both the item itself and its position in the list to be the same in order to return a memoized version of the handlers.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [ ] useSelect
- [ ] useCombobox
- [ ] Ready to be merged
      

Fixes https://github.com/downshift-js/downshift/issues/845
Closes https://github.com/downshift-js/downshift/issues/827

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,45 +1,59 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 113603,
-    "minified": 52349,
-    "gzipped": 11598
+    "bundled": 114798,
+    "minified": 52655,
+    "gzipped": 11620
   },
   "preact/dist/downshift.cjs.js": {
-    "bundled": 112278,
-    "minified": 51272,
-    "gzipped": 11492
+    "bundled": 113473,
+    "minified": 51578,
+    "gzipped": 11512
   },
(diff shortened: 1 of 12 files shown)


silviuavram:
I will probably not change this in the end, but may add docs to describe the problem and a possible solution.

PR #882, opened 2020-01-07
fix(useCombobox): pass down missing disabled prop in prop getters

jorgemoya:
**What**:

Currently, `getToggleButtonProps` & `getItemProps` accept a `disabled` prop to remove handlers but won't pass that prop down.

**Why**:

Allows components to be styled dependent on the `disabled` prop. Accessibility?

**How**:

Simply pass down the `disabled` prop.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] TypeScript Types N/A
- [ ] Flow Types N/A
- [x] Ready to be merged
      

Added tests for both `useSelect` and `useCombobox`.

### DIFF
--- a/src/hooks/useCombobox/__tests__/getToggleButtonProps.test.js
+++ b/src/hooks/useCombobox/__tests__/getToggleButtonProps.test.js
@@ -32,6 +32,7 @@ describe('getToggleButtonProps', () => {
       })
 
       expect(toggleButtonProps.onClick).toBeUndefined()
+      expect(toggleButtonProps.disabled).toBe(true)
     })
   })
 
--- a/src/hooks/useCombobox/index.js
+++ b/src/hooks/useCombobox/index.js
@@ -262,7 +262,6 @@ function useCombobox(userProps = {}) {
     onMouseMove,
     onClick,
     onPress,
-    disabled,
     ...rest
   } = {}) => {
     const itemIndex = getItemIndex(index, item, items)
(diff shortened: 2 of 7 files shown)


src/hooks/useCombobox/__tests__/getToggleButtonProps.test.js
```diff
@@ -32,6 +32,7 @@ describe('getToggleButtonProps', () => {
       })
 
       expect(toggleButtonProps.onClick).toBeUndefined()
+      expect(toggleButtonProps.disabled).toBeDefined()
```
silviuavram: we can expect this to be true.

src/hooks/useCombobox/index.js
```diff
@@ -309,6 +309,7 @@ function useCombobox(userProps = {}) {
       [refKey]: handleRefs(ref, toggleButtonNode => {
         toggleButtonRef.current = toggleButtonNode
       }),
+      disabled,
```
silviuavram: actually, for consistency, can we remove the `disabled` from the props spread, use `rest.disabled` for the check and then it will get passed automatically?

silviuavram:
This is a very good catch, thank you! Let's move forward with this and get it merged.

Can you please fix the other case in the same PR so we have everything added in the same release?

Also if you want you can add the `disabled` prop test to all other existing cases? There should also be cases in `useSelect` where I am sure we are not testing if `disabled` is passed and true.

Thank you again for these changes! Let me know what you think.

One last thing. After all the changes can you run `build` locally and commit the new size snapshot?

silviuavram:
@all-contributors please add @jorgemoya with code, bug.

PR #906, opened 2020-01-22
fix(useSelect): use mouse and touch events to trigger blur

jorgemoya:
**What**:

Much like in #893, I need to listen to mouse and touch events in the menu. 

**Why**:

If any of the items contains an element that is focusable, the menu will immediately blur. I want to keep the menu open and allow `onClick` event handlers for said elements to be triggered. I particularly had an issue with adding `a` elements for some of my items; clicking on them would close the menu and not open the link, as expected. I also want to be able to right click on them.

**How**:

I'm reusing the logic that was added in #893, now listening for mouse and touch events as well as checking if the `target` is within Downshift.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] TypeScript Types N/A
- [ ] Flow Types N/A
- [x] Ready to be merged
      

Added some of the missing tests that were included in #893 for useCombobox.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,38 +1,13 @@
 {
-  "dist/downshift.cjs.js": {
-    "bundled": 115780,
-    "minified": 53146,
-    "gzipped": 11693
-  },
-  "preact/dist/downshift.cjs.js": {
-    "bundled": 114501,
-    "minified": 52111,
-    "gzipped": 11589
-  },
-  "preact/dist/downshift.umd.js": {
-    "bundled": 142962,
-    "minified": 48856,
-    "gzipped": 13285
-  },
-  "preact/dist/downshift.umd.min.js": {
-    "bundled": 126810,
-    "minified": 41328,
-    "gzipped": 11504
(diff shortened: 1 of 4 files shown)


silviuavram:
I already addressed this in https://github.com/downshift-js/downshift/pull/907 I think. Once I launch an alpha with v5 can you check if it fixes it for you?

jorgemoya:
@silviuavram I looked yesterday and looks like it does. I will test it again when it's released. However, what's the timeline for v5? Would it make sense to release a patch for now?

silviuavram:
The plan is to have it next week.

jorgemoya:
Closing since fix is potentially included in #907

jorgemoya:
@silviuavram Do you think we can take an extra look at this PR to fix this issue without waiting for the braking changes?

silviuavram:
@jorgemoya I am considering doing that since I want to break that big PR into smaller ones.

silviuavram:
Can you re-open it please? And have it ready for merge?

PR #907, opened 2020-01-22
downshift v5

silviuavram:
Contains refactoring, test improvements, bug fixes.

BREAKING CHANGE: Removed `FunctionClearKeysSoFar` from state change types and TS typings. In useSelect, once the timeout for keeping the character keys in memory has expired, `FunctionSetInputValue` will be used. We are doing this as we are renaming `keysSoFar` with `inputValue`.

To migrate to the new change, simply check for `FunctionSetInputValue` with empty string as `inputValue` instead of checking for `FunctionClearKeysSoFar` in `stateReducer`.

BREAKING CHANGE: Both `getA11yStatusMessage` and `getA11ySelectionMessage` will be called with the same props as the `getA11yStatusMessage` in `<Downshift>`, apart from `previousResultCount`. In the TS typings it's now marked as optional, and all functions have the same interface definition, `A11yStatusMessageOptions`.

To migrate to the new changes, in `useSelect` and `useCombobox`, if you used `items` as parameters in any of the a11y message functions, now you should use `resultCount` as probably you only needed `items.length` from it anyway.

Also typings `UseSelectA11yMessageOptions` and `UseComboboxA11yMessageOptions` have been removed. Use `A11yStatusMessageOptions` instead.

Code Changes: tests have been enhanced by using `user input` from `RTL` and they now look better and cleaner. Also covered more use cases better. Code has been refactored as well, and bundle size slightly reduced.

Functional Improvement: better focus management for both `useSelect` and `useCombobox`.

Fixes https://github.com/downshift-js/downshift/issues/832.
Closes https://github.com/downshift-js/downshift/issues/892
Closes https://github.com/downshift-js/downshift/issues/891
Closes https://github.com/downshift-js/downshift/issues/873


silviuavram:
Unfortunately, keeping focus on the `toggleButton` in the case of `useSelect` will prevent screen readers from entering Forms Mode when navigating in the list. It's an accessibility concern we cannot have so we should leave the focus on the menu when the dropdown is open.

That being said, most changes in this PR are still relevant. To move forward, we will add the aria attributes and handlers back to the list, move the tests back, add the Menu stateChangeTypes back etc.

PR #915, opened 2020-01-28
✏️ Fix wrong github page link

cloud-walker:
It is pointing to `useSelect` markdown file


silviuavram:
Oups! Thanks for this finding!

PR #925, opened 2020-02-05
Do not call onStateChange if the state did not change

### LINKED ISSUE #924: Infinite loop when setting state in `onStateChange`
- `downshift` version: 4.0.8
- `node` version: 10.16.3
- `npm` (or `yarn`) version: yarn v1.21.1

**Relevant code or config**

```javascript
```

**What you did**:

I passed `onStateChange` into `useSelect`.

```
import React from "react";
import { render } from "react-dom";
import { useSelect } from "downshift";
import { items, menuStyles } from "./utils";

function DropdownSelect() {
  const [counter, setCounter] = React.useState(0);

  const {
    isOpen,
    selectedItem,
    getToggleButtonProps,
    getLabelProps,
    getMenuProps,
    highlightedIndex,
    getItemProps
  } = useSelect({
    items,
    onStateChange: () => {
      setCounter(prev => prev + 1);
    }
  });
  return (
    <div>
      {counter}

      <label {...getLabelProps()}>Choose an element:</label>
      <button {...getToggleButtonProps()}>{selectedItem || "Elements"}</button>
      <ul {...getMenuProps()} style={menuStyles}>
        {isOpen &&
          items.map((item, index) => (
            <li
              style={
                highlightedIndex === index ? { backgroundColor: "#bde4ff" } : {}
              }
              key={`${item}${index}`}
              {...getItemProps({ item, index })}
            >
              {item}
            </li>
          ))}
      </ul>
      {/* if you Tab from menu, focus goes on button, and it shouldn't. only happens here. */}
      <div tabIndex="0" />
    </div>
  );
}

render(<DropdownSelect />, document.getElementById("root"));
```

**What happened**:

onStateChange gets called in an infinite loop

**Reproduction repository**:
https://codesandbox.io/s/useselect-usage-slt20

**Problem description**:
Passing in `onStateChange` to `useSelect` resulst in an infinite render loop. Related to the changes in https://github.com/downshift-js/downshift/pull/912. The issue is that the `props` in the useEffect's dependency array is never going to be referentially equal to the previous version. So onStateChange will be called every time a render occurs. And if `onStateChange` triggers another render (which it would for a lot of typical use cases), we end up in a loop.

**Suggested solution**:
I believe the check
```
  if (props.onStateChange && changes !== undefined && changes !== state) {
    props.onStateChange(changes);
  }
```
in `callOnChangeProps` would prevent the infinite loop because we do not call `onStateChange` if the state didn't actually change.

TLadd:
Fix #924

**What**: Fixes #924

**Why**: Creates an infinite loop without this change

**How**: Tested this resolved my issue.

**Checklist**:

- [X] Documentation
- [X] Tests
- [X] TypeScript Types
- [X] Flow Types
- [X] Ready to be merged

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,38 +1,38 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 115938,
-    "minified": 53222,
-    "gzipped": 11708
+    "bundled": 116463,
+    "minified": 53336,
+    "gzipped": 11738
   },
   "preact/dist/downshift.cjs.js": {
-    "bundled": 114661,
-    "minified": 52189,
-    "gzipped": 11606
+    "bundled": 115186,
+    "minified": 52303,
+    "gzipped": 11634
   },
   "preact/dist/downshift.umd.js": {
-    "bundled": 143824,
-    "minified": 51110,
-    "gzipped": 13253
+    "bundled": 144433,
+    "minified": 51224,
+    "gzipped": 13276
   },
   "preact/dist/downshift.umd.min.js": {
-    "bundled": 127462,
-    "minified": 41220,
-    "gzipped": 11303
+    "bundled": 128071,
+    "minified": 41334,
+    "gzipped": 11319
   },
   "dist/downshift.umd.min.js": {
-    "bundled": 131692,
-    "minified": 42541,
-    "gzipped": 11806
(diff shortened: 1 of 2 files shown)


silviuavram:
Thank you for these changes! Let me trigger the CI again and if all is green we will merge.

PR #926, opened 2020-02-07
fix: only add aria-activedescendant when menu is open

### LINKED ISSUE #921: [useSelect] aria-activedescendant does not have valid value with controlled selectedItem
- `downshift` version: 4.0.8
- `node` version: 12.14.1
- `npm` (or `yarn`) version: yarn 1.21.1

**Relevant code or config**

```javascript
import React from 'react'
import { render } from 'react-dom'
import { useSelect } from 'downshift'

const items = ['Apples', 'Bananas', 'Oranges']

const Select = ({ items, selectedItem }) => {
  const {
    getItemProps,
    getLabelProps,
    getMenuProps,
    getToggleButtonProps,
    isOpen,
  } = useSelect({ items, selectedItem })

  return (
    <div>
      <label {...getLabelProps()}>Fruit: </label>
      <button {...getToggleButtonProps()}>{selectedItem}</button>
      {/* <ul {...getMenuProps(isOpen ? undefined : { "aria-activedescendant": null })}> */}
      <ul {...getMenuProps()}>
        {isOpen &&
          items.map((item, index) => (
            <li key={item} {...getItemProps({ item, index })}>
              {item}
            </li>
          ))}
      </ul>
    </div>
  )
}

const App = () => <Select items={items} selectedItem={items[0]} />

render(<App />, document.getElementById('root'))
```

**What you did**:
use `useSelect` with controlled props (`items` and `selectedItem`)

**What happened**:
when used with controlled props, `useSelect`'s `getMenuProps` ouputs an `aria-activedescendant` attribute pointing to a non-existing id (because the `<li>` elements are only rendered when the select is open.
Lighthouse therefore complains with "[aria-*] attributes do not have valid values".

**Reproduction repository**:

**Problem description**:

**Suggested solution**:
I'm currently working aroung this with `{...getMenuProps(isOpen ? undefined : { "aria-activedescendant": null })}` (see above), but i guess there is a better solution?

stefanprobst:
**What**: in `useSelect`/`useCombobox` only add `aria-activedescendant` attribute when menu is open to avoid pointing to non-existing id.

**Why**: Fixes #921

**How**: check if `isOpen` is `true` before adding `aria-activedescendant`.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged

### DIFF
--- a/src/hooks/useCombobox/__tests__/getToggleButtonProps.test.js
+++ b/src/hooks/useCombobox/__tests__/getToggleButtonProps.test.js
@@ -208,6 +208,8 @@ describe('getToggleButtonProps', () => {
         )
 
         fireEvent.click(toggleButton)
+        // fireEvent.blur(input)
+        input.blur() // the code above does not blur.
         fireEvent.click(toggleButton)
 
         expect(input.getAttribute('aria-activedescendant')).toBe(
--- a/src/hooks/useCombobox/__tests__/getToggleButtonProps.test.js
+++ b/src/hooks/useCombobox/__tests__/getToggleButtonProps.test.js
@@ -208,8 +208,7 @@ describe('getToggleButtonProps', () => {
         )
 
         fireEvent.click(toggleButton)
-        // fireEvent.blur(input)
-        input.blur() // the code above does not blur.
+        input.blur()
         fireEvent.click(toggleButton)
 
         expect(input.getAttribute('aria-activedescendant')).toBe(
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,38 +1,38 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 115938,
-    "minified": 53222,
-    "gzipped": 11708
+    "bundled": 116483,
+    "minified": 53342,
+    "gzipped": 11738
   },
   "preact/dist/downshift.cjs.js": {
-    "bundled": 114661,
-    "minified": 52189,
-    "gzipped": 11606
+    "bundled": 115206,
+    "minified": 52309,
+    "gzipped": 11633
   },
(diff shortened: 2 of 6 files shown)


src/hooks/useCombobox/__tests__/getToggleButtonProps.test.js
```diff
@@ -208,6 +208,8 @@ describe('getToggleButtonProps', () => {
         )
 
         fireEvent.click(toggleButton)
+        // fireEvent.blur(input)
```
silviuavram: Can we remove this?

stefanprobst: sure -- comment was just for symmetry with the comments above

silviuavram: I mean also the line below, I don't see why is it needed.

stefanprobst: the `input.blur()` is needed actually (or `toggleButton.focus()` alternatively). i had assumed it was just an oversight this test did not have that line while e.g. the one above did.
without manually blurring/focusing the actions fire in wrong order:

without manual blur:
* `__togglebutton_click__`
* `__togglebutton_click__`
* `__togglebutton_click__`
* `__input_blur__`

with:
* `__togglebutton_click__`
* `__togglebutton_click__`
* `__input_blur__`
* `__togglebutton_click__`

silviuavram:
As I merged a big change as part of v5 I re-created this PR. Ran the tests without the `.blur` in the discussion and they were passing. Please take a look at https://github.com/downshift-js/downshift/pull/945

PR #933, opened 2020-02-09
fix(useSelect): use mouse and touch events to trigger blur

jorgemoya:
Couldn't reopen #906 so I'm opening a new PR.

**What**:

Much like in #893, I need to listen to mouse and touch events in the menu.

**Why**:

If any of the items contains an element that is focusable, the menu will immediately blur. I want to keep the menu open and allow `onClick` event handlers for said elements to be triggered. I particularly had an issue with adding `a` elements for some of my items; clicking on them would close the menu and not open the link, as expected. I also want to be able to right click on them.

**How**:

I'm reusing the logic that was added in #893, now listening for mouse and touch events as well as checking if the target is within Downshift.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged
      

Added some of the missing tests that were included in #893 for useCombobox.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,52 +1,38 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 115777,
-    "minified": 53152,
-    "gzipped": 11697
+    "bundled": 117677,
+    "minified": 53993,
+    "gzipped": 11745
   },
   "preact/dist/downshift.cjs.js": {
-    "bundled": 114498,
-    "minified": 52117,
-    "gzipped": 11593
+    "bundled": 116400,
+    "minified": 52960,
+    "gzipped": 11637
   },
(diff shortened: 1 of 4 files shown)


jorgemoya:
Will resolve conflicts soon.

silviuavram:
Don't. Update to v5 and see if it solves the problem. Thanks!

jorgemoya:
@silviuavram It works, thanks!

silviuavram:
@jorgemoya do you think we can abstract that logic into a separate hook / function? We are duplicating in both useSelect and useCombobox and I was thinking if we can do better.

jorgemoya:
@silviuavram Sounds like a good idea! I will give it a try but might need to wait for next week.

PR #945, opened 2020-02-17
fix(hooks): only add aria-activedescendant when menu is open

silviuavram:
Based on https://github.com/downshift-js/downshift/pull/926, means to fix the issue where in hooks we can have `aria-activedescendant` with a item value even though the menu is closed and the item does not exist.

Fixes https://github.com/downshift-js/downshift/issues/921.


silviuavram:
@all-contributors please add @stefanprobst for code, tests

PR #952, opened 2020-02-23
fix: make useEnhancedReducer a pure function

silviuaavram:
Fixes https://github.com/downshift-js/downshift/issues/874.


src/hooks/utils.js
```diff
@@ -73,24 +78,29 @@ function callOnChangeProps(props, state, changes) {
...
   )
-
   const [state, dispatch] = useReducer(enhancedReducer, initialState)
   const dispatchWithProps = action => dispatch({props, ...action})
 
+  useEffect(() => {
+    if (prevState.current) {
+      callOnChangeProps(props, prevState.current, state)
+    }
+  })
```
mufasa71: useEffect will be called on every re-render, possible with performance penalty

PR #954, opened 2020-02-23
Migrate to import * as React from 'react'

TrySound:
**What**:

Migrated to `import * as React from 'react'`

**Why**:

Ref https://github.com/facebook/react/pull/18102

This is the recommended way to import react now.

**How**:

With the help of replace-in-files-cli package

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,6 @@ the library treeshaked (pruned) and given only the code you need. Since version
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
-
 - [Installation](#installation)
 - [Usage](#usage)
 - [Basic Props](#basic-props)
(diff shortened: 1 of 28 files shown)


silviuaavram:
Shouldn't we get warning from React? I don't see this being enforced. And could not tell from that issue if it's something that brings any value. Can you describe this a bit so we can have more context? Thank you!

TrySound:
I don't think it's easy to add warning before migrating to esm.

There is a value with bundlers like rollup when target format is esm.

When bundle apps you get optimisable output with named exports
```
import { e as createElement } from 'react';
createElement('div')
```
Default exports leads to hard member expression which can be optimised by babel or webpack plugins but not in natural way
```
import React from 'react';
React.createElement('div')
```

oliviertassinari:
@silviuaavram They might be a bit more context in #19802.

silviuaavram:
Understood :) will merge this, thank you!

PR #967, opened 2020-03-06
fix(hooks): call onStateChange with type and only changed props

silviuavram:
**What**:
Fixes https://github.com/downshift-js/downshift/issues/963.
Should call onStateChange with a type now, and also only with props that changed.

**Why**:
Previous implementation was not following the API.

**How**:
Filtered the changes that we call onStateChange with.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged

### DIFF
--- a/src/hooks/utils.js
+++ b/src/hooks/utils.js
@@ -1,6 +1,11 @@
 import PropTypes from 'prop-types'
 import {useCallback, useReducer} from 'react'
-import {scrollIntoView, getNextWrappingIndex, getState, generateId} from '../utils'
+import {
+  scrollIntoView,
+  getNextWrappingIndex,
+  getState,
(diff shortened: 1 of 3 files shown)


src/hooks/utils.js
```diff
@@ -81,7 +93,7 @@ function useEnhancedReducer(reducer, initialState, props) {
       const changes = reducer(state, action)
       const newState = stateReduceLocal(state, {...action, changes})
 
-      callOnChangeProps(action.props, state, newState)
+      callOnChangeProps(action, state, newState)
```
TrySound, line 96: Reducer is probably not the right place for side effects

PR #971, opened 2020-03-10
fix(hooks): replace Object.entries with Object.keys to support > IE 9

louisaspicer:
Object.entries is not required because the value in the object is not being used. Replaced with Object.keys as at least it is supported on IE 9 and later.

See https://github.com/downshift-js/downshift/issues/970

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] TypeScript Types N/A
- [ ] Flow Types N/A
- [x] Ready to be merged


silviuavram:
Changes look great! Thanks for your time to handle this!

@all-contributors please add @louisaspicer for code, bug.

PR #983, opened 2020-03-31
fix(typescript): add missing `scrollIntoView` definitions to hooks

moriko:
**What**:

Added missing `scrollIntoView` type definitions for `useSelect` and `useCombobox` hooks.

**Why**:

Was getting Typescript error while trying to pass `scrollIntoView` prop to `useCombobox`:
```
Object literal may only specify known properties, and 'scrollIntoView' does not exist in type 'UseComboboxProps<string>'
```

**How**:

A small edit to the `typings/index.d.ts` file

**Checklist**:

- [x] TypeScript Types
- [ ] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -287,6 +287,7 @@ export interface UseSelectProps<Item> {
   menuId?: string
   toggleButtonId?: string
   getItemId?: (index: number) => string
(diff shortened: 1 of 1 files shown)


silviuaavram:
Wow awesome! Thank you for this!

Will change it from `chore` to `fix` since we need it released in order for people to pick up changes.

PR #985, opened 2020-04-04
fix(hooks): add type to on change handler params

silviuaavram:
**What**:
As discussed [here](https://github.com/downshift-js/downshift/issues/879#issuecomment-608349780) we are adding `type` to other handlers as we claim to support this already in the docs.

**Why**:

We already claim to support these. We already pass `type` in `onStateChange`.
Fixes https://github.com/downshift-js/downshift/issues/1015.

**How**:

- Add `type` in other chance handlers as well (`onIsOpenChange`, `onSelectedItemChange` etc.)
- Fix TS typings for these handlers to reflect they can have partial state + type.
- Remove `props` from the `stateReducer` call param `actionAndChanges`. Now it should be called with `{...action, changes}`, where `action` has a `type` and other data passed by the hook in dispatch.
- Also fix typings for `stateReducer` in both hooks to reflect it's being called with `{...action, changes}`.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -299,19 +299,19 @@ export interface UseSelectProps<Item> {
   environment?: Environment
 }
 
-export interface UseSelectStateChangeOptions<Item> {
+export interface UseSelectStateChangeOptions<Item>
+  extends UseSelectDispatchAction {
   changes: Partial<UseSelectState<Item>>
-  action: UseSelectDispatchAction
-  props: UseSelectProps<Item>
-}
-export interface UseSelectStateChange<Item>
-  extends Partial<UseSelectState<Item>> {
-  type: UseSelectStateChangeTypes
 }
 
 export interface UseSelectDispatchAction {
   type: UseSelectStateChangeTypes
   [data: string]: any
+}
+
+export interface UseSelectStateChange<Item>
+  extends Partial<UseSelectState<Item>> {
+  type: UseSelectStateChangeTypes
 }
 
 export interface UseSelectGetMenuPropsOptions
@@ -448,20 +448,19 @@ export interface UseComboboxProps<Item> {
   environment?: Environment
 }
 
-export interface UseComboboxStateChangeOptions<Item> {
+export interface UseComboboxStateChangeOptions<Item>
+  extends UseComboboxDispatchAction {
   changes: Partial<UseComboboxState<Item>>
-  action: UseComboboxDispatchAction
-  props: UseComboboxProps<Item>
-}
-
-export interface UseComboboxStateChange<Item>
-  extends Partial<UseComboboxState<Item>> {
-  type: UseComboboxStateChangeTypes
 }
 
 export interface UseComboboxDispatchAction {
   type: UseSelectStateChangeTypes
   [data: string]: any
+}
+
+export interface UseComboboxStateChange<Item>
+  extends Partial<UseComboboxState<Item>> {
+  type: UseComboboxStateChangeTypes
 }
 
 export interface UseComboboxGetMenuPropsOptions
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -292,17 +292,26 @@ export interface UseSelectProps<Item> {
     state: UseSelectState<Item>,
     actionAndChanges: UseSelectStateChangeOptions<Item>,
   ) => UseSelectState<Item>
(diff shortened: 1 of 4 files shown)


kimroen:
I actually just went in to fix this myself and saw this PR. What's holding this back?

I implemented the fix in a very similar way, but as far as I could tell the `type` property is already passed in to the `onChange` handlers. Am I mistaken?

How is this a breaking change, by the way? I wouldn't think adding a property would break anything.

silviuaavram:
Hi @kimroen! It's breaking as a result of the TypeScript changes. It's already passed to `onStateChange` yes, but not to the others, like `onHighlightedIndexChange`.

I am waiting to converge a few more PRs with breaking changes before releasing v6. But I don't think that this PR is blocked by anything, as far as I see it it should be ready to be merged. You can review it also and tell me what you think, on both the functionality change and the Typescript changes. Thank you!

kimroen:
> It's breaking as a result of the TypeScript changes. It's already passed to `onStateChange` yes, but not to the others, like `onHighlightedIndexChange`.

Thanks for the explanation! I don't think I agree that adding a field to an object passed to a function is a breaking change, but I'm not extremely invested and I can go along with it ✨ 

Regarding the changes, I made pretty much the same change in my own fix, here:

https://github.com/kimroen/downshift/commit/c7b10628ad9931bb69702e89d4d2874dcf8542ea

In addition to that one though, I actually made another change which I would love your thoughts on:

https://github.com/kimroen/downshift/commit/c7d2264c2514771cb21a070475f16a7b44bef1ff

This makes it so the field that is relevant for the specific change callback function it is passed to is not optional.

In other words, in the change passed to `onInputValueChange`, the `inputValue` field is non-optional.

In practice this looks like this:

Skjermbilde 2020-06-07 kl  00 07 29

Skjermbilde 2020-06-07 kl  00 07 07

(Note which fields are optional in the two)

Is this something we could be interested in including? If so, I could make a PR for it, but it would conflict pretty heavily with this one, so I wanted to ask 😄

ALITHEALI:
Any update on this?

silviuaavram:
It's not just adding a field, it's replacing as well, for instance:
`onInputValueChange?: (changes: Partial<UseComboboxState<Item>>) => void`
with
`onSelectedItemChange?: (changes: Partial<UseComboboxState<Item>>) => void`
which is different (and correct).

As for status, please review this @ALITHEALI @kimroen and make sure it's ready to merge. It will be included in the next major release. I want to merge a few more breaking change PRs into it so that is why I am not merging it right now.

typings/index.d.ts
```diff
@@ -292,17 +292,26 @@ export interface UseSelectProps<Item> {
...
 
-export interface UseSelectStateChangeOptions<Item> {
+export interface UseSelectStateChangeOptions<Item>
+  extends UseSelectDispatchAction {
+  changes: Partial<UseSelectState<Item>>
+}
+
+export interface UseSelectDispatchAction {
+  type: UseSelectStateChangeTypes
+  [data: string]: any
```
kimroen, line 309: This key that has a key of `string` and a value of `any` is going to make autocomplete a little less useful - do we not have anything more complete we could put here instead, or is it entirely user-defined?

silviuaavram: Oups, edited with the acutal line of `onInputValueChange`.

That line with [data:string]: any is because the dispatch may have stuff in it required to change the state. Do you have a better suggestion?

kimroen: If we don't know what that "stuff" is ahead of time then it might be okay, but if we do then we should try to be more specific in the type. If we don't know, then it would be better (for typescript) if all the unknown fields are in a separate object so we can limit the unknowns to one place without polluting the main change object.

silviuaavram: I think we can determine what the fields are by going through all the dispatch calls in the hooks and add those fields as optional. Would that be ok?

silviuaavram:
@kimroen thank you for suggesting to update the types for `action`. How do they look now? :D

kimroen:
> @kimroen thank you for suggesting to update the types for `action`. How do they look now? :D

This is so much better, thank you very much! ✨

Issue #975, opened 2020-03-17
useCombobox: crash when using itemToString, and selectedItem is reset to null


rlue:
- `downshift` version: 5.0.5
- `node` version: 10.17.0
- `yarn` version: berry (2.0.0-rc.30)

### Relevant code

See reproduction repro below, or [view the JS file in question](https://github.com/rlue/downshift-debug/blob/master/app/index.js).

### What you did

Video demo [here](https://youtu.be/_jMg67z4a-U)

1. Use non-String values for item list;

   ```javascript
   // not this
   ['Alabama', 'Alaska', 'Arizona', 'Arkansas', 'California', 'Colorado', 'Connecticut']

   // but this
   [
     { name: 'Alabama' },
     { name: 'Alaska' },
     { name: 'Arizona' },
     { name: 'Arkansas' },
     { name: 'California' },
     { name: 'Colorado' },
     { name: 'Connecticut' }
   ]
   ```

2. which requires use of the `itemToString` prop;

   ```javascript
   itemToString: ({ name }) => name,
   ```

3. then, use the combobox in the browser;

4. enter a value, make a selection, repeat this as much as you like without any issue;

5. enter a value, highlight an item, **hit escape**—

### What happened

Kaboom, React crashes (with message `TypeError: _ref is null`).

### Reproduction repository

<https://github.com/rlue/downshift-debug>

### Problem description

I don't know enough about downshift internals to say for sure what's wrong here. I wonder if perhaps it's related to #719?

In any case, it appears that the following combination of conditions causes React to crash:

1. using a downshift combobox,
2. using items which are not plain strings,
3. thus the using `itemToString` prop,
4. after the `selectedItem` state has already been set on the combobox,
5. when resetting `selectedItem` to `null` (by highlighting a search result and pressing `<Esc>`).

### Suggested solution

No idea! Happy to take guidance on this point, though.

silviuavram:
well you did point out that selectedItem beocomes null, and you cannot destructure null into anything. so try `itemToString = item => item ? item.name || ''`

Oh and not sure if it's intentional, but in your use case try to use all combobox components (input, menu, items + combobox and maybe toggleButton and label) just to make the experience 100% accessible.

Let me know how it goes.

rlue:
> well you did point out that selectedItem beocomes null, and you cannot destructure null into anything. so try `itemToString = item => item ? item.name || ''`

I am embarrassed to say that this was absolutely it. Spent my whole day scratching my head over this.

[...]

PR #990, opened 2020-04-09
docs: Add null-check note for itemToString (#975)

rlue:
This commit includes one small, overall style change, as well: [most U.S. style guides recommend that "e.g." be followed by a comma][0], whereas the downshift docs leave it bare.

The core changes in this commit use the phrase "e.g." in a few places, so in the interest of consistency, this commit applies the comma both in new additions and existing instances of "e.g." throughout the docs.

[0]: https://www.dailywritingtips.com/comma-after-i-e-and-e-g/

### Notes

1. I did not run pre-commit hooks because `kcd-scripts test --findRelatedTests` just hung forever. Let me know if I need to make this work.

2. I removed the following point in the `useCombobox` docs:

   > It is also used to allow highlighting by typing character keys, when downshift looks for the items whose string version start with the keys typed.

   I might be wrong, but I believe this only applies to `useSelect`. (Any time I typed extra characters on the combobox, they went directly into the input field.)

3. For consistency, I also updated the "defaults to" code sample to reflect [the actual default callback definition](https://github.com/downshift-js/downshift/blob/master/src/hooks/utils.js#L39) (in `src/hooks/utils.js`).

4. From a package-usability perspective, I would have preferred to put the null checks directly into downshift in the first place; e.g.,

   ```javascript
   // src/hooks/useCombobox/utils.js
   - inputValue = props.itemToString(selectedItem)
   + inputValue = selectedItem ? props.itemToString(selectedItem) : ''
   ```

   but I get the impression that downshift errs on the side of letting users make their own decisions about things. Would you have preferred that, too, or is the current PR better?

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,10 +269,13 @@ the section "[Children Function](#children-function)".
 
 ### itemToString
 
-> `function(item: any)` | defaults to: `i => (i == null ? '' : String(i))`
+> `function(item: any)` | defaults to: `item => (item ? String(item) : '')`
 
-Used to determine the string value for the selected item (which is used to
-compute the `inputValue`).
+If your items are stored as, say, objects instead of strings, downshift still
+needs a string representation for each one (e.g., to set `inputValue`).
+
+**Note:** This callback _must_ include a null check: it is invoked with `null`
+whenever the user abandons input via `<Esc>`.
 
 ### onChange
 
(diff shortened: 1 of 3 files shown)


silviuaavram:
Thank you for these improvements! Much appreciated!

silviuaavram:
@all-contributors please add @rlue for docs

PR #993, opened 2020-04-11
fix(input): handle IME events on input Enter keydown

silviuaavram:
**What**:

Recreating the fix from https://github.com/downshift-js/downshift/pull/671.

**Why**:

Handle this case by default in Downshift without users having to create boiler plate handling.

Handle IME composing events.

**How**:

Check for `event.which === 229`. If true, then will wait for a subsequent event without composing.

More info on:
https://github.com/mui-org/material-ui/issues/19435
https://github.com/zendeskgarden/react-components/issues/598

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/downshift.get-input-props.js
+++ b/src/__tests__/downshift.get-input-props.js
@@ -385,6 +385,32 @@ test('enter on an input with an open menu does nothing without a highlightedInde
   expect(childrenSpy).not.toHaveBeenCalled()
 })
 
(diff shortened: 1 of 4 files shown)


silviuaavram:
@all-contributors please add neet for bug, code

Issue #998, opened 2020-04-14
Typings in useMultipleSelection: getSelectedItemProps (item vs selectedItem)


mattleonowicz:
- `downshift` version: 5.1.0
- `node` version: irrelevant
- `npm` (or `yarn`) version: irrelevant

Docs, examples and typings are inconsistent. Sometimes getSelectedItemProps accepts object with `item` and sometimes `selectedItem`. Looking into source I think it should be `selectedItem`, but I'm not sure

silviuaavram:
Yes, it should be `selectedItem` to avoid confusion with the `item` from the other hooks/component. Where do we have inconsistency? Can you create a PR to patch things up?

silviuaavram:
Closing, thank you for the fix!

silviuaavram:
@all-contributors please add @mattleonowicz for code

PR #1002, opened 2020-04-15
fix(useMultipleSelection): replace item with selectedItem in typings for `getSelectedItemProps`

mattleonowicz:
See #998

- [x] Documentation
- [ ] Tests
- [x] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -586,8 +586,10 @@ export interface A11yRemovalMessage<Item> {
 }
 
 export interface UseMultipleSelectionGetSelectedItemPropsOptions<Item>
-  extends GetItemPropsOptions<Item>,
-    GetPropsWithRefKey {}
+  extends React.HTMLProps<HTMLElement>, GetPropsWithRefKey {
+  index?: number
+  selectedItem: Item
+}
 
 export interface UseMultipleSelectionComboboxGetDropdownProps
   extends GetInputPropsOptions,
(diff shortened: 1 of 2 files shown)


typings/index.d.ts
```diff
@@ -586,8 +586,10 @@ export interface A11yRemovalMessage<Item> {
 }
 
 export interface UseMultipleSelectionGetSelectedItemPropsOptions<Item>
-  extends GetItemPropsOptions<Item>,
-    GetPropsWithRefKey {}
+  extends React.HTMLProps<HTMLElement>, GetPropsWithRefKey {
```
silviuaavram, line 589: can we have a separate one for this, similar to the one below but only with `index` and `selectedItem` ?
```ts
export interface GetItemPropsOptions<Item>
  extends React.HTMLProps<HTMLElement> {
  index?: number
  item: Item
  isSelected?: boolean
  disabled?: boolean
}
```

mattleonowicz: I'm not sure if I understand you. That's what I did. `UseMultipleSelectionGetSelectedItemPropsOptions` is no longer extending the wrong `GetItemPropsOptions` and instead has `index` and `selectedItem` props of its own

silviuaavram: Actually it doesn't matter since it will only be used only here.

I was suggesting to have a separate GetSelectedItemPropsOptions with index and selectedItem and make UseMultipleSelectionGetSelectedItemPropsOptions extend that, instead of having index and selectedItem directly in UseMultipleSelectionGetSelectedItemPropsOptions .

PR #1020, opened 2020-04-26
fix(useControlledState): use reducer + useEffect

silviuaavram:
**What**:
Reverting the `useState` usage for computing the controlled state and going back to `useReducer` + adding `useEffect` to call onChange handlers.

**Why**:
Fix https://github.com/downshift-js/downshift/issues/1018 and keep the fixes for React warning and inputValue not being updated. https://github.com/downshift-js/downshift/issues/962 https://github.com/downshift-js/downshift/issues/982

**How**:
Going back to previous implementation of the state handlers with useReducer but now calling on change handlers in a useEffect.
Also, for useCombobox, we will check if selectedItem is controlled and has been changed (via props or state) and based on that trigger a state change with the new inputValue.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged
      

Closes https://github.com/downshift-js/downshift/issues/1018

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,38 +1,38 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 135913,
-    "minified": 62673,
-    "gzipped": 13187
+    "bundled": 137464,
+    "minified": 63123,
+    "gzipped": 13314
   },
   "preact/dist/downshift.cjs.js": {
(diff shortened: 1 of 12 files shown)


silviuaavram:
This needs more tests.

PR #1042, opened 2020-05-09
feat: add `selectedItemChanged` prop to useCombobox

montezume:
**What**:

While working with `useCombobox`, I found a re-rendering bug I was not able to solve. While looking at Downshift, and the code for `useCombobox`, I realized that while downshift supports the prop `selectedItemChanged`, useCombobox does not.

In my project, I had to migrate my code from using useCombobox to using vanilla downshift. 

**Why**:

If you are dealing with objects as values, `useCombobox` will fire act as if the selectedItem changed, when it in fact did not.

**How**:

Added the same `selectedItemChanged` prop that downshift supports.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged

### DIFF
--- a/src/hooks/useCombobox/index.js
+++ b/src/hooks/useCombobox/index.js
@@ -35,6 +35,7 @@ function useCombobox(userProps = {}) {
     ...defaultProps,
     ...userProps,
   }
+
   const {
     initialIsOpen,
     defaultIsOpen,
--- a/src/hooks/useCombobox/index.js
+++ b/src/hooks/useCombobox/index.js
@@ -35,7 +35,6 @@ function useCombobox(userProps = {}) {
     ...defaultProps,
     ...userProps,
   }
-
   const {
     initialIsOpen,
     defaultIsOpen,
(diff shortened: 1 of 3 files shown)


silviuaavram:
Can you create a sandbox with your use case that is not working with `useCombobox`?

montezume:
Here, you'll notice when you select an item, code sandbox crashes

https://codesandbox.io/s/usecombobox-usage-tqtgz?file=/src/index.js

silviuaavram:
I don't know why items are re-created in the context of using a combobox, so for me this example is perhaps a forced one.

For the sake of backwards compatibility we can add the prop. I would also like unit tests for it and also documentation entry, should be similar to the one in Downshift.

montezume:
Our use-case is a quite complex Combobox that populates its options based on API calls to get user permissions and other user data. So the options can be recreated on re-renders, especially if some other data changes.

All I can say is that I spent about a day figuring out that the lack of `selectedItemChanged` was breaking our app, and it caused me some confusion.

sbousamra:
Any news on this? I need this change as well pretty badly, i'd like to forcefully tell downshift that an item has been selected even if that value has not changed.

montezume:
I migrated to use regular Downshift instead of the hooks @sbousamra

silviuaavram:
If @montezume will not continue on this, then we can pick where he left off, add unit tests, add Typescript and Documentation. Anyone interested in finishing this work? Thanks!

PR #1045, opened 2020-05-10
feat(hooks): allows virtualization for hooks

silviuaavram:
**What**:
Support virtualization in useCombobox and useSelect.

**Why**:
To allow users to use react-virtual, react-virtualized and react-window with minimum effort.

**How**:
We keep itemRefs in an Object by the ID as key, instead of pushing them in an array.
We also fixed a bug with `highlightedIndex` being returned as 0 on ArrowDown when there are no items available.

**Checklist**:

- [x] Tests
- [x] useSelect changes
- [x] Ready to be merged
      

Fixes https://github.com/downshift-js/downshift/issues/1041.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,38 +1,38 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 137091,
-    "minified": 62770,
-    "gzipped": 13314
+    "bundled": 137552,
+    "minified": 62918,
+    "gzipped": 13275
   },
   "preact/dist/downshift.cjs.js": {
-    "bundled": 135827,
-    "minified": 61750,
-    "gzipped": 13218
+    "bundled": 136289,
+    "minified": 61899,
+    "gzipped": 13179
   },
   "preact/dist/downshift.umd.min.js": {
-    "bundled": 147173,
-    "minified": 46825,
-    "gzipped": 12595
+    "bundled": 147655,
+    "minified": 46960,
+    "gzipped": 12608
   },
   "dist/downshift.umd.min.js": {
-    "bundled": 151367,
-    "minified": 48130,
-    "gzipped": 13149
+    "bundled": 151848,
+    "minified": 48265,
+    "gzipped": 13163
   },
   "preact/dist/downshift.umd.js": {
-    "bundled": 157327,
-    "minified": 55181,
-    "gzipped": 13926
+    "bundled": 157809,
+    "minified": 55316,
+    "gzipped": 13940
   },
   "dist/downshift.umd.js": {
-    "bundled": 186911,
-    "minified": 64106,
-    "gzipped": 16561
+    "bundled": 187392,
+    "minified": 64241,
+    "gzipped": 16575
   },
   "dist/downshift.esm.js": {
-    "bundled": 136473,
-    "minified": 62229,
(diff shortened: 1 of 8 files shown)


silviuaavram:
Damn, it seems to further fix this, we need to get the element from`this.props.environment.document` and do not rely on item refs anymore.

silviuaavram:
Did some more digging, actually the itemRefs approach suggested by Kent should work, so reverted the whole getItemById commit.

It may depend on how https://github.com/tannerlinsley/react-virtual/issues/22 is solved. Issue is: when we ArrowUp from first item, and going to last item, useCombobox says we have highlightedIndex as 25 (last item, which is correct), but react-virtual only gives us the last items all the way to 24. So we have 💣when using itemRefs (getItemProps was just not called for the last item) and no scroll when using getElementById (no element in the DOM but no kaboom).

kentcdodds:
FWIW, I think using the `itemRefs` approach is better than querying the document so good call there 👍

kentcdodds:
Thank you!

kentcdodds:
Hi @silviuaavram! I just tried this and it doesn't seem to work if I am on highlightedIndex of `0` and then hit `ArrowUp`: https://codesandbox.io/s/wizardly-cloud-0vfyl

image

Did you run into this in your testing?

silviuaavram:
@kentcdodds  I was testing with the react-virtual scroll in the state reducer.🤦‍♂️

It makes sense what you did by moving it in the state change handler, the reducers should be pure.

Lifecycle may not help us very much here. ArrowUp, state change, react-virtual scroll, downshift scroll in effect. We need another re-render to have those elements in the DOM in order for downshift scroll to work. It worked with the react-virtual scroll in the reducer because we had the new items available on the first re-render, and the downshift scroll in the effect could work. Now it's a bust, since we get the items on the second re-render, as a consequence of moving the react-virtual scroll in the onStateChange.

The stupid workaround can be to just pass a noop as scrollIIntoView and let react-virtual do the scrolling.

I cannot think of any way to fix this right now, but maybe you can suggest something since you know react lifecycle better than me. I will keep looking into this for a bit and see if I can find something.

kentcdodds:
Here's a working solution:

https://codesandbox.io/s/frosty-engelbart-wlrqk?file=/src/index.js:1252-1439

Basically:

```javascript
    // ...
    scrollIntoView: () => {},
    onHighlightedIndexChange({highlightedIndex}) {
      if (highlightedIndex >= 0) {
        rowVirtualizer.scrollToIndex(highlightedIndex)
      }
    },
    // ...
```

I'm actually pretty ok with this. Two things though:

1. It'd be cool if rowVirtualizer.scrollToIndex didn't explode when passing it a negative index (cc @tannerlinsley, this happens when closing the menu in a typical downshift component)
2. It'd be cool if the built-in scrollIntoView didn't do anything if passed `undefined`.

If both of those happened, then my final solution would be as simple as:

```javascript
    onHighlightedIndexChange({highlightedIndex}) {
      rowVirtualizer.scrollToIndex(highlightedIndex)
    },
```

Which would be pretty cool 👍

tannerlinsley:
K, that's deploying in React-Virtual.

kentcdodds:
I made a PR for the other part: #1047. Pretty straightforward I think.

PR #1051, opened 2020-05-14
feat(hooks): memoize all callbacks

### LINKED ISSUE #1050: Add useCallback to all prop getters and actions
If I identify a performance issue and want to memoize my components that use prop getters (like `getItemProps`), the prop getters are breaking memoization:

```javascript
function Item({isHighlighted, getItemProps, item, index}) {
  return (
    <li
      style={isHighlighted ? {backgroundColor: '#bde4ff'} : {}}
      key={`${item}${index}`}
      {...getItemProps({item, index})}
    >
      {item}
    </li>
  )
}
Item = React.memo(Item)
```

https://codesandbox.io/s/usecombobox-usage-1fs67?file=/src/index.js:431-454

If you pull up the React DevTools and run a profile, you'll observe that Items are getting re-rendered because `getItemProps` changed:

image

The solution should be pretty straightforward. Simply apply `useCallback` to all prop getters and action functions. So long as all dependencies are included (is eslint-plugin-react-hooks installed on this repo?) then it should be a patch version update.

kentcdodds:
**What**:

This PR memoizes all callbacks returned from downshift so we can memoize function components and pass callbacks in dependency arrays.

**Why**: Closes #1050

**How**:

This was a bit tricky, but basically the way we made this work properly to avoid the issues brought up in #1050 is by making a `latest` ref which contains the latest version of `props` and `state` and then we could reference that in our memoized callbacks without having memoization broken whenever those values changed because our callbacks all only care about the latest values anyway (this is the way the class-based downshift works as well).

Had to memoize the `dispatch` function as well. Minimal impact there.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] TypeScript Types N/A
- [ ] Flow Types N/A
- [x] Ready to be merged

### DIFF
--- a/src/__mocks__/utils.js
+++ b/src/__mocks__/utils.js
@@ -1,4 +1,4 @@
-const actualUtils = require.requireActual('../utils')
+const actualUtils = jest.requireActual('../utils')
 module.exports = Object.assign(actualUtils, {
   scrollIntoView: jest.fn(), // hard to write tests for this thing...
 })
--- a/src/hooks/useCombobox/__tests__/memo.test.js
+++ b/src/hooks/useCombobox/__tests__/memo.test.js
@@ -1,0 +1,9 @@
+import {renderUseCombobox} from '../testUtils'
+
+test('functions are memoized', () => {
+  const {result, rerender} = renderUseCombobox()
+  const firstRenderResult = result.current
+  rerender()
+  const secondRenderResult = result.current
+  expect(firstRenderResult).toEqual(secondRenderResult)
+})
--- a/src/hooks/useCombobox/index.js
+++ b/src/hooks/useCombobox/index.js
@@ -1,5 +1,5 @@
 /* eslint-disable max-statements */
-import {useRef, useEffect} from 'react'
+import {useRef, useEffect, useCallback, useMemo} from 'react'
 import {isPreact, isReactNative} from '../../is.macro'
 import {handleRefs, normalizeArrowKey, callAllEventHandlers} from '../../utils'
 import {
@@ -7,6 +7,7 @@ import {
   getPropTypesValidator,
   updateA11yStatus,
   useMouseAndTouchTracker,
+  useLatestRef,
 } from '../utils'
 import {
   getInitialState,
@@ -49,10 +50,12 @@ function useCombobox(userProps = {}) {
   const initialState = getInitialState(props)
 
   // Reducer init.
-  const [
-    {isOpen, highlightedIndex, selectedItem, inputValue},
-    dispatch,
-  ] = useControlledReducer(downshiftUseComboboxReducer, initialState, props)
+  const [state, dispatch] = useControlledReducer(
+    downshiftUseComboboxReducer,
+    initialState,
+    props,
+  )
+  const {isOpen, highlightedIndex, selectedItem, inputValue} = state
 
   /* Refs */
   const menuRef = useRef(null)
@@ -65,6 +68,7 @@ function useCombobox(userProps = {}) {
   const isInitialMount = useRef(true)
   const elementIdsRef = useRef(getElementIds(props))
   const previousResultCountRef = useRef()
+  const latest = useLatestRef({state, props})
 
   const getItemNodeFromIndex = index =>
     itemRefs.current[elementIdsRef.current.getItemId(index)]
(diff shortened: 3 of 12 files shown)


kentcdodds:
With these changes my example goes from this:

image

to this:

image

src/__mocks__/utils.js
```diff
@@ -1,4 +1,4 @@
-const actualUtils = require.requireActual('../utils')
+const actualUtils = jest.requireActual('../utils')
```
kentcdodds, line 1: Was getting eslint errors for this. `require.requireActual` is deprecated in favor of `jest.requireActual`

src/hooks/useCombobox/__tests__/memo.test.js
```diff
@@ -0,0 +1,9 @@
+import {renderUseCombobox} from '../testUtils'
+
+test('functions are memoized', () => {
```
kentcdodds, line 3: This is just a simple test to make sure nothing is changed between rerenders when nothing has changed 🙃

Before my changes, this test would fail.

src/hooks/useCombobox/index.js
```diff
@@ -49,10 +50,12 @@ function useCombobox(userProps = {}) {
...
-  const [
-    {isOpen, highlightedIndex, selectedItem, inputValue},
-    dispatch,
-  ] = useControlledReducer(downshiftUseComboboxReducer, initialState, props)
+  const [state, dispatch] = useControlledReducer(
+    downshiftUseComboboxReducer,
+    initialState,
+    props,
+  )
+  const {isOpen, highlightedIndex, selectedItem, inputValue} = state
```
kentcdodds, line 58: Mostly formatting changes. I just wanted to be able to create the `state` variable so I could use it in `useLatestRef`.

src/hooks/useCombobox/index.js
```diff
@@ -65,6 +68,7 @@ function useCombobox(userProps = {}) {
   const isInitialMount = useRef(true)
   const elementIdsRef = useRef(getElementIds(props))
   const previousResultCountRef = useRef()
+  const latest = useLatestRef({state, props})
```
kentcdodds, line 71: This just exists so we can use `latest` in our callbacks without worrying about these values breaking memoization.

src/hooks/useCombobox/index.js
```diff
@@ -170,309 +174,339 @@ function useCombobox(userProps = {}) {
     },
   )
   /* Event handler functions */
-  const inputKeyDownHandlers = {
```
kentcdodds, line 173: Most of everything below here is formatting changes due to memoization. It will be minorly easier to review this [with whitespace changes ignored](https://github.com/downshift-js/downshift/pull/1051/files?w=1).

src/hooks/useCombobox/index.js
```diff
@@ -170,309 +174,339 @@ function useCombobox(userProps = {}) {
...
+        if (event.which === 229) {
+          return
+        }
+        const latestState = latest.current.state
 
-    if (!isOpen && inputRef.current) {
-      inputRef.current.focus()
-    }
-  }
+        if (latestState.isOpen && latestState.highlightedIndex > -1) {
```
kentcdodds, line 221: Everywhere we were referencing state directly in our new useMemo and useCallback hooks, we replaced with our `latest` ref so we don't have to worry about breaking memoization.

Note in general, this is not a great idea. There's a good reason that this isn't the default for React hooks. However, for the use case of this library and the way these callbacks are called (typically *during* the render phase), it makes perfect sense. We don't really need to worry about preserving old closures.

src/hooks/useCombobox/index.js
```diff
@@ -170,309 +174,339 @@ function useCombobox(userProps = {}) {
...
-          type: stateChangeTypes.InputKeyDownEnter,
+          type: stateChangeTypes.InputKeyDownArrowDown,
+          shiftKey: event.shiftKey,
           getItemNodeFromIndex,
         })
-      }
-    },
-  }
-
-  // Event handlers.
```
kentcdodds, line 225: All of the event handlers have been moved inside the prop getters that used them. This made it much easier to memoize. Otherwise we'd have to memoize each of these as well (which is fine, but just extra work for both us and the JavaScript engine).

src/hooks/useCombobox/index.js
```diff
@@ -170,309 +174,339 @@ function useCombobox(userProps = {}) {
...
+          event.preventDefault()
+          dispatch({
+            type: stateChangeTypes.InputKeyDownEnter,
+            getItemNodeFromIndex,
+          })
+        }
+      },
+    }),
+    [dispatch, latest],
+  )
```
kentcdodds, line 231: Git is funny... This is the end of the `inputKeyDownHandlers` useMemo memoization.

src/hooks/useCombobox/index.js
```diff
@@ -170,309 +174,339 @@ function useCombobox(userProps = {}) {
...
 
   // Getter props.
```
kentcdodds, line 233: The prop getter functions kinda got destroyed WRT the git diff. Here's what happened for each of them:

1. Memoized via `useCallback` with dependencies on only things which are also memoized or are refs (`dispatch`, `latest`, or things like `inputKeyDownHandlers`).
2. Moved all event handlers inside the prop getter function
3. Swapped any reference to props or state with `latest.current.props` and `latest.current.state`

Everything else is unchanged.

src/hooks/useCombobox/index.js
```diff
@@ -170,309 +174,339 @@ function useCombobox(userProps = {}) {
...
-    role: 'combobox',
-    'aria-haspopup': 'listbox',
-    'aria-owns': elementIdsRef.current.menuId,
-    'aria-expanded': isOpen,
-    ...rest,
-  })
+    [latest],
+  )
 
   // returns
```
kentcdodds, line 462: All the `actions` got memoized. These were pretty straightforward and all simply needed the `dispatch` as the dependency (which is itself memoized so they should be stable).

src/hooks/utils.js
```diff
@@ -151,7 +162,11 @@ export function useEnhancedReducer(reducer, initialState, props) {
     [reducer],
   )
   const [state, dispatch] = useReducer(enhancedReducer, initialState)
-  const dispatchWithProps = action => dispatch({props, ...action})
+  const propsRef = useLatestRef(props)
+  const dispatchWithProps = useCallback(
+    action => dispatch({props: propsRef.current, ...action}),
+    [propsRef],
+  )
```
kentcdodds, line 169: This is all just to memoize the `dispatch` function. Behavior stays the same, but now you can pass `dispatch` to dependency arrays.

silviuaavram:
I fully watched the part of this getting developed and it was awesome. Love these changes, thank you so much!

Closing https://github.com/downshift-js/downshift/issues/730 as well. I want Downshift to stay the same, and if people want to benefit from this performance gain, they should migrate to the hooks.

kentcdodds:
Sweet!

Note that you *can* actually memoize things with the regular Downshift component. I'll comment on there with that.

PR #1053, opened 2020-05-17
chore(hooks): add controlled props and getter props warnings

silviuaavram:
**What**:
Adding non-production build errors in the hooks similar to the ones in Downshift: for controlled-uncontrolled props switch and also for incorrect call of getter props.

**Why**:
Fixes https://github.com/downshift-js/downshift/issues/1024.
Fixes https://github.com/downshift-js/downshift/issues/974.

**How**:
Keep a ref with previous props and check with current props if there was a switch between controlled and uncontrolled state.
Keep a ref to keep information about whether the getter props were called and if were called correctly (ref was correctly set).

If issues are found, we console.error messages in non-production environments.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] TypeScript Types
- [ ] Ready to be merged

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,38 +1,38 @@
 {
   "dist/downshift.cjs.js": {
-    "bundled": 137552,
-    "minified": 62918,
-    "gzipped": 13275
+    "bundled": 146382,
+    "minified": 66785,
+    "gzipped": 14090
   },
   "preact/dist/downshift.cjs.js": {
-    "bundled": 136289,
-    "minified": 61899,
-    "gzipped": 13179
+    "bundled": 145166,
+    "minified": 65813,
+    "gzipped": 14000
   },
   "preact/dist/downshift.umd.min.js": {
-    "bundled": 147655,
-    "minified": 46960,
-    "gzipped": 12608
+    "bundled": 155865,
+    "minified": 50004,
+    "gzipped": 13246
   },
   "dist/downshift.umd.min.js": {
-    "bundled": 151848,
-    "minified": 48265,
-    "gzipped": 13163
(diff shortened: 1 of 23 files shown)


slorber:
Hi,

Is this normal the suppressWarning thing is not in the TS defs?

```ts
export interface UseComboboxPropGetters<Item> {
  getToggleButtonProps: (
    options?: UseComboboxGetToggleButtonPropsOptions,
  ) => any
  getLabelProps: (options?: UseComboboxGetLabelPropsOptions) => any
  getMenuProps: (options?: UseComboboxGetMenuPropsOptions) => any
  getItemProps: (options: UseComboboxGetItemPropsOptions<Item>) => any
  getInputProps: (options?: UseComboboxGetInputPropsOptions) => any
  getComboboxProps: (options?: UseComboboxGetComboboxPropsOptions) => any
}
```

Also, it's probably a good idea to make the warning more actionable. It's not really clear what the user should do if this warning happens. At least worth mentioning the `{suppressRefError : true}`.


Personaly I need this warning because I render the dropdown in a portal  / Popper and I can't let it remain mounted if there's 0 option available, as it has a perf impact.

silviuaavram:
Hi! Can you create a PR with improvement suggestions?

TrySound:
getMenuProps should not emit warning when menu is closed. It's not rendered.

TrySound:
I also get dozens of these errors for the same component. You probably log it on every render.
> downshift: You forgot to call the getMenuProps getter function on your component / element.

wootencl:
Seeing the same as @TrySound. Seems strange to emit the error when the menu is closed/not rendered.

PR #1058, opened 2020-05-20
fix(a11y): wrong result count spoken when items change

ghost:
**What**:

Added `items` and `getA11yStatusMessage` to the `useEffect` hook inside `useSelect` and `useCombobox` that updates the `a11y` status.

**Why**:

Screen readers don't read the a11y status that should be read, as far as I can tell. This happens because the `itemCount` and `previousItemCount` are always the same, due to the missing dependencies in the `useEffect` hook, which means the right text is never returned.

**Reproducing**:

Go to [this sandbox link](https://codesandbox.io/s/downshift-algolias-react-instantsearch-kveo5?file=/index.js) and open the console pane. Start typing into the input. You'll notice that it outputs what the screen reader *would* say, and that the item counts are actually wrong on each call, and that the `itemCount` and `previousItemCount` are identical on every call.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] TypeScript Types N/A
- [ ] Flow Types N/A
- [x] Ready to be merged

### DIFF
--- a/src/hooks/useCombobox/index.js
+++ b/src/hooks/useCombobox/index.js
@@ -108,7 +108,7 @@ function useCombobox(userProps = {}) {
       environment.document,
     )
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [isOpen, highlightedIndex, selectedItem, inputValue, items])
+  }, [getA11yStatusMessage, isOpen, highlightedIndex, selectedItem, inputValue, items])
   // Sets a11y status message on changes in selectedItem.
   useEffect(() => {
     if (isInitialMountRef.current) {
--- a/src/hooks/useCombobox/index.js
+++ b/src/hooks/useCombobox/index.js
@@ -108,7 +108,7 @@ function useCombobox(userProps = {}) {
       environment.document,
     )
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [isOpen, highlightedIndex, selectedItem, inputValue])
+  }, [isOpen, highlightedIndex, selectedItem, inputValue, items])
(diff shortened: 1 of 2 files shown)


silviuaavram:
I need a repro case / sandbox with the issue in order to move forward with this.

craigthompsonhive:
> Hi! I need a repro case / sandbox with the issue in order to move forward with this.

Sorry @silviuaavram, I've added a test case to the PR.

craigthompsonhive:
I've also now added the `getA11yStatusMessage` prop to the dependency list, to handle cases where the actual method changes between renders

silviuaavram:
Tracked the bug, it makes sense to have these changes. However I don't think we should update the status just because of the getter function being changed. It's not something that's user facing. 

I don't like not to put the exhaustive depends. in the useEffect but for me it makes sense to update the a11y status only when state changes or the number of items change. So I would leave it with only the `items`.

Please add some tests as well. Check if, on `rerender` with a different `items` prop, it will show a new message if the new items length is different.

craigthompsonhive:
> I don't like not to put the exhaustive depends in the useEffect but for me it makes sense to update the a11y status only when state changes or the number of items change. So I would leave it with only the `items`.

This would be fine, except for the fact it makes updating the a11y method impossible later since you can't be quaranteed the latest method will be used. `getA11yStatusMessage` *is* user facing, so this would be very counter intuitive from a users perspective.

I'll add the tests cases for `item`, though, as suggested.

craigthompsonhive:
Closing this as it's fixed by https://github.com/downshift-js/downshift/issues/1076

PR #1085, opened 2020-06-14
chore(getterProps): only check on first render if getter props are called

silviuaavram:
**What**:

Check the required getter props are called only on first render.

**Why**:
We now memoize the calls and checking the call of re-renders will trigger errors, if calls are not performed due to memoization.
Fixes https://github.com/downshift-js/downshift/issues/1077.

**How**:
Only check on first render if getter props are called. Keeping the ref check as it is.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] TypeScript Types N/A
- [ ] Flow Types N/A
- [x] Ready to be merged


silviuaavram:
@TrySound @jorgemoya @wootencl please take a look.

PR #1107, opened 2020-06-30
fix(useCombobox): fix onSelectedItemChange not triggering with controlled selectedItem in React.StrictMode

### LINKED ISSUE #1092: [useCombobox] onSelectedItemChange never triggered when using controlled selectedItem in React.StrictMode
- `downshift` version: 5.4.5
- `node` version: 12.16.1
- `npm` version: 6.13.4

**What you did**:

Using `useCombobox` with controlled `selectedItem` while on `React.StrictMode`.

**What happened**:

`onSelectedItemChange` is never triggered, making it impossible to update the `selectedItem` in state.

**Reproduction repository**:

https://codesandbox.io/s/usecombobox-usage-ywyct?file=/src/index.js

**Problem description**:

This issue looks similar to #927, but seems to have been reintroduced in [5.2.3](https://github.com/downshift-js/downshift/releases/tag/v5.2.3).

The problem might be linked to [StrictMode double-invoking some functions to warn about side effects](https://github.com/downshift-js/downshift/issues/403#issuecomment-381830044).

IwalkAlone:
**What**:

Make sure onSelectedItemChange triggers when selectedItem is controlled in strict mode (Fixes #1092)

**Why**:

To ensure correct functionality under react strict mode

**How**:

I found that a side-effect happens during render inside `useControlledReducer` when the `previousSelectedItemRef.current` is set.
In strict mode, because of double rendering, this executes multiple times when an item is clicked and causes a `ControlledPropUpdatedSelectedItem` action to be dispatched even though the prop wasn't updated. This action doesn't update `selectedItem` from the initial value provided through the controlled prop, so the change to it is lost (I'm not 100% sure I understood exactly now this works, but it's my best guess). 

I wrapped the side-effect in `useEffect()` as suggested in the React docs, which fixed the issue - no extra action is triggered and `useEnhancedReducer` picks up the changes and calls `onSelectedItemChange` as expected.

**Checklist**:

- [ ] Documentation [N/A]
- [x] Tests
- [ ] TypeScript Types [N/A]
- [ ] Flow Types [N/A]
- [x] Ready to be merged

### DIFF
--- a/src/hooks/useCombobox/__tests__/props.test.js
+++ b/src/hooks/useCombobox/__tests__/props.test.js
@@ -1,3 +1,4 @@
+import React from 'react'
 import {renderHook, act as hooksAct} from '@testing-library/react-hooks'
 import {act} from '@testing-library/react'
 import {renderCombobox, renderUseCombobox} from '../testUtils'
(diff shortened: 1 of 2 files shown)


silviuaavram:
@IwalkAlone can this change have a unit test?

IwalkAlone:
@silviuaavram Sure! I've added a unit test that would fail without the change in this PR, please review

JReinhold:
This is awesome! Would love to see this released, as it's something I'm currently struggling with. <3

silviuaavram:
Thank you everyone!

silviuaavram:
@all-contributors please add @IwalkAlone for code, test

PR #1113, opened 2020-07-08
fix(useSelect): ensure the initial selected item is scrolled into view on first open

### LINKED ISSUE #1106: When opening the item list for the first time, the initially selected item is not scrolled into view
- `downshift` version: 5.4.5
- `node` version: N/A (codesandbox)
- `npm` (or `yarn`) version: N/A (codesandbox)

**Relevant code or config**
https://codesandbox.io/s/useselect-usage-53y36

**What you did**:
I configured the initialSelectedItem property in the useSelect hook. I expected the selected item to be scrolled into view automatically.

**What happened**:
The item was only scrolled into view on the second opening of the select.

**Reproduction repository**:
https://codesandbox.io/s/useselect-usage-53y36

**Problem description**:
There seems to be some internal confusion about the hightlighted item. In the first opening `onHighlightedIndexChange` is also not triggered. On subsequent close/open it does get triggerred with -1 / selectedIndex. 

**Suggested solution**:
N/A

IwalkAlone:
**What**:

When the dropdown is first opened with a selected item, the item should be scrolled into view and onHighlightedItemChange should be called (Fixes #1106)

**Why**:

Because the behavior is inconsistent between the first and subsequent opens.

**How**:

The useSelect `reducer` makes sure that whenever the dropdown would be closed (`isOpen: false`), the `highlightedIndex` is reset to -1 at the same time. However, if an item is selected initially, `initialState` would create a state with `isOpen: false` and `highlightedIndex` set to the item's index. This causes `onHighlightedIndexChange` not to fire on first open, as the index doesn't change from the initial state.

I've changed `initialState` to only set the `highlightedIndex` if the dropdown is already open, to keep the state internally consistent.

**Checklist**:

- [ ] Documentation [N/A]
- [x] Tests
- [ ] TypeScript Types [N/A]
- [ ] Flow Types [N/A]
- [x] Ready to be merged

### DIFF
--- a/src/hooks/useCombobox/__tests__/props.test.js
+++ b/src/hooks/useCombobox/__tests__/props.test.js
@@ -1007,6 +1007,60 @@ describe('props', () => {
       expect(onHighlightedIndexChange).not.toHaveBeenCalled()
     })
 
+    test('is called on first open when initialSelectedItem is set', () => {
+      const index = 2
+      const onHighlightedIndexChange = jest.fn()
+      const {clickOnToggleButton} = renderCombobox({
+        initialSelectedItem: items[index],
+        onHighlightedIndexChange,
+      })
+
+      clickOnToggleButton()
+
+      expect(onHighlightedIndexChange).toHaveBeenCalledTimes(1)
+      expect(onHighlightedIndexChange).toHaveBeenCalledWith(
+        expect.objectContaining({
+          highlightedIndex: index,
+        }),
+      )
+    })
+
+    test('is called on first open when selectedItem is set', () => {
+      const index = 2
+      const onHighlightedIndexChange = jest.fn()
+      const {clickOnToggleButton} = renderCombobox({
+        selectedItem: items[index],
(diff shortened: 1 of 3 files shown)


silviuaavram:
Looks great! I was wondering if we also need to cover the cases for `defaultSelectedItem` and `selectedItem`. Also for `useCombobox`.

Can you check if these cases would be broken without your fix, write the tests and re-apply the `isOpen` fix?

Thanks very much!

IwalkAlone:
@silviuaavram Done. Can confirm that none of those cases worked without the fix.

silviuaavram:
Thank you so much @IwalkAlone !

pmweeks98:
Hey @IwalkAlone  & @silviuaavram, going to preform a bit of necromancy on this issue
I'm stuck using downshift 5.2.1 as its a downstream dependency of a downstream dependency of a dependency 

Would you be able to suggest a workaround for getting this fix to an older version?

PR #1121, opened 2020-07-21
(fix) UseMultipleSelection itemToString type

mattrothenberg:
**What**: Updates the `itemToString` type definition for the `UseMultipleSelection` interface.

**Why**: The current definition causes errors and is inconsistent with the other hooks.

```ts
Type '((item: T) => string) | undefined' is not assignable to type '((item: T | null) => string) | undefined'.
  Type '(item: T) => string' is not assignable to type '(item: T | null) => string'.
    Types of parameters 'item' and 'item' are incompatible.
      Type 'T | null' is not assignable to type 'T'.
        'T' could be instantiated with an arbitrary type which could be unrelated to 'T | null'.
          Type 'null' is not assignable to type 'T'.
            'T' could be instantiated with an arbitrary type which could be unrelated to 'null'.
```

**How**: A type definition update.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] TypeScript Types
- [ ] Flow Types N/A
- [x] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -576,7 +576,8 @@ export interface UseMultipleSelectionProps<Item> {
   selectedItems?: Item[]
   initialSelectedItems?: Item[]
   defaultSelectedItems?: Item[]
-  itemToString?: (item: Item) => string
+
+  itemToString?: (item: Item | null) => string
   getA11yRemovalMessage?: (options: A11yRemovalMessage<Item>) => string
   stateReducer?: (
     state: UseMultipleSelectionState<Item>,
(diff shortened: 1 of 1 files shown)


silviuaavram:
Hi! Not sure why you get the error, itemToString for the multiple hook has been left untouched in terms of types. And that is expected, since it is called on item deletion, which means it cannot receive a null. Can you point out the change in v6 that broke the types, maybe I am missing something.

mattrothenberg:
@silviuaavram Hmm, I think I see what you're saying. Closing this out.

PR #1149, opened 2020-08-14
Edit typing of selectItem to allow to set null

fabb:
**What**:

This PR enables to call `selectItem(null)` without having to typecast this at the call site in Typescript strict mode.

**Why**:

When `items` is a controlled prop, but `selectedItem` is not, it is necessary to manually reset the selectedItem after an action like `SelectedItemClick` has been executed. 

**How**:

Simple ;-)

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests - not sure if tests are necessary for a types-only change
- [x] TypeScript Types
- [x] Flow Types
- [ ] Ready to be merged
      

In the example at https://www.downshift-js.com/use-combobox, `selectItem(null)` is used, so I added `null` and not `undefined`.

### DIFF
--- a/flow-typed/npm/downshift_v2.x.x.js.flow
+++ b/flow-typed/npm/downshift_v2.x.x.js.flow
@@ -209,7 +209,7 @@ declare module downshift {
     openMenu: (cb?: Callback) => void;
     closeMenu: (cb?: Callback) => void;
     toggleMenu: (otherStateToSet?: {}, cb?: Callback) => void;
-    selectItem: (item: Item, otherStateToSet?: {}, cb?: Callback) => void;
+    selectItem: (item: ?Item, otherStateToSet?: {}, cb?: Callback) => void;
     selectItemAtIndex: (
       index: number,
       otherStateToSet?: {},
--- a/flow-typed/npm/downshift_v2.x.x.js.flow
+++ b/flow-typed/npm/downshift_v2.x.x.js.flow
@@ -45,7 +45,7 @@ declare module downshift {
     highlightedIndex: number | null;
     inputValue: string | null;
     isOpen: boolean;
-    selectedItem: Item;
+    selectedItem: Item | null;
   }
   declare export interface DownshiftProps<Item> {
     defaultSelectedItem?: Item;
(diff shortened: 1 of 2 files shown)


flow-typed/npm/downshift_v2.x.x.js.flow
```diff
@@ -209,7 +209,7 @@ declare module downshift {
     openMenu: (cb?: Callback) => void;
     closeMenu: (cb?: Callback) => void;
     toggleMenu: (otherStateToSet?: {}, cb?: Callback) => void;
-    selectItem: (item: Item, otherStateToSet?: {}, cb?: Callback) => void;
+    selectItem: (item: ?Item, otherStateToSet?: {}, cb?: Callback) => void;
```
fabb, line 212: i never used flow. is this correct syntax? i have copied it from here: https://flow.org/en/docs/types/maybe/

but maybe (hah), rather `null` should be explicitly stated instead of a maybe type to disallow `undefined`?

fabb: ok i changed it to `| null` and also edited the state type as well to include `null` for `selectedItem`, that was missing from the flow types, it was there for the typescript types already

silviuaavram:
Don't you have `reset` for this?

fabb:
Hm, maybe. I just fear it would reset too much, would need to try.
Still, the type changes make sense anyways since setting `null` is possible and has a desirable effect in some use cases.

silviuaavram:
I am all for flexibility, but here I think it's a little bit too much. So I am going to ask for a use you can only use `selectItem(null)` and not `reset()`. 

Thank you for the feedback!

fabb:
I have not yet tried it! The documentation of `reset` just says it resets to sensible defaults, this is a bit vague.

Also `selectItem(null)` would fit nicely with the other state setters, e.g. `setHighlightedIndex(-1)` is possible too. And it‘s even part of an example: https://www.downshift-js.com/use-combobox

levenleven:
Examples on https://www.downshift-js.com/use-combobox use `selectItem(null)` to clear selection, but typings do not allow this

silviuaavram:
Is this something that might be a breaking change? Adding another possiblity (null) shouldn't break anything in someone's typings, am I correct?

levenleven:
Change in typescript typings is not breaking (function will just allow wider input)
Change In flow typings `selectedItem: Item | null;` in `DownshiftState` is kinda breaking, but I think still right thing to do (item indeed can be null)

silviuaavram:
Then will keep this up for a major version bump.

silviuaavram:
Will add these changes in https://github.com/downshift-js/downshift/pull/1415.

PR #1150, opened 2020-08-15
fix(useCombobox): dispatch on Enter keydown event

silviuaavram:
**What**:

Call dispatch on `Enter` keydown.
Prevent event default (form submission) if menu is open.
Do not return highlightedIndex from reducer if menu is closed.
Fix the MUI example for useSelect on the docsite.

**Why**:
Fixes https://github.com/downshift-js/downshift/issues/1114.
Also to fix other potential issues.

**How**:
Fix the reducer for the `useCombobox` hook.
Add itemsAsObjects to the MUI example useCombobox hook.
Check state.isOpen before returning a `highlightedIndex` in the Home/End actions.

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged

### DIFF
--- a/docs/hooks/useSelect.mdx
+++ b/docs/hooks/useSelect.mdx
@@ -148,7 +148,7 @@ equialent of each item object.
   {() => {
     // import React from 'react'
     // import { useSelect } from 'downshift'
-    // import { items, useStyles } from './utils'
+    // import { itemsAsObjects, useStyles } from './utils'
     // import ExpandMoreIcon from '@material-ui/icons/ExpandMore'
     // import {Button, FormLabel, List, ListItem, ListItemText } from '@material-ui/core'
     function MaterialDropdownSelect() {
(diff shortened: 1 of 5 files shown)


fabb:
Haha, this broke my workaround for the old behavior 😆
Don't get me wrong, thanks for fixing! Thankfully I had unit tests in place using @testing-library.
The funny thing was that the tests were green locally, but red on the CI server. It took me a while to find out that I locally had 6.0.2 installed, and the CI server installed 6.0.5 which already includes this fix... (yes, I'm using a package-lock.json file + run `npm ci`, but unfortunately this dependency is one that is hoisted via lerna and therefore bypasses the package-lock.json)

silviuaavram:
How did you test that the submit event is not dispatched (if you did indeed manage to test it). @fabb

fabb:
In my custom component that uses the downshift hooks, I have a few callback props that get called when some state changes (like onBlur, onSelectedItemsChanged, onInputValueChanged). In the jest tests I pass a jest mock for these props, and check how often the mock callbacks have been called and with which parameters.

My unit tests have a lot more lines than the actual implementation 😅

PR #1170, opened 2020-09-01
fix(useCombobox): select item on blur only when menu is open

ndeom:
…andleBlur to ensure that items could not be selected by clicking the left arrow key when the dropdown was closed

**What**:

An additional conditional statement was added to prevent a bug where an item could be selected when using useMultipleSelection with useCombobox while the dropdown menu was closed

**Why**:

An issue for this change was made.

**How**:

Additional logic was added to ensure that in order for an item to be selected the dropdown menu must be open.

**Checklist**:

- [ X] Documentation
- [ X] Tests
- [ X] TypeScript Types
- [ X] Flow Types
- [ X] Ready to be merged
      

Yes.

There was technically an error during the commit to my local branch, but I believe this is due the value of "total" being 0 for test/setup.js, but I could be wrong.

image

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,23 +1,18 @@
 {
-  "downshift.cjs.js": {
-    "bundled": 150704,
-    "minified": 70561,
-    "gzipped": 14462
-  },
   "downshift.umd.min.js": {
-    "bundled": 155781,
-    "minified": 50361,
-    "gzipped": 13442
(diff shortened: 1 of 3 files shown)


silviuaavram:
Awesome work @ndeom !

silviuaavram:
@all-contributors please add @ndeom for code, bug

PR #1193, opened 2020-11-13
fix: UseComboboxGetComboboxPropsOptions typing

clementgarbay:
**What**:

I fixed the wrong typing of the `UseComboboxGetComboboxPropsOptions` interface. The generic type of the `HTMLProps` extend was `HTMLLabelElement` instead of `HTMLDivElement`.

**Why**:

From the related documentation (https://www.downshift-js.com/use-combobox), the DOM element where the `getComboboxProps` should be spread is `div`.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -500,7 +500,7 @@ export interface UseComboboxGetInputPropsOptions
     GetPropsWithRefKey {}
 
 export interface UseComboboxGetComboboxPropsOptions
-  extends React.HTMLProps<HTMLLabelElement> {}
+  extends React.HTMLProps<HTMLDivElement> {}
 
 export interface UseComboboxPropGetters<Item> {
   getToggleButtonProps: (


silviuaavram:
Rats! You're right, thanks you very much!

I see the build system is not feeling well. `58:21  error  `userEvent.type` does not need `await` operator  testing-library/no-await-sync-events`

I probably need to update the dependencies and fix this error before merging anything.

Thank you again for the contribution!

silviuaavram:
@clementgarbay can you rebase this PR? It should pass now. Thank you very much!

One question: this fix can't be considered a breaking change right? I don't think it's very probable for someone to use label related field without them working.

clementgarbay:
@silviuaavram I just rebased and the build seems now fine! I don't think this fix should be considered as a breaking change

silviuaavram:
@all-contributors please add @clementgarbay for code

PR #1210, opened 2020-12-26
perf(scroll): replace useEffect with useLayoutEffect

silviuaavram:
**What**:

Replace `useEffect` with `useLayoutEffect` in `useScrollIntoView`.

**Why**:
Prevent an unnecessary double paint.

**How**:
Replace the `useEffect` with `useLayoutEffect`.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged

### DIFF
--- a/src/hooks/utils.js
+++ b/src/hooks/utils.js
@@ -1,5 +1,11 @@
 import PropTypes from 'prop-types'
-import {useRef, useCallback, useReducer, useEffect} from 'react'
+import {
+  useRef,
+  useCallback,
+  useReducer,
+  useEffect,
+  useLayoutEffect,
+} from 'react'
 import {
   scrollIntoView,
   getNextWrappingIndex,
@@ -487,7 +493,7 @@ export function useScrollIntoView({
   // used not to scroll on highlight by mouse.
   const shouldScrollRef = useRef(true)
   // Scroll on highlighted item if change comes from keyboard.
-  useEffect(() => {
+  useLayoutEffect(() => {
     if (
       highlightedIndex < 0 ||
       !isOpen ||


src/hooks/utils.js
```diff
@@ -487,7 +493,7 @@ export function useScrollIntoView({
   // used not to scroll on highlight by mouse.
   const shouldScrollRef = useRef(true)
   // Scroll on highlighted item if change comes from keyboard.
-  useEffect(() => {
+  useLayoutEffect(() => {
```
TrySound, line 496: Does it work in SSR?

silviuaavram: I didn't consider this, oups. I see there is a hacky solution of checking if `window` does not exists and then using `useEffect` instead.

I was pretty confident that this is an improvement for avoiding a browser re-paint.

What do you think to do about it?

vikuscz: Hi, for example Next.js throws warnings if useLayoutEffect is used and some libraries (eg. `react-laag` and `react-use`) use that isomorfic implementation which you consider as hacky and it just works. So I am sure useLayoutEffect if `typeof window !== "undefined"` and useLayoutEffect else is absolutely sufficient🙂

silviuaavram: Thanks guys for pointing this out, I will fix this tomorrow with an isomorphic implementation and it should be available in 6.0.10.

silviuaavram: Created https://github.com/downshift-js/downshift/issues/1211. If you want to create a PR yourself I would be happy to review it. Thanks!

PR #1230, opened 2021-02-05
fix(typings): make item optional in GetItemPropsOptions

knpwrs:
**What**: Making `item` optional in `GetItemPropsOptions`

**Why**: The runtime works just fine with only an `index` value provided, but the typings _require_ an `item` to be provided.

**How**: vim

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] TypeScript Types
- [ ] Flow Types N/A
- [x] Ready to be merged
      

It may be worth looking into using `RequireAtLeastOne` from `type-fest`. I briefly attempted to use it for this library but it was proving difficult to test properly: https://github.com/sindresorhus/type-fest/blob/4c9835b3c42d7a9f1d10bae19c334ce1d8e8c8a4/source/require-at-least-one.d.ts

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -138,7 +138,7 @@ export interface GetPropsWithRefKey {
 export interface GetItemPropsOptions<Item>
   extends React.HTMLProps<HTMLElement> {
   index?: number
-  item: Item
+  item?: Item
   isSelected?: boolean
   disabled?: boolean
 }


silviuaavram:
Although I agree the current typings are not really according to the docs, I don't want to perform the change in this PR and make item optional. What do you mean by difficult to test properly?

knpwrs:
> What do you mean by difficult to test properly?

I was copying the types from `type-fest` into the project without installing the package and it wasn't working as I expected. Would the project maintainers be open to having `type-fest` be a dependency?

silviuaavram:
We can try. Let's see it implemented and if it works then we should consider adding them in the next breaking change release.

silviuaavram:
No activity here, closing, not sure if we can achieve this with typescript, but open to improvements. Basically, adjust the type to have either item or index, never none of them.

PR #1246, opened 2021-03-05
chore: Add type for GetLabelPropsReturnValue, to reduce usage of :any

greydnls:
**What**: Adding an interface to replace the `any` typing used for `getLabelProps` prop-getter return value. 

**Why**: 
To provide a more robust type system for downshift usage in typescript. 

**How**:
I essentially reused the interface from the flow type definition.                                             

**Checklist**:

- [ ] ~~Documentation~~ N/A
- [ ] ~~Tests~~ N/A
- [x] TypeScript~~ Types 
- [ ] ~~Flow Types~~ N/A
- [x] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -116,6 +116,11 @@ export interface GetInputPropsOptions
 export interface GetLabelPropsOptions
   extends React.HTMLProps<HTMLLabelElement> {}
 
+export interface GetLabelPropsReturnValue  {
+  htmlFor: string;
+  id: string;
+}
+
 export interface GetToggleButtonPropsOptions
   extends React.HTMLProps<HTMLButtonElement> {
   disabled?: boolean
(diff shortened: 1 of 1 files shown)


karlhorky:
@silviuaavram what do you think about this change?

I guess ideally, it would also take into account the props object value parameter passed in, eg:

```ts
getLabelProps: <Options extends GetLabelPropsOptions>(options?: Options) => Options & GetLabelPropsReturnValue
```

silviuaavram:
I agree with it, I am keeping this PR open until v7, since type changes are breaking changes. I am planning to re-write the hooks using typescript so will probably include this change as well. Thanks!

silviuaavram:
Actually, the way our API works, the getter props may return anything, since they return also the extra arguments you pass to them. getLabelProps({foo: 'bar'}) will return {foo: 'bar', htmlFor: 'something', id: 'something else'}.

This may require generics. `getLabelProps: <T = void>(options?: GetLabelPropsOptions & T) => T & {htmlFor: string, id: string}`.

It can be a great improvement to get rid of our `any` returns.

silviuaavram:
@karlhorky what do you think about
```ts
export interface GetLabelPropsReturnValue {
  htmlFor: string
  id: string
}
...
  getLabelProps: <Options>(
    options?: Options & GetLabelPropsOptions,
  ) => GetLabelPropsReturnValue & Options
```

karlhorky:
Nice, it's been a while since I've looked at this, so not sure off the top of my head, but seems like it could work?

PR #1261, opened 2021-04-11
Add ARIA W3 page link to README

GCrispino:
**What**: this PR adds a link to the W3.org [page](https://www.w3.org/TR/wai-aria-practices/) about ARIA.

**Why**: I was reading the README but didn't know what ARIA was, so I figured this might be useful for newcomers, specially noting that when combobox and select are mentioned in the same page, they have links to their respective sections in W3's ARIA specification.

**How**: A link was added to ARIA'S first mention in the "The Problem" README section.

**Checklist**:

- [x] Documentation
- [ ] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ autocomplete, combobox or select dropdown components.</p>
 
 You need an autocomplete, a combobox or a select experience in your application
 and you want it to be accessible. You also want it to be simple and flexible to
-account for your use cases. Finally, it should follow the ARIA design pattern
+account for your use cases. Finally, it should follow the [ARIA][aria] design pattern
 for a [combobox][combobox-aria] or a [select][select-aria], depending on your
 use case.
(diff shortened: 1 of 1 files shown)


GCrispino:
Yes, I think this might be useful to have a general grasp of what ARIA is, by linking to the specification document's "root".

About putting it in the hooks README files, I see how this would be added in the [main hooks README](https://github.com/downshift-js/downshift/tree/master/src/hooks), but not in each specific hook README.
For instance, the [`useSelect` README](https://www.downshift-js.com/use-select) mentions the ARIA design pattern, but this mention already contains a link to the specific select patttern.
What do you think?

silviuaavram:
I think that the main readme should be enough. Thanks!

PR #1264, opened 2021-04-16
fix(environment): use Node variable within the provided window context

KaiminHuang:
**What**: This change is made to address an issue being described in [Issue-1225](https://github.com/downshift-js/downshift/issues/1255)

**Why**: otherwise `downshift` does not work properly when rendering from an iframe to the parent window using React Portal

**How**: replace `instanceof Node` with `instanceof environment.Node`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged

### DIFF
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -1065,7 +1065,7 @@ class Downshift extends Component {
         const contextWithinDownshift = targetWithinDownshift(
           event.target,
           [this._rootNode, this._menuNode],
-          this.props.environment.document,
+          this.props.environment,
         )
         if (!contextWithinDownshift && this.getState().isOpen) {
           this.reset({type: stateChangeTypes.mouseUp}, () =>
@@ -1091,7 +1091,7 @@ class Downshift extends Component {
         const contextWithinDownshift = targetWithinDownshift(
           event.target,
           [this._rootNode, this._menuNode],
-          this.props.environment.document,
+          this.props.environment,
           false,
         )
         if (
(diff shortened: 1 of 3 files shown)


mxlje:
Unfortunately this is a breaking change when implementing a custom environment as is described in the readme. The readme links to [this gist](https://gist.github.com/Rendez/1dd55882e9b850dd3990feefc9d6e177), which implements only three properties on the environment: `doc`, `addEventListener`, `removeEventListener`. 

Before this change (6.1.2) everything worked as expected, but now that `Node` is expected to be available on the environment it throws an error when clicking an item to select it as my custom environment doesn’t implement it.

I understand that the default implementation uses `window` as the environment (on which `Node` is set), but the documentation around what the environment is expected to implement is misleading/incomplete.

What methods / properties does the environment actually need to respond to? How can I find out without searching through the entire code?

Thank you.

silviuaavram:
Sorry for the issue. We use the environment prop as the substitute for window, when needed. For instance, when we need to check where a click happened (as you can see in this PR), or to add the aria-live element that we use for a11y related messages (it will at some point go away, but now it's still here). There are also others (scroll into view) and probably more that I cannot think about now.

Any improvement is appreciated, just create an issue and a PR to fix it. Thanks!

rrabii:
@KaiminHuang @silviuaavram still doesn't work when using iframe on iOS browsers.

Any help would be appreciated

PR #1269, opened 2021-04-21
chore: update `validate.yml`

MichaelDeBoey:
https://nodejs.medium.com/node-js-16-available-now-7f5099a97e70

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -16,13 +16,11 @@ jobs:
     if: ${{ !contains(github.head_ref, 'all-contributors') }}
     strategy:
       matrix:
-        node: [12]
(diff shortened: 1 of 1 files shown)


silviuaavram:
@MichaelDeBoey it seems that it fails to build, should those react versions be updated/downgraded?

Also, since I know you worked on linting with typescript in Kent's repo, can you help me out with https://github.com/downshift-js/downshift/pull/1268? It seems that it doesn't find the typescript rules, but those are in the eslint repo https://github.com/kentcdodds/eslint-config-kentcdodds.

Thanks for the help! Let me know what you think!

silviuaavram:
Will merge the follow up PR instead.

PR #1271, opened 2021-05-04
chore: update `validate.yml`

silviuaavram:
Based on https://github.com/downshift-js/downshift/pull/1269


silviuaavram:
This is currently blocked by docz peer dependency of react. I don't want to downgrade everything as a consequence just yet.

PR #1286, opened 2021-06-10
Add `Node` to `Environment` TS type

joealden:
**What**:

Fix the `Environment` TypeScript type.

**Why**:

#1264 introduced the usage of `environment.Node`, so environments need to have this for Downshift to work correctly.

**How**:

Adds `Node` to the `Environment` TypeScript type.

**Checklist**:

- [ ] Documentation N/A?
- [ ] Tests N/A?
- [x] TypeScript Types
- [ ] Flow Types N/A?
- [x] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -79,6 +79,7 @@ export interface DownshiftProps<Item> {
 }
 
 export interface Environment {
+  Node: typeof window.Node
   addEventListener: typeof window.addEventListener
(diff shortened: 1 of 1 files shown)


silviuaavram:
Will cover this change in https://github.com/downshift-js/downshift/pull/1463. It needs some more stuff added, and it will considered a breaking change. Consequently, it will land in v8. Thank you for the change!

PR #1310, opened 2021-08-09
fix(TS): add tslib as a dependency

chandrasekhar1996:
**What**: Add tslib dependency in pacakge.json.
Closes https://github.com/downshift-js/downshift/issues/1309

**Why**: @rollup/plugin-typescript has tslib as a peer dependency and requires to be installed separately. The dependency not being there is currently causing an [issue](https://github.com/downshift-js/downshift/issues/1309). 

**How**: Added the dependency and ran npm run validate

**Checklist**:

- [N/A] Documentation
- [x] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged
      

For futuure PR: I don't have much idea about tslib but looking at the package, it would be a good improvement to use it in downshift-js as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,8 @@
     "@babel/runtime": "^7.14.8",
     "compute-scroll-into-view": "^1.0.17",
     "prop-types": "^15.7.2",
-    "react-is": "^17.0.2"
+    "react-is": "^17.0.2",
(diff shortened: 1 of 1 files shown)


silviuaavram:
@all-contributors please add @chandrasekhar1996 for bug, code

PR #1348, opened 2022-01-04
Fix error when attempting to highlight virtualised items with arrow keys

aaronnuu:
## What
When virtualising items, the 'getNextNonDisabledIndex' function would only check if the node didn't exist (and therefore was virtualised) once, so if you have a non-virtual item as the last item in the list, when pressing up on the arrow keys it could sometimes result in an error because it's checking for a property on a node that doesn't exist for that index

Example here: https://codesandbox.io/s/autumn-breeze-0m82p?file=/src/index.js

Instructions:
- Select one of the items that are visible on mount
- Focus the text input and remove all text
- Press up using the arrow keys

## Why
So that there can be static items that sit further down in the items list than the virtualised items

## How
I have made the 'getNextNonDisabledIndex' function recursive so that we are always using the same logic to check for the disabled (or virtualised) items.

The short circuits for this recursion are either when we have checked every index, or when we reach the end (or beginning) of the items, when the `circular` arg is `false`.

To achieve this I have added a new arg, `startIndex` which allows us to check if we've hit every index.

I also changed the arg `baseIndex` to `targetIndex` because I thought that it better describes the intention now, however I am definitely not stuck to this and can change it back if you think it makes sense.

## Caveats
If the virtualised item that is selected because we can't find the node is disabled, this function does not know about this and will return the `highlightedIndex` of that item anyway.
I have not fixed that in this PR as this bug is present in the current version of `downshift` at the moment and being able to fix it would require a much bigger refactor of the code

**Checklist**:

- [x] Tests - Passing
- [x] Ready to be merged

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -359,51 +359,67 @@ function getNextWrappingIndex(
  * Returns the next index in the list of an item that is not disabled.
  *
  * @param {number} moveAmount Number of positions to move. Negative to move backwards, positive forwards.
- * @param {number} baseIndex The initial position to move from.
+ * @param {number} targetIndex The initial position to move from.
  * @param {number} itemCount The total number of items.
  * @param {Function} getItemNodeFromIndex Used to check if item is disabled.
  * @param {boolean} circular Specify if navigation is circular. Default is true.
- * @returns {number} The new index. Returns baseIndex if item is not disabled. Returns next non-disabled item otherwise. If no non-disabled found it will return -1.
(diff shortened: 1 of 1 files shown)


silviuaavram:
So, is this not relevant anymore? The function changed here got some updates as part of v7 I think.

aaronnuu:
>So, is this not relevant anymore? The function changed here got some updates as part of v7 I think.

It might still be an issue for this pretty specific case, I don't have the capacity or the need to follow up on it anymore though.

PR #1353, opened 2022-01-13
fix(useMultipleSelect): removeSelectedItem adding duplicate items to selected items

### LINKED ISSUE #1316: useMultipleSelection hook's `removeSelectedItem` incorrectly adds an item
Hey! Love the library! I noticed an issue with `removeSelectedItem` that I wanted to notify you of and possibly contribute a fix for:

- `downshift` version: 6.1.7
- `node` version: 14.17
- `yarn` version: 1.22.10

**Overview**:
I was using `useMultipleSelection` with `useCombobox` and I had two items selected [A, B]. When I called `removeSelectedItem` with undefined as the first argument, my selected items changed to [A, A, B].

**Reproduction repository**:
You can see the bug at play in this [sandbox](https://codesandbox.io/s/downshift-usemultipleselection-remove-item-bug-wolws?file=/src/App.js).

**Problem description**:
If we look at the reducer for useMultipleSelection and step through the [`FunctionRemoveSelectedItem`](https://github.com/downshift-js/downshift/blob/master/src/hooks/useMultipleSelection/reducer.js#L70-L88) case we can see how this bug might arise. 
We don't need to pay attention to the `activeIndex` here because it's not used in this block to compute the `selectedItems` array.

```js
// globals
selectedItems = ['test1', 'test2']
selectedItem = undefined

case stateChangeTypes.FunctionRemoveSelectedItem: {
    //let newActiveIndex = activeIndex
    const selectedItemIndex = selectedItems.indexOf(selectedItem) // = -1

    //if (selectedItems.length === 1) { // false, doesn't enter
    //  newActiveIndex = -1
    //} else if (selectedItemIndex === selectedItems.length - 1) { // false, doesn't enter
    //  newActiveIndex = selectedItems.length - 2
    //}

  changes = {
    selectedItems: [
      ...selectedItems.slice(0, selectedItemIndex), // slice(0, -1) = ['test1']
      ...selectedItems.slice(selectedItemIndex + 1), // slice(0) = ['test1', 'test2']
    ], // = ['test1', 'test1', 'test2']
    ...{activeIndex: newActiveIndex},
  }
  break
}
```

**Suggested solution**:
If we determine the `selectedItemIndex` is less than 0, we can break out of this case early.

drewbrend:
fixes #1316

**What**:

Fixing the handling of undefined/null being provided to  `useMultipleSelection`'s `removeSelectedItem` function.

**Why**:

Currently when undefined/null is provided one of the items is duplicated in `selectedItems` 

**How**:

Only modifying the `selectedItems` and `activeIndex` if `selectedItemIndex` is greater than or equal to zero

**Checklist**:

- [x] Documentation N/A
- [x] Tests
- [x] TypeScript Types N/A
- [x] Flow Types N/A
- [x] Ready to be merged

### DIFF
--- a/src/hooks/useMultipleSelection/reducer.js
+++ b/src/hooks/useMultipleSelection/reducer.js
@@ -83,7 +83,7 @@ export default function downshiftMultipleSelectionReducer(state, action) {
             ...selectedItems.slice(0, selectedItemIndex),
             ...selectedItems.slice(selectedItemIndex + 1),
           ],
-          ...{activeIndex: newActiveIndex},
+          activeIndex: newActiveIndex,
         }
       }
       break
--- a/src/hooks/useMultipleSelection/reducer.js
+++ b/src/hooks/useMultipleSelection/reducer.js
@@ -71,18 +71,20 @@ export default function downshiftMultipleSelectionReducer(state, action) {
       let newActiveIndex = activeIndex
       const selectedItemIndex = selectedItems.indexOf(selectedItem)
 
-      if (selectedItems.length === 1) {
-        newActiveIndex = -1
-      } else if (selectedItemIndex === selectedItems.length - 1) {
-        newActiveIndex = selectedItems.length - 2
-      }
+      if (selectedItemIndex >= 0) {
+        if (selectedItems.length === 1) {
+          newActiveIndex = -1
+        } else if (selectedItemIndex === selectedItems.length - 1) {
+          newActiveIndex = selectedItems.length - 2
+        }
 
-      changes = {
-        selectedItems: [
-          ...selectedItems.slice(0, selectedItemIndex),
-          ...selectedItems.slice(selectedItemIndex + 1),
-        ],
-        ...{activeIndex: newActiveIndex},
+        changes = {
+          selectedItems: [
+            ...selectedItems.slice(0, selectedItemIndex),
+            ...selectedItems.slice(selectedItemIndex + 1),
+          ],
+          ...{activeIndex: newActiveIndex},
+        }
       }
       break
     }
--- a/src/hooks/useMultipleSelection/__tests__/returnProps.test.js
+++ b/src/hooks/useMultipleSelection/__tests__/returnProps.test.js
@@ -60,6 +60,34 @@ describe('returnProps', () => {
       expect(result.current.activeIndex).toEqual(0)
     })
 
+    test('removeSelectedItem handles undefined item without modifying the selected array or the active index', () => {
+      const {result} = renderUseMultipleSelection({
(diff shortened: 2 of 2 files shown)


lveillard:
Hello! Not sure but i think this PR could fix this also:
https://github.com/downshift-js/downshift/issues/1355

Will try it once is good.
Anything we can do to speed up the merge?
Thanks!

drewbrend:
> Hello! Not sure but i think this PR could fix this also: #1355
> 
> Will try it once is good. Anything we can do to speed up the merge? Thanks!

I don't think it will, downshift relies on the items being reference equal so you probably need to fix how you're handling/re-creating your items

lveillard:
so we have to drop all the good immutability practices? Why not being dependent on the content of the array instead of it's reference? Replacing/deleting object properties may not be the best option...

Also the example 5 is broken even just by adding some initial choices, no need to manipulate the object with .map() to caue the error.

And also it only happens with the inside of the popper, removing badges works as expected

silviuaavram:
Hi! Will take a look at this and will merge if it's good to go. Thank you!

lveillard:
Hello @silviuaavram would be lovely to get some news about this, I think this is creating some weird bugs like this one: https://github.com/downshift-js/downshift/issues/1355

drewbrend:
@lveillard this will prevent you from getting duplicated values but won't fix your issue, you need to provide an item _referentially equal_ to one of the provided items

silviuaavram:
Finally fixed the CI pipeline. Let's see what this PR.

src/hooks/useMultipleSelection/reducer.js
```diff
@@ -71,18 +71,20 @@ export default function downshiftMultipleSelectionReducer(state, action) {
...
-          ...selectedItems.slice(0, selectedItemIndex),
-          ...selectedItems.slice(selectedItemIndex + 1),
-        ],
-        ...{activeIndex: newActiveIndex},
+        changes = {
+          selectedItems: [
+            ...selectedItems.slice(0, selectedItemIndex),
+            ...selectedItems.slice(selectedItemIndex + 1),
+          ],
+          ...{activeIndex: newActiveIndex},
```
silviuaavram, line 86: I know it's not related to your change, but let's change this to `], activeIndex: newActiveIndex` instead of the thing I wrote initially.

silviuaavram (commented):
Great changes and tests, congrats!

I think we should rebase this with latest master, since it has been a while, do that small change and merge it.

Thank you again for your patience!

silviuaavram:
@all-contributors please add @drewbrend for code

PR #1382, opened 2022-05-17
fix(TS): allow selectedItem to accept array

vacas5:
What:

Change DownshiftProps interface selectedItem property to accept an array
of Item generic.

Why:

This is necessary to support the Array(any) for selectedItem prop as
outlined in the readme
(https://github.com/downshift-js/downshift#selecteditem)

How:

Update the typings index file

**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -61,7 +61,7 @@ export interface DownshiftProps<Item> {
   highlightedIndex?: number | null
   inputValue?: string | null
   isOpen?: boolean
-  selectedItem?: Item | null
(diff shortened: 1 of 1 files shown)


vacas5:
If that's the case then the docs need to be changed.

silviuaavram:
Can you point where, exactly? We will fix as needed.

PR #1416, opened 2022-09-27
fix: Improve `ControllerStateAndHelpers` typing

alan-nf:
**What**: Fix typings (TS & Flow)

**Why**: The `type` field [is set on the state](https://github.com/downshift-js/downshift/blob/1c7698080805e05f5180dd223cbc17932a85fd42/src/downshift.js#L351) that is passed to the `onInputValueChange` callback prop but the typings don't include it. This field is set correctly for other state change callbacks, it is just missing for this one.

**How**: Just edited index.d.ts to include the correct type

**Checklist**:

- [ ] Documentation n/a
- [ ] Tests n/a
- [x] TypeScript Types
- [x] Flow Types
- [x] Ready to be merged


silviuaavram:
I want to deprecate the component, and useCombobox has the type property correctly specified.

PR #1436, opened 2022-11-07
fix: add clean up to avoid effect firing twice in strict mode

jeanpan:
**What**:

Add clean up function to `isInitialMountRef` initialization. 

**Why**:

It fixes the bug that when the page load and if the dropdown is not above the content, the dropdown is focused and cases the page scrolled in strict mode. 

**How**:

As react [official doc](https://beta.reactjs.org/learn/synchronizing-with-effects#how-to-handle-the-effect-firing-twice-in-development), we should add clean-up function for initialization on effect to avoid the effect firing twice in the strict mode (development mode) 

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged

### DIFF
--- a/src/hooks/useCombobox/index.js
+++ b/src/hooks/useCombobox/index.js
@@ -139,6 +139,10 @@ function useCombobox(userProps = {}) {
   // Make initial ref false.
   useEffect(() => {
     isInitialMountRef.current = false
+
+    return () => {
+      isInitialMountRef.current = true
+    }
   }, [])
   // Reset itemRefs on close.
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


silviuaavram:
Hi! Thanks for the fix!

Can you run prettier over the code? The build fails because of too many statements, and it can be easily be fixed by prettier since it already has the rule from kcd-scipts.

jeanpan:
@silviuaavram, sorry for the late reply. Updated to fix the lint error.

silviuaavram (approved):

silviuaavram:
@all-contributors please add @jeanpan for code

PR #1445, opened 2022-11-23
feat/fix(ssr): Fix hydration errors, remove need for resetIdCounter when useId is available (i.e. React 18+) 

klarstrup:
This fixes hydration errors of the sort:
```
Warning: Prop `aria-label` did not match. Server: "downshift-1-label" Client: "downshift-0-label"
```
Even with proper usage of `resetIdCounter()`, this error can crop up when doing sufficiently concurrent SSR.

This PR adds a warning for when `resetIdCounter` is called unnecessarily. 

---

Following the addition of React 18 support the resettable hacky increment-only IDs can be dropped in favor of the first party `useId()` API (details: https://reactjs.org/docs/hooks-reference.html#useid)

This should not be a breaking change per se but the new IDs could result in broken snapshot tests among consumers.

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] TypeScript Types N/A
- [ ] Flow Types N/A
- [x] Ready to be merged
      

------

Prior art: https://github.com/hello-pangea/dnd/pull/430


silviuaavram:
Thank you! Will take a look!

PR #1451, opened 2022-12-05
docs: align docs with v7

YaseenAlGailani:
**What**: Remove reference to usage of shift modifier with arrow keys from useSelect and useCombobox docs

**Why**: Since v7, using shift modifier key + arrow key is not supported.


silviuaavram (approved):

silviuaavram:
Cool stuff, thank you very much!

PR #1457, opened 2023-01-04
fix: set `browserlist` so that js files are transpiled

tjenkinson:
**What**:

This PR changes the JS to be transpiled.

**Why**:

Unfortunately at the moment we cannot use downshift because it is tripping up our storybook and webpack builds due to some syntax that is too new, such as `??`.

**How**:

Configure the `browserlist` in package.json.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] TypeScript Types N/A
- [ ] Flow Types N/A
- [x] Ready to be merged

### DIFF
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -6,7 +6,15 @@ module.exports = defineConfig({
     baseUrl: 'http://localhost:6006',
     video: false,
     setupNodeEvents(on) {
-      on('file:preprocessor', webpackPreprocessor())
+      on(
+        'file:preprocessor',
+        webpackPreprocessor({
+          webpackOptions: {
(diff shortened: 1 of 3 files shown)


silviuaavram:
Hi! Thanks for the PR! Few questions:

What does the empty array actually mean? Does it revert to the default browserslist value? Shouldn't we use at least "defaults" or maybe install the browserslist package?

silviuaavram:
Makes sense. Thank you for the changes!

silviuaavram (approved):

silviuaavram:
@all-contributors please add @tjenkinson for fix, infra

PR #1477, opened 2023-03-01
fix(useCombobox): controlled selectedItem extra renders

mariusc23:
I noticed we are getting some flickering when updating the value of a combobox built using downshift. After much debugging the culprit seems to be the logic in `useControlledReducer`. 

We store the ID of the selected item as a query parameter and when updating the value through Next.js's router we get the flickering behavior. If I store the id in local state, the flickering doesn't happen.

As far as I can tell, the reason we see the flickering is because there is an extra render and this effect gets called again before the prop updated.

This PR changes to only run the effect when `props.selectedItem` or `state.selectedItem` change. This fixes the issue in our case. Making an effect run on every render seems not ideal anyway.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests
- [ ] TypeScript Types N/A
- [ ] Flow Types N/A
- [x] Ready to be merged

### DIFF
--- a/src/hooks/useCombobox/utils.js
+++ b/src/hooks/useCombobox/utils.js
@@ -105,7 +105,7 @@ function useControlledReducer(reducer, initialState, props) {
           ? props.selectedItem
           : state.selectedItem
     }
-  })
+  }, [props.selectedItem, state.selectedItem])
 
   return [getState(state, props), dispatch]
 }


silviuaavram:
Thank you @mariusc23 ! The changes make sense. Let me take a look as well on local and I will merge this.

silviuaavram (approved):

PR #1482, opened 2023-03-13
types: return values for getter props

silviuaavram:
**What**:
Replace `any` with proper return values in getter props Typescript types.
Return value should be Downshift returns + Props passed as Options, which, in turn, should be HTML props for those elements + any optional props passed by the user.

**Why**:
To better reflect the return value of the getter props.

Use cases:

- props accepted by getter function should stay the same.
- props returned by getter function should be the ones returned by Downshift / hooks.
- ~~if prop types are passed to getter that have a different type from the one returned by Downshift / hooks, the type of the returned prop should be changed. For example, `getLabelProps({htmlFor: true})` will have the return type `{id: string, htmlFor: true}`.~~
- returned props should contain the custom props passed to the getter function. For example, `getLabelProps({foo: 123})` will have the return type `{id: string, htmlFor: string, foo: number}`
- props that are returned by getter function by default should be overriden in type if user passes the same prop as getter prop argument. For instance, `getToggleButtonProps({"aria-haspopup": "dialog"})["aria-haspopup"]` will have its type changed from `listbox` to `dialog`.

**How**:
```ts
type Overwrite<T, U> = Pick<T, Exclude<keyof T, keyof U>> & U

export interface GetLabelPropsOptions
  extends React.HTMLProps<HTMLLabelElement> {}

export interface GetLabelPropsReturnValue {
  htmlFor: string
  id: string
}

export interface PropGetters<Item> {
  getLabelProps: <Options>(
    options?: GetLabelPropsOptions & Options,
  ) => Overwrite<GetLabelPropsReturnValue, Options>
  // other getters
}
```

In VScode I'm getting this result in the `basic.test.tsx`:

Screenshot 2023-03-13 at 10 20 36

**Checklist**:

- [x] Documentation
- [ ] Tests
- [x] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged
      

_Deprecated implementation that used `Overwrite` for the getter prop function arguments was dropped. This was problematic because it allowed passing any type for options that were used by the getter props._
```ts
type Overwrite<T, U> = Pick<T, Exclude<keyof T, keyof U>> & U

export interface GetLabelPropsOptions
  extends React.HTMLProps<HTMLLabelElement> {}

export interface GetLabelPropsReturnValue {
  htmlFor: string
  id: string
}

export interface PropGetters<Item> {
  getLabelProps: <Options>(
    options?: Overwrite<GetLabelPropsOptions, Options>,
  ) => Overwrite<GetLabelPropsReturnValue, Options>
  // other getters
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,6 +95,7 @@
     "@testing-library/user-event": "^14.4.3",
     "@types/jest": "^26.0.24",
     "@types/react": "^17.0.15",
+    "@types/react-native": "^0.71.3",
     "@typescript-eslint/eslint-plugin": "^4.28.5",
     "@typescript-eslint/parser": "^4.28.5",
     "babel-plugin-macros": "^3.1.0",
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -505,7 +505,7 @@ class Downshift extends Component {
       role: 'combobox',
       'aria-expanded': isOpen,
       'aria-haspopup': 'listbox',
-      'aria-owns': isOpen ? this.menuId : null,
+      'aria-owns': isOpen ? this.menuId : undefined,
       'aria-labelledby': this.labelId,
       ...rest,
     }
--- a/src/downshift.js
+++ b/src/downshift.js
@@ -833,7 +833,7 @@ class Downshift extends Component {
       'aria-activedescendant':
         isOpen && typeof highlightedIndex === 'number' && highlightedIndex >= 0
           ? this.getItemId(highlightedIndex)
-          : null,
+          : undefined,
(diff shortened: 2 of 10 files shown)


silviuaavram:
> if prop types are passed to getter that have a different type from the one returned by Downshift / hooks, the type of the returned prop should be changed. For example, getLabelProps({htmlFor: true}) will have the return type {id: string, htmlFor: true}.

Not sure why this is a good idea, as some options are going to be used by the getter props, and they need to have a correct type. Will change the types to show an error if stuff such as `htmlFor` or `onKeyDown` are passed with a wrong type.

PR #1486, opened 2023-03-15
fix: change getToggleButtonProps typing to support div

drewbrend:
**What**:

`getToggleButtonProps` should support HTMLButtonElement **_or_** HTMLDivElement props

**Why**:

[In the downshift docs](https://www.downshift-js.com/use-select/), `useSelect` recommends applying `getToggleButtonProps` to a div element but `getToggleButtonProps` specifies its `options` are specific to a button.  This works fine, unless you want to provide options to `getToggleButtonProps` that are unique to a div element, in my case I ran into this when providing a `ref` to my HTMLDivElement - this conflicted because `Ref<HTMLDivElement>` is not applicable to `Ref<HTMLButtonElement>`

**How**:

Simply added `| HTMLDivElement` to its type.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -117,7 +117,7 @@ export interface GetLabelPropsOptions
   extends React.HTMLProps<HTMLLabelElement> {}
 
 export interface GetToggleButtonPropsOptions
-  extends React.HTMLProps<HTMLButtonElement> {
+  extends React.HTMLProps<HTMLButtonElement | HTMLDivElement> {
   disabled?: boolean
 }
 


silviuaavram:
Thanks @drewbrend for the fix! I will make sure to address it in https://github.com/downshift-js/downshift/pull/1482. The reason is that only the UseSelectToggleButtonPropsOptions should have this change, which is going to most probably be only HTMLElement, since someone may want to use anything else other than a div. We are using a div, aria uses a div in their example, but that's not really a hard requirement. It most probably should not be a button, since button with a role of combobox is not really correct, but it might not be a div anyway.

And it's going to be a breaking change, so expect this to arrive in v8.

PR #1489, opened 2023-03-19
feat(reactNativeWeb): support in Downshift and hooks

silviuaavram:
**What**:
Support React Native Web in the hooks and in Downshift.

**Why**:
React Native is supported fully, we also should support React Native Web.

**How**:
As recommended in https://github.com/downshift-js/downshift/issues/1315#issuecomment-1474289581, we will create a react native web build and use the isReactNativeWeb variable to return the correct code in React native web (`onPress`, `event.nativeEvent.text` etc).

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,9 +9,10 @@
   "sideEffects": false,
   "browserslist": [],
   "scripts": {
-    "build": "npm run build:web --silent && npm run build:native --silent",
+    "build": "npm run build:web --silent && npm run build:native --silent && npm run build:nativeWeb --silent",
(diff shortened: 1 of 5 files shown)


DaveWelling (approved):
I haven't had much success with unit testing react native, so I can understand why there aren't any for this PR.  I wish I had this figured out so I could contribute that aspect.  Maybe I'll get it worked out in our project at sstid (my work) and then I can contribute some more of that here.  In any case it is what it is for now, and FWIW I don't see any obvious problems here.

PR #1530, opened 2023-08-08
Update prop types to allow for `environment` to be an empty object (the default)

zmdavis:
**What**:

Allows for the `environment` type to be an empty object in both the propTypes and Typescript

**Why**:

The [default value](https://github.com/downshift-js/downshift/blob/master/src/downshift.js#L112) of environment is set to an empty object ({}) during server-side rendering (or in any environment where window is undefined), but an empty object violated the [shape defined](https://github.com/downshift-js/downshift/blob/master/src/downshift.js#L54-L64) for environment in prop types. This in turn triggers a console.warn.

**How**:

In https://github.com/downshift-js/downshift/issues/1525 it was suggested to remove `isRequired` from all properties of `environment`, but this had a few downsides:  
1. If `environment` is provided, it should be of the proper shape
2. Typescript types and prop types would diverge unless we also updated the Typescript types (which would exacerbate the first problem)

Instead, I opted to allow for `environment` to be an empty object in both the propTypes and Typescript typings. However, if `environment` is specified (and not an empty object) it will be properly typed.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] TypeScript Types
- [x] Flow Types
- [x] Ready to be merged
      

Fixes https://github.com/downshift-js/downshift/issues/1525


src/downshift.js
```diff
@@ -52,15 +52,15 @@ class Downshift extends Component {
...
+      removeEventListener: PropTypes.func,
       document: PropTypes.shape({
         createElement: PropTypes.func.isRequired,
         getElementById: PropTypes.func.isRequired,
         activeElement: PropTypes.any.isRequired,
         body: PropTypes.any.isRequired,
-      }).isRequired,
-      Node: PropTypes.func.isRequired,
+      }),
+      Node: PropTypes.func,
```
zmdavis, line 63: Note that these changes as is would mean the prop types and Typescript types will diverge slightly. 

I typed out some potential options but realized that the most accurate is probably just to have the types reflect the reality that `environment` can either be a full `Environment` (with required props) or an empty object. Will make that change.

silviuaavram:
Closing this in favour of https://github.com/downshift-js/downshift/pull/1538. That should handle the prop type issue, along with other improvements.

silviuaavram:
@all-contributors please add @zmdavis for code, bug.

PR #1532, opened 2023-08-18
Updated onBlur to always trigger when input is blurred

jordanoverbye:
**What**:

When using the `stateReducer` function of `useCombobox`, the `InputBlur` change type only gets called if the combobox is currently open.

**Why**:

I believe the current behaviour is confusing and unexpected behaviour. I believe `InputBlur` should always be called when the input is blurred. Consumers of downshift have access to the current state and next changes in the state reducer, so they could replicate the current behaviour if they wish.

**How**:

I've made this change by removing the if statement for checking if combobox is opened in the `inputHandleBlur` function.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged

Thank you 🙂

### DIFF
--- a/src/hooks/useCombobox/index.js
+++ b/src/hooks/useCombobox/index.js
@@ -419,10 +419,7 @@ function useCombobox(userProps = {}) {
       }
       const inputHandleBlur = event => {
         /* istanbul ignore else */
-        if (
-          latestState.isOpen &&
-          !mouseAndTouchTrackersRef.current.isMouseDown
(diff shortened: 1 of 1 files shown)


silviuaavram:
The `stateReducer` is called only before downshift's state changes. On input blur with the menu closed, downshift's state does not change. So `stateReducer` should not be called in this case.

PR #1533, opened 2023-08-30
Customizable useSelect input debounce wait

tpict:
**What**: The hard-coded 500ms wait until `useSelect`'s input is cleared is being changed to a customizable prop.

**Why**: So this behavior can be customized or opted out of.

**How**: Replaced the hard-coded number with a prop. If the customized value is below zero, the value won't be cleared at all.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged

Let me know what you think, and I can write docs etc if needed 🙂

### DIFF
--- a/src/hooks/useSelect/index.js
+++ b/src/hooks/useSelect/index.js
@@ -40,6 +40,7 @@ function useSelect(userProps = {}) {
     itemToString,
     getA11ySelectionMessage,
     getA11yStatusMessage,
+    inputCleanupDebounceWait,
   } = props
   // Initial state depending on controlled props.
   const initialState = getInitialState(props)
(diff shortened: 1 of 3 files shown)


silviuaavram:
Hi @tpict !

I can understand the need for having a custom value option instead of the hardcoded `500ms`, but I don't understand the need to never clear the search query. Once you start typing stuff, but mistyped something, or just want to start typing for something else, you will never be able to do it, consequently you will get stuck and unable to use the character keys to search for anything, unless you remount the component.

Can you describe the use case better, please? Thanks!

PR #1536, opened 2023-09-05
fix(typescript): missing generic parameter of React.RefObject

fabb:
**What**:

Adds the generic type parameter for React.RefObject.

**Why**:

In `@types/react`, the type definition looks like this:

```
    interface RefObject<T> {
        readonly current: T | null;
    }
```

This means the `T` generic parameter needs to be set mandatorily. 

In case `skipLibCheck` is **not** set to `true`, this causes a typescript type error:

```
../../node_modules/downshift/typings/index.d.ts:112:9 - error TS2314: Generic type 'RefObject<T>' requires 1 type argument(s).

112   ref?: React.RefObject
            ~~~~~~~~~~~~~~~
```

**How**:

I've added `any` as type parameter, which is not ideal but avoids adding a generic type parameter to the parent types (e.g. `GetRootPropsOptions`).

Open for better suggestions.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] TypeScript Types
- [ ] Flow Types N/A
- [ ] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -109,7 +109,7 @@ type StateChangeFunction<Item> = (
 
 export interface GetRootPropsOptions {
   refKey?: string
-  ref?: React.RefObject
+  ref?: React.RefObject<any>
 }
 
 export interface GetRootPropsReturnValue {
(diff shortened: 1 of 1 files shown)


silviuaavram:
The HTML element proposal will not work with React Native. I believe that having `any` at the moment is good enough to unblock us, and the overall types are already an improvement over the previous ones. If anyone can propose a better way with generics, we can change it later.

silviuaavram (approved):

PR #1559, opened 2023-12-16
fix(type): Nullable setHighlightedIndex

miya:
**What**:

Changed the argument (index) of `setHighlightedIndex`  from number to number | null.

**Why**:

The implementation is supposed to allow null, but the interface did not accept null. Also it seems that a pull request implementing an approach that respects null, as also mentioned in #66 has been merged

**How**:

I've only made corrections to the interface.

**Checklist**:

- [x] Documentation
- [ ] Tests
- [x] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged


silviuaavram:
Why `null`? What's the use case to use `null` and not `-1`?

miya:
Indeed, I think using `-1` should be sufficient. I

've adjusted the interface to accommodate the fact that the implementation allows for null. If it's unnecessary, feel free to close."

PR #1568, opened 2024-01-30
fix: align useCombobox getToggleButtonProps type with useSelect

drewbrend:
**What**:

Allow comboboxes to have a toggle button of any element type in TypeScript (not just an HTMLButtonElement)

**Why**:

useCombobox's getToggleButtonProps was only compatible with HTMLButtonElement

useSelect is compatible with any HTMLElement

This is related to my previous PR: https://github.com/downshift-js/downshift/pull/1486

I am now upgrading our repo to downshift V8 and noticed that this is fixed for selects, however comboboxes are still locked to using a button by this type.

**How**:

I simply changed HTMLButtonElement to HTMLElement to match select's behaviour.

I am not familiar enough with these types to say for sure, but it seems like we could simply align [UseSelectGetToggleButtonPropsOptions](https://github.com/downshift-js/downshift/blob/master/typings/index.d.ts#L400-L404) with this [UseComboboxGetToggleButtonPropsOptions](https://github.com/downshift-js/downshift/blob/master/typings/index.d.ts#L601-L603)?  The only difference after this PR is that `UseComboboxGetToggleButtonPropsOptions` also has a `disabled` prop (through [GetToggleButtonPropsOptions](https://github.com/downshift-js/downshift/blob/master/typings/index.d.ts#L152-L155)).  Are implementations using `useSelect` able to disable the toggle button?  If so, I think these could be simplified to use a shared type.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -150,7 +150,7 @@ export interface GetLabelPropsReturnValue {
 }
 
 export interface GetToggleButtonPropsOptions
-  extends React.HTMLProps<HTMLButtonElement> {
+  extends React.HTMLProps<HTMLElement> {
   disabled?: boolean
   onPress?: (event: React.BaseSyntheticEvent) => void
 }


drewbrend:
🦗

PR #1571, opened 2024-02-19
fix(onMouseMove): only call handler on non-touch devices

silviuaavram:
**What**:
Don't call the onMouseMove handler when a touch event is detected.

**Why**:
Not really needed, since on mobile we don't highlight items with touch.
It also causes a sequence of batched dispatch calls that are difficult to understand and, most probably, wrong.
On mobile, when touching to select an item, we had a couple of batched dispatch calls:
1. item click
2. item mouse move + item click.
To me, it did not make sense. On non-mobile, it was fine, mouse move when mouse actually moved, and click after click. Nothing else.

Now we should only get itemClick on mobile when we touch to select an item.
Closes https://github.com/downshift-js/downshift/issues/1540.
Closes https://github.com/downshift-js/downshift/issues/1566.
Closes https://github.com/downshift-js/downshift/issues/1447.

It should also fix our example with [Basic multiple selection](https://www.downshift-js.com/use-combobox#basic-multiple-selection). Once this PR will get merged, the example also needs to have the `selectedItem` passed as `null`, as the example already does not care about it since we keep the selectedItems as custom state. Passing selectedItem as null controlled prop will also make the onSelectedItemChange get called on successive item clicks.

**How**:
Check if we have onTouchDown, and if so, don't call dispatch in the onMouseMove handler. We update the `useMouseAndTouchTracker` to provide us with the value.
Also a few changes:
1. Don't call `ControlledPropUpdatedSelectedItem` on first render, since we already get the proper selectedItem + input value combination in the initial state setup.
2. Use a custom hook to get isInitialRender info.
3. Check if state changed by also merging previous state with actual props, since that makes sense in the controlled component scenario.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged

### DIFF
--- a/src/hooks/__tests__/utils.test.js
+++ b/src/hooks/__tests__/utils.test.js
@@ -149,7 +149,7 @@ describe('utils', () => {
       )
 
       expect(result.current).toEqual({
-        current: {isMouseDown: false, isTouchMove: false},
+        current: {isMouseDown: false, isTouchMove: false, isTouchEnd: false},
       })
     })
   })
--- a/src/hooks/useCombobox/__tests__/getItemProps.test.js
+++ b/src/hooks/useCombobox/__tests__/getItemProps.test.js
@@ -280,6 +280,37 @@ describe('getItemProps', () => {
         await mouseMoveItemAtIndex(disabledIndex)
         expect(input).toHaveAttribute('aria-activedescendant', '')
       })
+
+      // Test that we don't call the mouse move handler on mobile.
+      test('does not set highlight the item if a touch event ocurred', async () => {
+        let touchEndHandler
+        const index = 1
+
+        renderCombobox({
+          isOpen: true,
+          environment: {
+            addEventListener: (name, handler) => {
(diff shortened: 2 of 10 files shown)


jeffijoe:
This also happens when using browser automation (Playwright in my case), and the fix for mobile doesn't apply there. Adding a `delay` to the `click` action in Playwright is a workaround but I don't want that to be a permanent solution.

Basically what seems to be happening is the item's `onMouseMove` is triggered in the same frame as the `click` (which doesn't normally happen on non-mobile devices).

I'm working on a simple reproduction repository including the Playwright setup and will open a new issue.

jeffijoe:
I have reproduced this with Playwright and opened an issue https://github.com/downshift-js/downshift/issues/1597

silviuaavram:
as long as this works for the end user, I would consider it closed. please reopen if you can repro as a normal user. thanks!

Issue #1574, opened 2024-02-28
useComboBox always grabbing focus on render


FredrikMWold:
- `downshift` version: 8.3.2
- `node` version: v18.19.0
- `npm` (or `yarn`) version: 10.2.3

**Relevant code or config**

```javascript
  useEffect(() => {
    if (isInitialMount) {
      return
    }

    previousResultCountRef.current = items.length
  })
  // Focus the toggle button on first render if required.
  useEffect(() => {
    const focusOnOpen = getInitialValue(props, 'isOpen')
    if (focusOnOpen && toggleButtonRef.current) {
      toggleButtonRef.current.focus()
    }
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [])
```

**What you did**:

**What happened**: Since useRef returns an object the if statment will always be true

**Reproduction repository**:

**Problem description**: When rendered the component always grabs focus

**Suggested solution**: `if (isInitialMount)` -> `if (isInitialMount.current)`

PhilipSampson:
Agreed, this is a problem for me too. Certainly looks to be attributable to the 8.3.1 -> 8.3.2 update.

Powerplex:
This is also bugged on `useSelect`, we updated our dropdown to use Downshift `8.3.2` and now it focuses the last dropdown on the page (and scrolls it into view): https://sparkui.vercel.app/?path=/docs/components-dropdown--docs


https://github.com/downshift-js/downshift/assets/2033710/78033851-54ed-4e12-bb61-b7e8a054b8db

silviuaavram:
Will fix this one tomorrow unless someone else creates a PR first.

PR #1575, opened 2024-02-29
fix: fixes issue causing combobox to grab focus on render

FredrikMWold:
**What**: fixes issue causing combobox to grab focus on render

**Why**: Closes: #1574

**How**: The latest release adds `isInitalMount` to the dependency array. Whichs causes it to rerender when `isInitialMount`  is set to false and triggering `.focus.`

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged

### DIFF
--- a/src/hooks/useMultipleSelection/index.js
+++ b/src/hooks/useMultipleSelection/index.js
@@ -93,7 +93,8 @@ function useMultipleSelection(userProps = {}) {
     } else if (selectedItemRefs.current[activeIndex]) {
       selectedItemRefs.current[activeIndex].focus()
     }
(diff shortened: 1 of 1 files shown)


FredrikMWold:
Would say this is the simplest solution. Another way if you dont want to disable eslint for that line is to change the useIsIninitalMount to return the ref instead of the current value. Then it would be a stable reference in the dependency array.

silviuaavram:
Let's keep it like that, for now. Thanks!

silviuaavram:
@all-contributrs please add @FredrikMWold for code, bug

PR #1585, opened 2024-03-27
Fix bugs in new `getA11yStatusMessage` option in hooks

aliceHendicott:
**What**:

1. Only update the A11Y message in the document if a non empty message is given
2. Update the return type to include `undefined` as an option

**Why**:

If using the `getA11yStatusMessage` prop to apply an a11y message for a specific scenario and returning undefined in other scenarios, it is possible for the `getA11yStatusMessage` function to run once to set the message, then again and sets undefined overriding anything previously set (because of the `debounce` function used).

I've also updated the types to allow returning of `undefined` (this is done in your examples of this function).

I've prepared a [CodeSandbox](https://codesandbox.io/p/sandbox/downshift-accessible-message-5g67dt?file=%2Fsrc%2FApp.tsx%3A12%2C17) showing these two issues.

- This example uses the `useMultipleSelection` hook and applies the example `getA11yStatusMessage` provided in your docs.
- If you select an option and then remove it by clicking the 'X' on the option, you'll see Voiceover does not announce that it has been removed. This is because an additional state update is overriding this to an empty message.
- You can also see the type error complaining about returning `undefined` instead of `string`

**How**:

**Checklist**:

- [ ] Documentation - N/A
- [x] Tests
- [x] TypeScript Types
- [ ] Flow Types - N/A
- [x] Ready to be merged

### DIFF
--- a/src/hooks/useCombobox/__tests__/props.test.js
+++ b/src/hooks/useCombobox/__tests__/props.test.js
@@ -285,6 +285,25 @@ describe('props', () => {
         'a11y-status-message',
       )
     })
+
+    test('is not overriden if it is called once with a value and again without', async () => {
+      const a11yStatusMessage = 'bla bla'
+      const getA11yStatusMessage = jest
+        .fn()
+        .mockReturnValueOnce(a11yStatusMessage)
+        .mockReturnValueOnce(undefined)
+      renderCombobox({
+        items,
+        getA11yStatusMessage,
+      })
+
+      await clickOnToggleButton()
+      await keyDownOnInput('{ArrowDown}')
+      waitForDebouncedA11yStatusUpdate()
+
+      expect(getA11yStatusMessage).toHaveBeenCalledTimes(2)
+      expect(getA11yStatusContainer()).toHaveTextContent(a11yStatusMessage)
+    })
   })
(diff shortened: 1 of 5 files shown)


aliceHendicott:
I can't seem to get the validate script to pass, when I run all the test scripts individually locally they pass so am not sure what I'm missing here, would be great to get some support on that.

silviuaavram:
Hi, @aliceHendicott! Thanks for the effort! I am looking over the v9 changes and you are right, there is an issue over there. It's about the dependency array being passed to the useEffect that triggers the a11y status message.

I will go ahead and fix this asap, already did the changes, and we can check whether this also fixed your issue.

About the `undefined` part, I would probably just change the docs examples, and return an empty string instead of undefined. This way we don't have to update the types and returning an empty string is not confusing, I believe. What do you think?

We can do a status update once 9.0.1 will land and we can check your use case afterwards. Thank you again!

silviuaavram:
https://github.com/downshift-js/downshift/pull/1586

aliceHendicott:
Thanks @silviuaavram! Looks like it's working as expected with v9.0.1 so I'll close off this PR.

Issue #1295, opened 2021-07-28
Disabled items are highlighted when select is opened with Arrow Up/Down


Eldraed:
- `downshift` version: 6.1.3
- `node` version: 14.17.3
- `npm` (or `yarn`) version: npm 6.14.13

**Relevant code or config**

Example taken from my reproduction :
```javascript
<ul {...getMenuProps()} style={menuStyles}>
  {isOpen &&
    items.map((item, index) => (
      <li
        style={{
          ...(highlightedIndex === index
            ? {backgroundColor: '#bde4ff'}
            : {}),
          // First and last elements are disabled
          ...(index === 0 || index === items.length - 1
            ? {opacity: 0.6}
            : {}),
        }}
        key={`${item}${index}`}
        {...getItemProps({
          item,
          index,
          disabled: index === 0 || index === items.length - 1,
        })}
      >
        {item}
      </li>
    ))}
</ul>
```
I just consider that my first and last elements in my list are disabled.

**What you did**:
I have a list of items with some of them disabled. I want to navigate through my enabled items only with keyboard.

**What happened**:
With `useSelect` (the only one i use):
When i use my keyboard for navigation, with focus on my Toggle button, if i press Arrow up/down to open the list, first or last element will be highlighted even if it's disabled.

![HighlightOnDisabled](https://user-images.githubusercontent.com/16618948/127330855-3ec8f912-5b2a-4466-80f8-17ad50296675.gif)
- First, i just open it with my mouse, we can see that the first and last items can't be highlighted
- Then, i open it with *Arrow down*, first item is highlighted, even if it's disabled
- Same with *Arrow up* and last item

**Reproduction repository**:

Here is where i have recorded the video :
https://codesandbox.io/s/snowy-cloud-t5zyw?file=/src/hooks/useSelect/basic-usage.js
It's a fork of Downshift examples with changes to reproduce on useSelect basic usage example.

**Problem description**:

Disabled elements should not be highlighted on menu opening with arrow keys

**Suggested solution**:

The reducer here : https://github.com/downshift-js/downshift/blob/master/src/hooks/useSelect/reducer.js#L42-L63 is using `getHighlightedIndexOnOpen` to know which item to highlight on menu opening.
But it returns `0` or `items.length - 1` if it doesn't fit with `initial`, `default` and `selectedItem` values : https://github.com/downshift-js/downshift/blob/master/src/hooks/utils.js#L304

[...]

PR #1587, opened 2024-04-02
fix(hooks): do not highlight disabled items on menu open

silviuaavram:
**What**:
Check if items are disabled before highlighting them, when opening the menu with initialHighlightedIndex or defaultHighlightedIndex.

**Why**:
Fixes https://github.com/downshift-js/downshift/issues/1584.

**How**:
Check if the items are disabled before returning defaultHighlightedIndex / initialHighlightedIndex as the highlightedIndex on menu open.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [x] Ready to be merged

### DIFF
--- a/src/hooks/useCombobox/__tests__/props.test.js
+++ b/src/hooks/useCombobox/__tests__/props.test.js
@@ -334,6 +334,71 @@ describe('props', () => {
     expect(input).toHaveAttribute('aria-activedescendant', expectedItemId)
   })
 
+  test('initialHighlightedIndex is ignored if item is disabled', async () => {
+    const initialHighlightedIndex = 2
+    renderCombobox({
+      initialHighlightedIndex,
+      isItemDisabled(item) {
+        return items.indexOf(item) === initialHighlightedIndex
+      },
+    })
+
+    await clickOnInput()
+
+    expect(getInput()).toHaveAttribute('aria-activedescendant', '')
+  })
+
+  test('initialHighlightedIndex is ignored and defaultHighlightedIndex is chosen if enabled', async () => {
+    const initialHighlightedIndex = 0
+    const defaultHighlightedIndex = 2
+    renderCombobox({
+      initialHighlightedIndex,
+      defaultHighlightedIndex,
+      isItemDisabled(item) {
+        return items.indexOf(item) === initialHighlightedIndex
+      },
+    })
+
+    await clickOnInput()
+
+    expect(getInput()).toHaveAttribute(
+      'aria-activedescendant',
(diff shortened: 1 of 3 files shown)


silviuaavram:
Found the issue. Will fix it shortly. Indeed, it's a different unhandled case with Arrows.

conflma:
https://codesandbox.io/p/sandbox/morning-morning-326kwz?file=%2Fsrc%2Findex.tsx%3A12%2C18

Just finished the Sandbox in the moment you commented.

Steps to reproduce (Do NOT use the Mouse, Keys only):

- Step into the input by Hitting "Tab"
- Press "Arrow Down"
- Press "Enter"

=> Not only the disabled item is highlighted, it is even selectable and produces basically a falsy / not allowed value within the input.

conflma:
@silviuaavram FYI #1295 also describes the problem. Selection of disabled items is not possible, when the menu is opened with a click and you'll then go trough the list with the keys. But if you OPEN the menu with the keyboard (by pressing ArrowDown, ArrowDown) - basically using only the keyboard - it is selectable.

silviuaavram:
Fix is merged, please wait for 9.0.2 and confirm it works as expected. Thank you!

conflma:
@silviuaavram Thanks! Looking forward to it :) BTW, you mean 9.0.3? 9.0.2 is currently released?

silviuaavram:
yeah, sure. my bad :D

conflma:
@silviuaavram I can confirm it works - it is not selecting a first or last item or any disabled item at the end or the start when, if it is marked as disabled and the menu is either triggered by `ArrowUp` or `ArrowDown`. 

See updated sandbox (https://codesandbox.io/p/sandbox/morning-morning-326kwz)

Thanks for the quick fix!

However, from a UX / a11y perspective, wouldn't it be nice, if the first item (either from the top if the menu was opened with ArrowDown or the end if the menu was opened with ArrowUp), which is not disabled would be highlighted / selected? Currently you have to press either ArrowUp / ArrowDown twice to get to the first item, that is not disabled.

What is your opinion on that? Valid feature request?

silviuaavram:
I think it makes sense. We can update `getHighlightedIndexOnOpen` to leverage `getHighlightedIndex`. Could your create a ticket with the spec? You can also create a PR with the implementation, I'd be happy to review it. Thanks!

conflma:
@silviuaavram I've created the ticket (#1593). I fear that I won't be able to implement it myself due to quite a shortage of time in the moment. I've just stumbled across the problem, because I was building some components for an internal library with downshift.

PR #1627, opened 2024-09-10
aria-activedescendant: undefined instead of ''

### LINKED ISSUE #1626: ARIA attributes must conform to valid values
**What you did**:

Run an a11y testing suite.

**What happened**:

It returns an error saying that the id for aria-controls, aria-labelledby, etc.. are invalid. 

**Reproduction repository**:

I am using `useCombobox` and conditionally displaying the items in a popover using isOpen. Run pa11y using axe.

**Problem description**:
Edit:

This might actually be aria-activedescendant="" causing this problem

```
 • ARIA attributes must conform to valid values
   (https://dequeuniversity.com/rules/axe/4.4/aria-valid-attr-value?application=axeAPI)

   (#shared-preferences-theme-select)

   <input class="css-8tk2dk-input-input" aria-activedescendant=""
   aria-autocomplete="list" aria-controls="downshift-_r7_-menu"
   aria-expanded="false" aria-labelledby="downshift-_r8_-label"
   autocomplete="off" id="shared-preferences-theme-select" role="com...

``` 

___
Related to https://github.com/facebook/react/issues/26839

Accessibility testing tools are likely using `querySelector` to find the correct elements. That requires that the ID attribute is also a
valid CSS selector, which it isn't when using `useId`

**Suggested solution**:
Do something like this
  `useId().replace(/:/g, '_')`

For end users the solution would be

```
  const menuId = `downshift-${useId().replace(/:/g, '_')}-menu`;
  const labelId = `downshift-${useId().replace(/:/g, '_')}-label`;
```
and then apply it to the useCombobox (in my case) hook.

tskarhed:
**What**:

An empty string aria-activedescendant may be considered a faulty value. 

**Why**:

Possibly Fixes #1626. I need to doublecheck when I am able to run these changes against our a11y testing suite.

ARIA attribute has invalid value.

**How**:

Switched from '' to undefined when aria-activedescendant isn't set.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] TypeScript Types
- [ ] Flow Types
- [ ] Ready to be merged

### DIFF
--- a/src/hooks/useCombobox/__tests__/getInputProps.test.js
+++ b/src/hooks/useCombobox/__tests__/getInputProps.test.js
@@ -91,21 +91,21 @@ describe('getInputProps', () => {
       const {result} = renderUseCombobox({highlightedIndex: 2})
       const inputProps = result.current.getInputProps()
 
-      expect(inputProps['aria-activedescendant']).toBe('')
+      expect(inputProps['aria-activedescendant']).toBe(undefined)
     })
 
     test('do not assign aria-activedescendant if no item is highlighted', () => {
       const {result} = renderUseCombobox()
       const inputProps = result.current.getInputProps()
 
-      expect(inputProps['aria-activedescendant']).toBe('')
+      expect(inputProps['aria-activedescendant']).toBe(undefined)
     })
 
     test('do not assign aria-activedescendant if no item is highlighted and menu is open', () => {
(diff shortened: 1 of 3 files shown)


tskarhed:
Nvm, I have tried workarounds for both this and useId, it doesn't seem to help.

PR #1681, opened 2026-03-13
fix(useEnhancedReducer): bring back props in action

### LINKED ISSUE #1678: Combobox input issue starting from v9.2.0 and missing changelog
When we upgrade from v9.0.13 to v9.2.0, the search behavior of our Combobox breaks so that when you try to type, the text disappears. We have used the cursor jump workaround to get around issue that cursor jumps to end when trying to type in the middle of the input text. Removing this workaround fixes the search issue but then we get the cursor jump issue, and it seems that cursor jump issue is never going to be fixed / is unfixable from Downshift's point of view.

So would be really nice to see from somewhere what has happened between those 2 versions that causes this. I could use the compare feature in releases pages but with 128 files changed between the versions and not knowing the codebase makes it kinda hard to try to pinpoint what causes this.

### LINKED ISSUE #1680: useCombobox hook isOpen state "regression"  starting from v9.2.0 ?
Hello! :> The behavior of `useCombobox` seems to have changed when upgrading from v9.0.13 to v9.2.0. And there is no clear changelog to see what happen.  I would not have noticed it if one of my integration tests hadn’t caught it. After debugging, it looks like the `isOpen` state returned by useCombobox behaves differently when some logic is implemented inside the `onStateChange` prop.

However, I am not sure  if there is an intentional change in how `isOpen `is update and it's just happen that on my implementation I rely on custom logic inside the` useCombobox.stateChangeTypes.InputKeyDownEscape` condition block, which now behaves differently.  

### What I observed ###

In v9.0.13:

After navigating with the keyboard in the combobox input that has an initialValue, if I press the `Escape` key without selecting anything: the initialValue correctly persists, and `isOpen` state from `useCombobox` is still true when log it from the `onStateChange` block.

In v9.2.0+:

Running the same scenario: the `isOpen` state from `useCombobox` is already false, which changes the behavior if custom logic depends on that value (write inside the `onStateChange` for my use case)

### To Reproduce ###

Here I try to reproduce and add one simple test example with vitest and RTL in stackblitz (I use vitest browser mode but looks like stackblitz do not support it)  the test pass in v9.013 but failed when upgrade to v9.2 or above.  https://stackblitz.com/edit/vitejs-vite-12uqadji?file=src%2FCombobox.tsx

silviuaavram:
# Pull Request

## What

Bring back action props.

## Why
Regression caused in control props being passed
Closes https://github.com/downshift-js/downshift/issues/1680
Closes https://github.com/downshift-js/downshift/issues/1678

## How
Bring back the props in action.
Improve the internal types to better reflect the enhanced reducer.

## Checklist

- [ ] Documentation
- [ ] Tests
- [ ] TypeScript Types
- [ ] Ready to be merged


silviuaavram:
The fix is ready, but since this involves typescript changes, I will merge it as soon as I am sure that these changes won't conflict with the following TS migration for the hooks. This last part is in progress, but I'm almost done with it.

silviuaavram:
Closing in favor of https://github.com/downshift-js/downshift/pull/1683.

PR #1690, opened 2026-06-03
feat: fix React Compiler compatibility.

phryneas:
# Pull Request

## What

Fix React Compiler compatibility.

## Why

Currently, the `getter` functions are impure and stable. That leads to situations where the React compiler will never re-execute them, keeping state like `value` at the initial value for the whole lifetime of components.

See https://github.com/downshift-js/downshift/issues/1646

This violated the `react-hooks/refs` lint rule: https://react.dev/reference/eslint-plugin-react-hooks/lints/refs

## How

This PR now makes a split:
* effects and handers read from refs
* getters read actual state, and their stability mimicks the stability of state.

It also updates `eslint-plugin-react-hooks` and enables the `/refs` rule. There are more rules that should be enabled, but this is the most important one for what this library is doing here.

## Changes

I tried to do small commits here, so please look at the list of commits.

Note that this changes the signature of a few helpers. This might warrant a major release.

## Checklist

- [ ] Documentation
- [x] Tests
- [x] TypeScript Types
- [ ] Ready to be merged


phryneas:
Oh dang, of course I had to work on an outdated branch... Trying to rebase.

src/hooks/utils/useEnhancedReducer.ts
```diff
@@ -26,41 +26,50 @@ export function useEnhancedReducer<
   isStateEqual: (prev: S, next: S) => boolean,
 ): [S, (action: A) => void] {
   const prevStateRef = React.useRef<S>({} as S)
-  const actionRef = React.useRef<Action<S, A, P>>()
   const enhancedReducer = React.useCallback(
-    (state: S, action: Action<S, A, P>): S => {
-      actionRef.current = action
```
phryneas, line 32: It's not allowed to write a ref in a reducer, so I had to add the `lastAction` to state instead.

src/hooks/utils/useEnhancedReducer.ts
```diff
@@ -26,41 +26,50 @@ export function useEnhancedReducer<
...
   )
-  const propsRef = useLatestRef(props)
+  const propsRef = React.useRef(props)
+  React.useEffect(() => {
+    propsRef.current = props
+  }, [props])
   const dispatchWithProps = React.useCallback(
     (action: A) => dispatch({...action, props: propsRef.current}),
     [propsRef],
   )
```
phryneas, line 55: You could theoretically skip all of this, and just access `props` directly in the reducer (and save it to the side there, the same way I did with `lastAction`), while removing the `useCallback` on the reducer.

The logic behind this is that calling `dispatch` will immediately rerender the reducer-owning component, and since the parents and other inputs won't change, `props` will always be the newest. See [[Why useReducer Is the Cheat Mode of Hooks](https://overreacted.io/a-complete-guide-to-useeffect/#why-usereducer-is-the-cheat-mode-of-hooks)](https://overreacted.io/a-complete-guide-to-useeffect/#why-usereducer-is-the-cheat-mode-of-hooks).

That said, you seem to recently have touched this a lot, so I tried to leave it alone.

src/utils/useLatestRef.ts
```diff
@@ -1,12 +0,0 @@
-import * as React from 'react'
-
-export function useLatestRef<T>(val: T): React.MutableRefObject<T> {
-  const ref = React.useRef(val)
-  // technically this is not "concurrent mode safe" because we're manipulating
-  // the value during render (so it's not idempotent). However, the places this
-  // hook is used is to support memoizing callbacks which will be called
-  // *during* render, so we need the latest values *during* render.
```
phryneas, line 8: This access during render was pretty much the core problem leading to this PR, so I decided to delete this central hook and inline it where it's still needed with an `useEffect`. Happy to change it to a `useLayoutEffect` tho.

phryneas:
@silviuaavram I've also added a compiled version of the `useCombobox` E2E test. 
This fails on `master`, but succeeds on this branch. 
Try it manually in the browser, too - typing in the input field doesn't work at all with the `master` build.

phryneas:
Also, note that this PR only fixes the hooks. I don't believe the `Downshift` component itself can be fixed apart from a full rewrite - which I'm not going to do.

This might need to be documented on top - users of React Compiler would need to use the hooks.

silviuaavram:
Thank you @phryneas I will take a look!

silviuaavram:
Will be taking a look, thanks @phryneas !

src/hooks/utils/useEnhancedReducer.ts
silviuaavram: I initially refactored as you said, but then we got some issues:

https://github.com/downshift-js/downshift/issues/1678

silviuaavram:
brought back useLatestRef with a useLayoutEffect update. let me know what you think!

phryneas:
@silviuaavram Looks good to me!

silviuaavram (approved):

